PR #6, opened 2020-06-29
State NV peer mem driver

adrianchiris:
This commit adds a new state that deploys
NV Peer memory driver container on nodes that have
both mellanox and nvidia hardware.

- Add manifest file
- Nv peer memory state implementation
- Instantiate new state in factory
- Extend node attributes with nvidia gpu pci label


moshe010:
Please fix conflict

adrianchiris:
@moshe010 PR rebase agains master

PR #11, opened 2020-07-12
Add Deployment Example

adrianchiris:
Add some examples on how to:

1. Persistent netdev name via udev rule
2. Deploy network-operator
3. Depoloy secondary Network: multus + macvlan + whereabouts IPAM
4. Deploy Workloads - RDMA and GPUDirect RDMA

Example should be slightly modified to fit each and every deployment
but provides a good starting point to getting the components deployed.

### DIFF
--- a/example/delete-operator.sh
+++ b/example/delete-operator.sh
@@ -1,4 +1,17 @@
 #!/bin/bash
+# Copyright 2020 NVIDIA
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
 
 echo "Deleting Network Operator:"
 echo "##########################"
--- a/example/delete-operator.sh
+++ b/example/delete-operator.sh
@@ -1,0 +1,10 @@
+#!/bin/bash
+
+echo "Deleting Network Operator:"
+echo "##########################"
+kubectl delete -f deploy/operator.yaml
+kubectl delete -f deploy/role_binding.yaml
+kubectl delete -f deploy/service_account.yaml
+kubectl delete -f deploy/role.yaml
+kubectl delete -f deploy/crds/mellanox.com_nicclusterpolicies_crd.yaml
+kubectl delete -f deploy/operator-ns.yaml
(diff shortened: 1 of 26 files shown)


adrianchiris:
@moshe010  Should we have an example for a basic deployment of the operator ?
if yes, then i have compiled something here.

This is regardless of documentation that should be added in general for the project.

example/delete-operator.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
```
moshe010, line 1: copyright for bash?

example/rdma-test-pod2.yml
```diff
@@ -0,0 +1,27 @@
...
+spec:
+  nodeSelector:
+    kubernetes.io/hostname: ubuntu00
+  restartPolicy: OnFailure
+  containers:
+  - image: mellanox/rping-test 
+    name: rdma-test-ctr
+    securityContext:
+      capabilities:
+        add: [ "IPC_LOCK","NET_ADMIN" ]
```
moshe010, line 16: why NET_ADMIN?

example/udev/00_persistent_netdev
```diff
@@ -0,0 +1,4 @@
+# Persistent netdev example using PCI
```
moshe010, line 1: why this is needed?

adrianchiris: its an example of how to do persistent dev name,
you need the same netdev name across setups for rdma device plugin and macvlan cni

i can add a comment explaining that in the file

example/rdma-test-pod2.yml
adrianchiris: if you are not using a cluster IPAM on secondary network and are assigning manually.

Today we dont have an official solution for secondary network IPAM what is cluster unique right ?

there are two net-attach-def CRs in the example, one with whareabouts IPAM and one without.

example/delete-operator.sh
adrianchiris: yes!

example/rdma-test-pod2.yml
moshe010: why not static cni? https://github.com/containernetworking/plugins/tree/master/plugins/ipam/static
You can assign with it in the pod spec. Why we need to elevate the pod with  NET_ADMIN just to put static ip address?

example/udev/00_persistent_netdev
moshe010: why the operator is not doing it by default?

adrianchiris: This was not in the scope, we should discuss more.
At machine boot you need to have some sort of constant that is the same across machines, would that be the PCI address ?

example/rdma-test-pod2.yml
adrianchiris: i was not aware of this CNI.
This is a better alternative than configuring within the pod which requires additional caps.

will modify

PR #14, opened 2020-07-14
Test DCO

adrianchiris:


adrianchiris:
@moshe010  DCO bot is working, thanks !

PR #24, opened 2020-09-03
[WIP] OpenShift Support Added.

sjug:
For review and comments.

Missing resource update changes.

Do not merge.


sjug:
@adrianchiris PTAL

adrianchiris:
@sjug could you rebase the PR so we can start review,

example/deploy/role.yaml
```diff
@@ -171,31 +171,76 @@ metadata:
...
-  - watch
-- apiGroups:
+  - apiGroups:
+    - rbac.authorization.k8s.io
+    resources:
+    - roles
+    - rolebindings
+    verbs:
+    - '*'
+  - apiGroups:
```
adrianchiris, line 181: id like to keep cluster scope resource access to a minimum as well as the verbs the operator is allowed to perform on them.

can you give a bit more context why these new additions are needed ?

example/deploy/role.yaml
```diff
@@ -171,31 +171,76 @@ metadata:
...
+    - patch
+    - update
     - watch
-- apiGroups:
+  - apiGroups:
     - apps
-  resources:
-    - daemonsets
-  verbs:
+    resources:
```
adrianchiris, line 203: do we need to be able to CRUD all resources under apps ?
can this be limited to a subset ?

manifests/stage-ofed-driver/0100_service_account.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: v1
+kind: ServiceAccount
```
adrianchiris, line 2: can you provide some info about why is that needed ?

manifests/stage-ofed-driver/0100_service_account.yaml
```diff
@@ -0,0 +1,5 @@
...
+metadata:
+  name: ofed-driver
+  namespace: mlnx-network-operator
```
adrianchiris, line 5: as an FYI, the resources network-operator deploys are now in a fixed namespace : `mlnx-network-operator-resources`
does this service account need be deployed in that namespace ? if so, you should use the template : `namespace: {{ .RuntimeSpec.Namespace }}`

manifests/stage-ofed-driver/0100_service_account.yaml
adrianchiris: also is it needed per stage ? i see you added it just for ofed-driver stage.

pkg/state/state_skel.go
```diff
@@ -91,32 +92,103 @@ func (s *stateSkel) updateObj(obj *unstructured.Unstructured) error {
...
+		kind == "PrometheusRule" {
+		return true
+	}
+	return false
+}
+
+func checkNestedFields(found bool, err error) {
+	if !found || err != nil {
+		log.Error(err, "")
+		os.Exit(1)
```
adrianchiris, line 112: why not just return an error ?
os.Exit() will exit the entire process right ?

pkg/state/state_skel.go
```diff
@@ -91,32 +92,103 @@ func (s *stateSkel) updateObj(obj *unstructured.Unstructured) error {
...
+	}
+}
+
+func updateResourceVersion(req, found *unstructured.Unstructured) error {
+	kind := found.GetKind()
+
+	if needToUpdateResourceVersion(kind) {
```
adrianchiris, line 119: can you add a comment on why we need to update resource version ?

also i see you are just updating the field with the same value, is that ok ?

im probably lacking some k8s specific knowledge about `resourceVersion` so apologies if the questions seem obvious

pkg/state/state_skel.go
```diff
@@ -91,32 +92,103 @@ func (s *stateSkel) updateObj(obj *unstructured.Unstructured) error {
...
+		version, fnd, err := unstructured.NestedString(found.Object, "metadata", "resourceVersion")
+		checkNestedFields(fnd, err)
+
+		if err := unstructured.SetNestedField(req.Object, version, "metadata", "resourceVersion"); err != nil {
+			return errors.Wrap(err, "Couldn't update ResourceVersion")
+		}
+	}
+
+	if kind == "Service" {
```
adrianchiris, line 128: can you add a comment on why we need to update ClusterIP ?

also i see you are just updating the field with the same value, is that ok ?

im probably lacking some k8s specific knowledge about `ClusterIP ` so apologies if the questions seem obvious

pkg/state/state_skel.go
```diff
@@ -91,32 +92,103 @@ func (s *stateSkel) updateObj(obj *unstructured.Unstructured) error {
...
 		}
-		if !k8serrors.IsAlreadyExists(err) {
-			// Some error occurred
-			return err
+
+		if err != nil {
+			return errors.Wrap(err, "unexpected error")
+		}
+
+		// Short circuit updating ServiceAccounts and Pods
```
adrianchiris, line 172: is that because updates are not allowed on them ?
can you add the reason in the comment ?

adrianchiris (commented):
Thanks for proposing the PR !

comments inline

my main concerns are:
1. the additional permissions added to the operator at cluster scope
2. the addition of role/role binding/service account in stage-ofed

so more info on those is appreciated also i see the additions of objects was only done in stage-ofed, is it not needed for other stages ? (nv peer mem or rdma device plugin ?)

Also id like to flag the openshift specific resources they may not be deployed in non openshift scenarios when helm support is added.

example/deploy/role.yaml
sjug: @adrianchiris Are you asking specifically about this `"" apiGroup` or in general about all the cluster scoped resources?

example/deploy/role.yaml
sjug: As a first pass I likely added more operations under apps in the interest of expedience, I can narrow them down.

manifests/stage-ofed-driver/0100_service_account.yaml
sjug: In order to add the SecurityContextConstraints (SCC) below, we need the corresponding ServiceAccount, etc.

manifests/stage-ofed-driver/0100_service_account.yaml
sjug: Ack changed the namespaces for the `manifests/stage-ofed-driver/*.yaml`

pkg/state/state_skel.go
sjug: Ack changed.

pkg/state/state_skel.go
sjug: That's just the standard pattern for k8s afaik, but some OpenShift resources don't need it/are different API wise.

pkg/state/state_skel.go
sjug: Just lessons learned from other operator issues, I will check into why this is required for `Services`.

pkg/state/state_skel.go
sjug: Added comment

example/deploy/role.yaml
adrianchiris: in general, sorry i should have placed this comment on top at L#174

manifests/stage-ofed-driver/0300_rolebinding.yaml
```diff
@@ -0,0 +1,16 @@
...
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: ofed-driver
+  namespace: {{ .RuntimeSpec.Namespace }}
+subjects:
+- kind: ServiceAccount
+  name: ofed-driver
+  namespace: {{ .RuntimeSpec.Namespace }}
+userNames:
```
adrianchiris, line 15: is the `userNames` attribute intended for older versions of K8s ?  i see the V1 api does not have this attribute[1]

[1] https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#rolebinding-v1-rbac-authorization-k8s-io

manifests/stage-ofed-driver/0400_scc.openshift.yaml
```diff
@@ -0,0 +1,47 @@
+# Please edit the object below. Lines beginning with a '#' will be ignored,
```
adrianchiris, line 1: I believe this will also be needed for nv-peer-mem driver as well,
IMO we should have a separate stage to deploy OpenShift's security related orbjects. WDYT ?

the alternative is having this duplicated in every stage as `stage-ofed-driver` can be opted out.

pkg/state/state_skel.go
```diff
@@ -91,32 +91,104 @@ func (s *stateSkel) updateObj(obj *unstructured.Unstructured) error {
...
+		return errors.Wrap(err, "nested field error")
+	}
+}
+
+func updateResourceVersion(req, found *unstructured.Unstructured) error {
+	kind := found.GetKind()
+
+	if needToUpdateResourceVersion(kind) {
+		version, fnd, err := unstructured.NestedString(found.Object, "metadata", "resourceVersion")
+		checkNestedFields(fnd, err)
```
adrianchiris, line 119: `checkNestedFields` will now silently fail, i don't think we want that right ?

As we (now) don't call `os.exit` i think an if statement is needed to check for errors.
either remove `checkNestedFields` and implement the logic directly or have it return an error and check it here.

pkg/state/state_skel.go
```diff
@@ -91,32 +91,104 @@ func (s *stateSkel) updateObj(obj *unstructured.Unstructured) error {
...
+	return false
+}
+
+func checkNestedFields(found bool, err error) {
+	if !found || err != nil {
+		return errors.Wrap(err, "nested field error")
+	}
+}
+
+func updateResourceVersion(req, found *unstructured.Unstructured) error {
```
adrianchiris, line 114: we update both resourceVersion and clusterIP for service.
down the road it might update some other fields depending on the object kind.

perhaps call it: `updateInternalFields` ?

pkg/state/state_skel.go
adrianchiris: From what i can tell from the below this is needed for implementing optimistic concurrency: 

https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#concurrency-control-and-consistency

pkg/state/state_skel.go
```diff
@@ -91,32 +91,109 @@ func (s *stateSkel) updateObj(obj *unstructured.Unstructured) error {
...
+	kind := found.GetKind()
+
+	if needToUpdateResourceVersion(kind) {
+		version, fnd, err := unstructured.NestedString(found.Object, "metadata", "resourceVersion")
+		if err := checkNestedFields(fnd, err); err != nil {
+			return errors.Wrap(err, "Couldn't check nested fields.")
+		}
+
+		if err := unstructured.SetNestedField(req.Object, version, "metadata", "resourceVersion"); err != nil {
```
adrianchiris, line 124: nit: why not use:  `req.SetResourceVersion(version)` ?

PR #25, opened 2020-09-07
nicclusterpolicy as Cluster resource

adrianchiris:
Change nicclusterpolicy CRD to be cluster scoped

This commit modifies nicclusterpolicy CRD to be cluster
scope instead of namespace scope as the CRD defines
cluster wide parameters that are not confined to a certain
namespace.

- Update CRD to be Cluster scoped
- Update deploy templates
- Update example deployment templates


adrianchiris:
CI needs to be adjusted to accomodate the changes.
This has been tested manually

Issue #27, opened 2020-09-08
Log levels are not aligned with log implementation constants 


adrianchiris:
By default `INFO` logging level is enabled, however leveled logging are not being logged.
it is needlesly required to specify the log level via `--zap-level` flag.

defined log level values should align with current log implementation (zap).

example log output:
```
{"level":"Level(-3)","ts":1599571522.4922545,"logger":"state","msg":"Object updated successfully"}
{"level":"Level(-3)","ts":1599571522.4922922,"logger":"state","msg":"Checking related object states"}
{"level":"Level(-3)","ts":1599571522.492306,"logger":"state","msg":"Checking object","Kind:":"ConfigMap","Name":"rdma-devices"}
{"level":"Level(-3)","ts":1599571522.49233,"logger":"state","msg":"Get Object","Namespace:":"mlnx-network-operator-resources","Name:":"rdma-devices"}
{"level":"Level(-3)","ts":1599571522.495664,"logger":"state","msg":"Object is ready","Kind:":"ConfigMap","Name":"rdma-devices"}
{"level":"Level(-3)","ts":1599571522.495705,"logger":"state","msg":"Checking object","Kind:":"DaemonSet","Name":"rdma-shared-dp-ds"}
{"level":"Level(-3)","ts":1599571522.4957476,"logger":"state","msg":"Get Object","Namespace:":"mlnx-network-operator-resources","Name:":"rdma-shared-dp-ds"}
{"level":"Level(-3)","ts":1599571522.50007,"logger":"state","msg":"Object is ready","Kind:":"DaemonSet","Name":"rdma-shared-dp-ds"}
```

adrianchiris:
addressed in PR #29

PR #29, opened 2020-09-24
Align log levels with operator-sdk default impl

adrianchiris:
This commit aligns the defiend log levels with
the default logging implementation used (zap).

Leveled log messages will now be printed with
their proper string and adhere to the log level
provided to the process via --zap-level flag.


adrianchiris:
fixes issue #27

PR #40, opened 2020-10-13
Remove kernel ver from mofed driver container

adrianchiris:
This PR removed Kernel version dependency from ofed-driver stage

### DIFF
--- a/pkg/nodeinfo/attributes.go
+++ b/pkg/nodeinfo/attributes.go
@@ -43,22 +43,25 @@ type AttributeType int
 
 // Attribute type Enum, add new types before Last and update the mapping below
 const (
-	AttrTypeOS = iota
-	AttrTypeKernel
+	AttrTypeOSNameFull = iota
 	AttrTypeHostname
 	AttrTypeCPUArch
+	AttrTypeOSName
+	AttrTypeOSVer
 	AttrTypeLast
 )
 
 var attrToLabel = [...][]string{
 	// AttrTypeOS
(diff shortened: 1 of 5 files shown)


moshe010 (approved):

moshe010:
DCO is failing

moshe010 (changes requested):
Dco is falling

adrianchiris:
Fixed DCO

pkg/nodeinfo/attributes.go
```diff
@@ -43,22 +43,25 @@ type AttributeType int
 
 // Attribute type Enum, add new types before Last and update the mapping below
 const (
-	AttrTypeOS = iota
-	AttrTypeKernel
+	AttrTypeOSNameFull = iota
 	AttrTypeHostname
 	AttrTypeCPUArch
+	AttrTypeOSName
```
moshe010, line 49: Why do you need OSNameFull? I understand it just OSName + OSVer

adrianchiris: originally i started removing it (and eventually reverted it), it introduced quite abit of code and template changes that hinders readability of this PR.

technically you dont need, it can be handled with the two newly added attributes, its currently kept for convenience.
there can be a separate PR to remove it.

moshe010: ok lets open github issue and assign @Mmduh-483

moshe010 (approved):

PR #48, opened 2020-10-26
Update helm template example for secondary network and MacvlanNetwork

### LINKED ISSUE #46: [enhancement] Support MacVlan secondary network in helm
PR #37 introduced support for macvlan secondary network. We should enable it via helm.

Mmduh-483:
Fixes #46

### DIFF
--- a/deployment/network-operator/README.md
+++ b/deployment/network-operator/README.md
@@ -56,8 +56,8 @@ $ helm install -n network-operator --create-namespace --wait mellanox/network-op
 $ kubectl -n network-operator get pods
 ```
 
->__Note:__ By default the operator is deployed without an instance of `NicClusterPolicy`
-custom resource. The user is required to create it later with configuration matching the cluster or use
+>__Note:__ By default the operator is deployed without an instance of `NicClusterPolicy` and `MacvlanNetwork`
+custom resources. The user is required to create it later with configuration matching the cluster or use
 chart parameters to deploy it together with the operator.
 
 ## Chart parameters
@@ -75,7 +75,7 @@ We have introduced the following Chart parameters.
 | `operator.tag` | string | `None` | Network Operator image tag, if `None`, then the Chart's `appVersion` will be used |
 | `deployCR` | bool | `false` | Deploy `NicClusterPolicy` custom resource according to provided parameters |
 
-### Custom resource parameters
+### NicClusterPolicy Custom resource parameters
 
 #### Mellanox OFED driver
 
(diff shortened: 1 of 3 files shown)


manifests/stage-macvlan-cni/0010-macvlan-ds.yml
```diff
@@ -0,0 +1,57 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: macvlan-cni-ds
```
moshe010, line 17: I don't think we want to maintain images per generic cni. 
Another option is just to install containernetworking-plugins

moshe010 (changes requested):

manifests/stage-macvlan-cni/0010-macvlan-ds.yml
adrianchiris: This discussion should be on PR#37 IMO :)

adrianchiris: So here is what i propose,

rename the stage to be : `stage-containernetworking-plugins` which will deploy a daemonset that will install (cp to /opt/cni/bin on host) the provided cni plugins.

it means we build one image for all plugins based on:
binaries published : https://github.com/containernetworking/plugins/releases

i prefer not to rely on distro specific repos

@moshe010 WDYT ?

deployment/network-operator/README.md
```diff
@@ -122,6 +123,55 @@ resources:
...
+| Name | Type | Default | description |
+| ---- | ---- | ------- | ----------- |
+| `ipamPlugin.deploy` | bool | `true` | Deploy IPAM CNI Plugin Secondary Network  |
+| `ipamPlugin.image` | string | `whereabouts` | IPAM CNI Plugin image name  |
+| `ipamPlugin.repository` | string | `dougbtv` | IPAM CNI Plugin image repository  |
+| `ipamPlugin.version` | string | `latest` | IPAM CNI Plugin image version  |
+
+### MacvlanNetwork Custom resource parameters
+
+| Name | Type | Default | description |
```
adrianchiris, line 167: Im kinda leaning towards not having this part of the chart.
but rather have the user define after deployment.

maybe the user needs more than one ? or none ?

deployment/network-operator/README.md
```diff
@@ -122,6 +122,45 @@ resources:
...
+
+##### Multus CNI Secondary Network
+
+| Name | Type | Default | description |
+| ---- | ---- | ------- | ----------- |
+| `multus.deploy` | bool | `true` | Deploy Multus Secondary Network  |
+| `multus.image` | string | `multus` | Multus image name  |
+| `multus.repository` | string | `nfvpe` | Multus image repository  |
+| `multus.version` | string | `v3.6` | Multus image version  |
+| `multus.config` | string | `` | Multus CNI config if config is missing or empty then multus config will be automatically generated from the CNI configuration file of the master plugin (the first file in lexicographical order in cni-conf-dir)  |
```
adrianchiris, line 153: > Multus CNI config if config is missing or empty then multus config will be automatically...

Multus CNI config, if empty then config will be automatically....

deployment/network-operator/README.md
```diff
@@ -135,7 +174,7 @@ of the network operator in the following manner:
 
 #### Example 1
 Network Operator deployment with a specific version of OFED driver and a single RDMA resource mapped to `enp1`
-netdev.
+netdev, also deploying secondary network with Mutlus CNI, containernetworking-plugins CNI plugins and whereabouts IPAM CNI Plugin.
```
adrianchiris, line 177: lets have this as example 3 which focuses on secondary network enablement with RDMA

i.e
deployment of device plugin and secondary network components.

device selectors PR can then be integrated into it to have a simple recipe to enable RDMA on secondary network.

adrianchiris (changes requested):
small rework to the README and we are good to go !

deployment/network-operator/README.md
```diff
@@ -169,3 +208,34 @@ devicePlugin:
     - name: rdma_shared_device_b
       devices: [ib0]
 ```
+
+#### Example 3
+Network Operator deployment with the default version of RDMA resources, second mapped to `ib0`,
```
adrianchiris, line 213: i dont understand this sentence

adrianchiris (changes requested):

adrianchiris (approved):

PR #50, opened 2020-10-27
Increase ofed-driver startupProbe timeout to 10 minutes

AbdYsn:
When deploying the ofed-driver on a non-supported kernel, the
container will try to rebuild the KDMS to support the new kernel,
this process has varying times based on the server capabilities.
This patch increases the timeout of ofed-driver startupProbe to 10
minutes as a precaution for somewhat old servers to not restart
container build waiting for the modules to be rebuilt.

Signed-off-by: abdallahyas <abdallahyas@mellanox.com>


adrianchiris (approved):
LGTM

PR #54, opened 2020-10-28
Support HTTP proxy setting to be passed down to containers

### LINKED ISSUE #23: [Enhancement] Support HTTP proxy setting to be passed down to containers
When certified for OLM, network-operator may be deployed in openshift environment.
yum repositories may be behind proxy so yum requests should be sent to that proxy.

operator should pass down proxy settings to relevant pods.

Proxy settings will be provided to helm when the operator is installed. these should propagate to the operands created by the operator.

```
  env: 
#    - name: HTTP_PROXY
#      value: {{.Proxy.HttpProxy}}
#    - name: HTTPS_PROXY
#     value: {{.Proxy.HttpsProxy}}
#   - name: NO_PROXY
#      value: {{.Proxy.NoProxy}}
```

Mmduh-483:
Proxy settings will be provided to helm when the operator is installed. these should propagate to the operands created by the operator.

Fixes #23

### DIFF
--- a/deployment/network-operator/README.md
+++ b/deployment/network-operator/README.md
@@ -75,6 +75,16 @@ We have introduced the following Chart parameters.
 | `operator.tag` | string | `None` | Network Operator image tag, if `None`, then the Chart's `appVersion` will be used |
 | `deployCR` | bool | `false` | Deploy `NicClusterPolicy` custom resource according to provided parameters |
 
+### Proxy parameters
+These proxy parameter will translate to HTTP_PROXY, HTTPS_PROXY, NO_PROXY environment variables to be used by the network operator and relevant resources it deploys.
+Production cluster environment can deny direct access to the Internet and instead have an HTTP or HTTPS proxy available.
+
+| Name | Type | Default | description |
+| ---- | ---- | ------- | ----------- |
+| `proxy.httpProxy` | string | `None` | proxy URL to use for creating HTTP connections outside the cluster. The URL scheme must be http |
+| `proxy.httpsProxy` | string | `None` | proxy URL to use for creating HTTPS connections outside the cluster |
+| `proxy.noProxy` | string | `None` | A comma-separated list of destination domain names, domains, IP addresses or other network CIDRs to exclude proxying |
+
 ### NicClusterPolicy Custom resource parameters
 
 #### Mellanox OFED driver
--- a/pkg/consts/consts.go
+++ b/pkg/consts/consts.go
@@ -32,3 +32,9 @@ const (
 	NetworkOperatorResourceNamespace = "mlnx-network-operator-resources"
 	NicClusterPolicyResourceName     = "nic-cluster-policy"
 )
+
+const (
+	HTTPProxy  = "HttpProxy"
+	HTTPSProxy = "HttpsProxy"
+	NoProxy    = "NoProxy"
+)
--- a/deploy/operator.yaml
+++ b/deploy/operator.yaml
@@ -38,7 +38,7 @@ spec:
           # Replace this with the built image name
           image: REPLACE_IMAGE
           command:
-          - network-operator
+            - network-operator
           imagePullPolicy: Always
           env:
             - name: STATE_MANIFEST_BASE_DIR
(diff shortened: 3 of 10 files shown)


pkg/consts/consts.go
```diff
@@ -32,3 +32,9 @@ const (
 	NetworkOperatorResourceNamespace = "mlnx-network-operator-resources"
 	NicClusterPolicyResourceName     = "nic-cluster-policy"
 )
+
+const (
```
adrianchiris, line 36: Should be the same naming convention as the above

deployment/network-operator/README.md
```diff
@@ -75,6 +75,14 @@ We have introduced the following Chart parameters.
 | `operator.tag` | string | `None` | Network Operator image tag, if `None`, then the Chart's `appVersion` will be used |
 | `deployCR` | bool | `false` | Deploy `NicClusterPolicy` custom resource according to provided parameters |
 
+### Proxy parameters
+
```
adrianchiris, line 79: after a bit of googling, it seems these proxy conventions are honored in both ubuntu and centOS so i dont think this needs to be separated per platform.

can you add a small paragraph explaining the these proxy parameters and what they are used for.

adrianchiris (changes requested):

adrianchiris:
minor comments.

I have updated the related issue. IMO this should be put on hold for the time being until there will be a concrete request.
the original use-case was for openshift, however, in-cluster support was added for proxies in OCP 4.6

pkg/consts/consts.go
Mmduh-483: chaged

Mmduh-483:
> the original use-case was for openshift, however, in-cluster support was added for proxies in OCP 4.6

@adrianchiris set WIP

Mmduh-483:
> the original use-case was for openshift, however, in-cluster support was added for proxies in OCP 4.6

@adrianchiris set WIP

deployment/network-operator/README.md
Mmduh-483: added

adrianchiris: i would also add:

These proxy parameter will translate to `HTTP_PROXY`, `HTTPS_PROXY`, `NO_PROXY` environment variables to be used by the network operator and relevant resources it deploys.

adrianchiris (changes requested):
Small comment otherwise LGTM given this enhancement is still needed.

adrianchiris:
After an offline discussion, this is still needed for OpenShift as the cluster wide proxy embeds injects these environment variables only to a subset of workloads. these would be injected to the operator and its expected for it to propagate to its operands where needed.

so we should conditionally propagate in the various states only if needed (e.g we do apt-get update)

@Mmduh-483  FYI

manifests/stage-rdma-device-plugin/0020-k8s-rdma-shared-dev-plugin-ds.yaml
```diff
@@ -48,6 +48,13 @@ spec:
         imagePullPolicy: IfNotPresent
         securityContext:
           privileged: true
+        env:
```
adrianchiris, line 51: We need to propagate these variables only to relevant stages.
e.g stages that their workload require internet connectivity. (e.g do `apt-get update`)

its not needed for this stage

pkg/state/state_shared_dp.go
```diff
@@ -54,7 +54,10 @@ type stateSharedDp struct {
 }
 
 type sharedDpRuntimeSpec struct {
-	Namespace string
+	Namespace  string
+	HttpProxy  string
```
adrianchiris, line 58: not needed for this stage

pkg/state/state_shared_dp.go
```diff
@@ -115,7 +118,10 @@ func (s *stateSharedDp) getManifestObjects(
 	renderData := &sharedDpManifestRenderData{
 		CrSpec: cr.Spec.DevicePlugin,
 		RuntimeSpec: &sharedDpRuntimeSpec{
-			Namespace: consts.NetworkOperatorResourceNamespace,
+			Namespace:  consts.NetworkOperatorResourceNamespace,
+			HttpProxy:  utils.GetEnv(consts.HttpProxy),
```
adrianchiris, line 122: not needed for this stage

adrianchiris (changes requested):
Need also to update the `deploy/operator.yaml` , adding these env variables with `""` value

adrianchiris (approved):
Can you check why CI fails ?

adrianchiris (approved):

PR #55, opened 2020-10-28
Add resource version for object when using CreateAndUpdateObjs

### LINKED ISSUE #47: Modify state.createAndUpdateObjs logic to set resource version for objects
CRD instance needs to modify set resource version for objects on update, which in not currently handled by `state.createAndUpdateObjs`

Mmduh-483:
ResourceVersion is required to update CRD instances

Fixes #47

### DIFF
--- a/pkg/state/state_skel.go
+++ b/pkg/state/state_skel.go
@@ -112,6 +112,17 @@ func (s *stateSkel) createOrUpdateObjs(
 			return err
 		}
 
+		// Set resource version
+		// ResourceVersion must be passed unmodified back to the server.
+		// ResourceVersion helps the kubernetes API server to implement optimistic concurrency for PUT operations
+		// when two PUT requests are specifying the resourceVersion, one of the PUTs will fail.
+		currentObj := desiredObj.DeepCopy()
+		if err := s.getObj(currentObj); err != nil {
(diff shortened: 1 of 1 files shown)


pkg/state/state_skel.go
```diff
@@ -112,6 +112,14 @@ func (s *stateSkel) createOrUpdateObjs(
 			return err
 		}
 
+		// Set resource version
```
adrianchiris, line 115: So, i think this deserves some more info in the comment.

https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#concurrency-control-and-consistency

Mmduh-483: From the doc this value must be passed unmodified back to the server and the  only way for a client to know the expected value of resourceVersion is to have received it from the server

adrianchiris: It is used by the kubernetes API server to implement optimistic concurrency for PUT operations.
id like you to add a comment here on why we are setting the resource version so its clear.

adrianchiris (approved):

adrianchiris (changes requested):
Just add more context to commit message and this can be merged

Issue #56, opened 2020-11-03
MOFED driver container arch name change


adrianchiris:
MOFED driver containers are expected to be named with the following format:

mofed-<branch-name>-<cpu-arch>:<release-version>-<os name><os-ver>
e.g
`mofed-5.1-x86_64:5.1-2.5.1.1-ubuntu18.04`

stage ofed driver currently embeds cpu arch part of the image tag and not user name and in addition uses k8s node arch which is `amd64`.

It is required to adjust the code to be able to pull mofed images with the new naming scheme.

adrianchiris:
final format is : mofed-<branch>-<version>:<os>-<arch>

PR #60, opened 2020-11-09
Update state OFED with new container image name format

adrianchiris:
Latest official MOFED driver containers have changed
image fromat to:
mofed-<Branch Name>:<Version>-<OS Name><OS Version>-<CPU Arch>

e.g mofed-5.1-2:5.1-2.5.1.1-ubuntu18.04-x86_64

modify template and state ofed code to fit the new format.
- ofed-driver daemonset template
  - Modify container image name
  - Remove CPU arch from daemonset name and templates as it
    is not required and kubernetes forbits _ in object names.
- state ofed driver: translate golang amd64 CPU Arch to x86_64


adrianchiris:
Addresses Issue #56

moshe010 (approved):

PR #62, opened 2020-11-15
Update components version and README

adrianchiris:
This performs the following updates

- Update k8s-rdma-shared-dp to last release tag of v1.1.0
- Change Whereabouts IPAM CNI to point to the last release tag
  of the upstream repository (github.com/dougbtv/whereabouts)
  instead of openshift downstream repo
- README updates of the above and minor typo fix
- Edit deployment README and remove custom version from examples
- Add section to deployment README about NFD
- update nicclusterpolicy CR example with the versions

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,7 +76,7 @@ to be deployed on RDMA & GPU supporting nodes (required for GPUDirect workloads)
 - `SecondaryNetwork`: Specifies components to deploy in order to facilitate a secondary network in Kubernetes. It consists of the folowing optionally deployed components:
     - [Multus-CNI](https://github.com/intel/multus-cni): Delegate CNI plugin to support secondary networks in Kubernetes
     - CNI plugins: Currently only [containernetworking-plugins](https://github.com/containernetworking/plugins) is supported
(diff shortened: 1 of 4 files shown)


adrianchiris:
@moshe010 LMK if the updates are OK

README.md
```diff
@@ -210,7 +211,7 @@ Can be found at: `example/deploy/crds/mellanox.com_v1alpha1_macvlannetwork_cr.ya
 * RDMA capable hardware: Mellanox ConnectX-4 NIC or newer.
 * NVIDIA GPU and driver supporting GPUDirect e.g Quadro RTX 6000/8000 or Tesla T4 or Tesla V100 or Tesla V100.
 (GPU-Direct only)
-* Operating System: Ubuntu18.04 with kernel `4.15.0-109-generic`.
```
moshe010, line 213: but we support Ubuntu 18.04 in the alpha so why to remove it?

adrianchiris: will keep 18.04, its not clear whether we need to keep support for it or not imo as egx stack is switching to 20.04

jliberma:
We should keep 18.04, ideally we support same OS as GPU operator.

Jacob

moshe010 (approved):

PR #66, opened 2020-12-09
Make whereabouts daemonset use hostNetwork

Mmduh-483:
hereabouts is IPAM plugin which should not depend on another cni to be deployed or deleted

### DIFF
--- a/manifests/stage-whereabouts-cni/0050-whereabouts-ds.yaml
+++ b/manifests/stage-whereabouts-cni/0050-whereabouts-ds.yaml
@@ -32,6 +32,7 @@ spec:
         app: whereabouts
         name: whereabouts
     spec:
+      hostNetwork: true
       serviceAccountName: whereabouts
       nodeSelector:
         beta.kubernetes.io/arch: amd64


adrianchiris:
@Mmduh-483 can you elaborate why this is needed ?

Mmduh-483:
@adrianchiris

manifests/stage-whereabouts-cni/0050-whereabouts-ds.yaml
```diff
@@ -32,6 +32,7 @@ spec:
         app: whereabouts
         name: whereabouts
     spec:
+      hostNetwork: true
```
adrianchiris, line 35: so i see whereabouts repo also deploys with hostNetwork: https://github.com/dougbtv/whereabouts/blob/899672448c2cd301f641eaaa5995f20a4e2cdf05/doc/daemonset-install.yaml#L60

adrianchiris (approved):
Need CI to pass but LGTM

AbdYsn:
@adrianchiris The patch is ready

PR #67, opened 2020-12-09
Mount /dev/inside rdma shared device plugin

### LINKED ISSUE #31: [bug] rdma shared dp container fails with "non-Rdma Device" after reboot
after deploying the nic operator with rdma-shared device plugin ,ofed driver, and nv_peer_mem, and successfully creating traffic. if the machine is rebooted, the rdma shared device plugin container may fail because it starts before the ofed driver container is finished loading the modules and causing the rdma shared device plugin to complain in the logs (the container finishes its work and it will be in ready state) with "non-Rdma Device <interace name>".

Workaround: restart rdma shared device plugin containers

Mmduh-483:
Currently rdma resources "/dev/infiniband" are mounted by using hostNetwork, where rdma resources will not be tracked when kernel modules/drivers are installed after deploying rdma shared device plugin.

This patch mounts "/dev/infiniband" inside the container with volumes which keeps tracking the changes to "/dev/infiniband"

Fixes #31

### DIFF
--- a/manifests/stage-rdma-device-plugin/0020-k8s-rdma-shared-dev-plugin-ds.yaml
+++ b/manifests/stage-rdma-device-plugin/0020-k8s-rdma-shared-dev-plugin-ds.yaml
@@ -53,6 +53,8 @@ spec:
             mountPath: /var/lib/kubelet/
           - name: config
             mountPath: /k8s-rdma-shared-dev-plugin
+          - name: devs
+            mountPath: /dev/
       volumes:
         - name: device-plugin
           hostPath:
@@ -63,5 +65,8 @@ spec:
             items:
             - key: config.json
               path: config.json
+        - name: devs
+          hostPath:
+            path: /dev/
       nodeSelector:
         feature.node.kubernetes.io/pci-15b3.present: "true"


manifests/stage-rdma-device-plugin/0020-k8s-rdma-shared-dev-plugin-ds.yaml
```diff
@@ -53,6 +53,8 @@ spec:
             mountPath: /var/lib/kubelet/
           - name: config
             mountPath: /k8s-rdma-shared-dev-plugin
+          - name: infiniband
```
adrianchiris, line 56: nit: rdma-devs

manifests/stage-rdma-device-plugin/0020-k8s-rdma-shared-dev-plugin-ds.yaml
```diff
@@ -63,5 +65,8 @@ spec:
             items:
             - key: config.json
               path: config.json
+        - name: infiniband
```
adrianchiris, line 68: nit: rdma-devs

adrianchiris (commented):
just a small nit, so its clear that this is the folder where rdma devices are located and not just "infiniband" which is a bad legacy name in kernel.

btw good commit message Mamduh !

manifests/stage-rdma-device-plugin/0020-k8s-rdma-shared-dev-plugin-ds.yaml
```diff
@@ -53,6 +53,8 @@ spec:
             mountPath: /var/lib/kubelet/
           - name: config
             mountPath: /k8s-rdma-shared-dev-plugin
+          - name: rdma-devs
```
adrianchiris, line 56: so now the name is incorrect as the path now points to all devices :)  

maybe:
`name: devs`

adrianchiris:
@Mmduh-483 can you do a similar PR in k8s-rdma-shared-dp project as well ?

PR #71, opened 2020-12-13
Update ofed image from ofed-driver to mofed with latest version

Mmduh-483:


adrianchiris:
@Mmduh-483 we should wait with this until MOFED image is officially released,
In addition need to update README as well as other examples throughout the project.

adrianchiris:
we should wait until an official version of mofed driver container is released

adrianchiris:
GA driver container : mofed-5.2-1.0.4.0

Mmduh-483:
@adrianchiris Updated

adrianchiris (approved):

PR #72, opened 2020-12-13
Update NvGPU node selector to nvidia.com/gpu.present

### LINKED ISSUE #70: use gpu-feature-discovery lables to identify gpu nodes
we should use:

```
nodeSelector:
        nvidia.com/gpu.present: "true"
```

for nv-peer-memory driver container as this is made available as of v1.2.0 of GPU Operator

Mmduh-483:
Fixes #70


adrianchiris (approved):
LGTM, i will reword commit message to have more context

PR #75, opened 2020-12-14
Remove multus generate config file when undeploy multus

### LINKED ISSUE #74: Network operator fails to delete after deploying secondary network multus
When deploying Multus secondary network, it generates config file `00-multus.conf` which will be CNI config file, when Multus has removed the config file and bin file will stay which leading to use Multus even if the secondary network was deleted.

When Muluts is deleted also the permission will be deleted, thus lead to failure in creating/removing any pod that uses CNI including removing the network operator itself.

https://github.com/intel/multus-cni/pull/593 fixes the issue in Multus

Mmduh-483:
When deploying Multus secondary network, it generates config file
00-multus.conf which will be CNI config file, when Multus has
removed the config file and bin file will stay which leading to
use Multus even if the secondary network was deleted.

When Muluts is deleted also the permission will be deleted, thus
lead to failure in creating/removing any pod that uses CNI
including removing the network operator itself.

This commit fixes the issue deleting the config file generated by
multus

Fixes #74

### DIFF
--- a/manifests/stage-multus-cni/0050-multus-ds.yml
+++ b/manifests/stage-multus-cni/0050-multus-ds.yml
@@ -35,7 +35,15 @@ spec:
           command: ["/entrypoint.sh"]
           args:
             - "--cni-version=0.4.0"
+            # /tmp/multus-conf/00-multus.conf is where multus-cfg ConfigMap is mounted then entrypoint.sh copy it to
+            # /host/etc/cni/net.d/00-multus.conf
             - "--multus-conf-file={{- if .CrSpec.Config -}}/tmp/multus-conf/00-multus.conf{{- else -}}auto{{- end -}}"
+          # Remove multus config file to prevent failing of creating/deleting pods since multus will fail due to
+          # permission issue, https://github.com/intel/multus-cni/issues/592
+          lifecycle:
+            preStop:
+              exec:
+                command: ["/bin/sh", "-c", "rm -f /host/etc/cni/net.d/00-multus.conf"]
           resources:
             requests:
               cpu: "100m"


manifests/stage-multus-cni/0050-multus-ds.yml
```diff
@@ -36,6 +36,10 @@ spec:
           args:
             - "--cni-version=0.4.0"
             - "--multus-conf-file={{- if .CrSpec.Config -}}/tmp/multus-conf/00-multus.conf{{- else -}}auto{{- end -}}"
+          lifecycle:
```
adrianchiris, line 39: @Mmduh-483 
can you add a comment here explaining why we do that with a link to the multus issue  ?

Mmduh-483: Added

Mmduh-483:
After internal discussion, it is decided to go with this approach to let the network operator cleans the Multus-CNI generated config and rollback to the original primary network

adrianchiris (approved):

manifests/stage-multus-cni/0050-multus-ds.yml
```diff
@@ -36,6 +36,12 @@ spec:
           args:
             - "--cni-version=0.4.0"
             - "--multus-conf-file={{- if .CrSpec.Config -}}/tmp/multus-conf/00-multus.conf{{- else -}}auto{{- end -}}"
```
e0ne, line 38: nit: it would be good to mention why we use mount configmap to /tmp/multus-conf/00-multus.conf  here and remove other file

moshe010: so I think @Mmduh-483  change will work only when --multus-conf-file=auto. 
@Mmduh-483  have you check the flow if .CrSpec.Config ?

e0ne: @moshe010 actually, my comment doesn't relate to this change but having lines ##38 and ##44 together looks a bit confusing without some code investigation. That's why I suggested adding a commennt

Mmduh-483: @moshe010 it is working either use auto or custom config `00-multus.conf` is going to be generated

Mmduh-483: @e0ne `/tmp/multus-conf/00-multus.conf` is where the configMap mounted in the container

moshe010: So just explain that in a comment

Mmduh-483: Added comment for it

moshe010 (approved):

PR #87, opened 2021-01-20
Add MacvlanNetwork CR example

e0ne:


README.md
```diff
@@ -209,7 +209,7 @@ spec:
     }
 ```
 
-Can be found at: `example/deploy/crds/mellanox.com_v1alpha1_macvlannetwork_cr.yaml`
+Can be found at: `./deploy/crds/mellanox.com_v1alpha1_macvlannetwork_cr.yaml`
```
adrianchiris, line 212: can you copy the file under example folder  (all example files are there)? i think it was just missing.

e0ne: It will lead to maintaining two copies of the file. Do we want it?

adrianchiris: i would change it to some generic placeholder under /deploy/  similarly to what was done in:

https://github.com/Mellanox/network-operator/blob/master/deploy/crds/mellanox.com_v1alpha1_nicclusterpolicy_cr.yam

adrianchiris: the operator-framework creates some "sample" CRs when auto-generating the CRDs, these usually should have no meaningful content. i must have missed that mamduh put this here and not under example folder

e0ne: It's already in /deploy directory: https://github.com/Mellanox/network-operator/tree/master/deploy/crds

adrianchiris (changes requested):
as discussed, its better to have the macvlan cr under `example`

and have a simple template under `deploy/crds`

moshe010 (approved):

PR #88, opened 2021-01-20
MacvlanNetwork CR Status.Reason changed in a controller now

e0ne:


pkg/state/state_macvlan_network.go
```diff
@@ -118,8 +113,6 @@ func (s *stateMacvlanNetwork) Sync(customResource interface{}, _ InfoCatalog) (S
 
 	// Get NetworkAttachmentDefinition SelfLink
 	if err := s.getObj(netAttDef); err != nil {
-		// TODO handling MacvlanNetwork CR Status should be done in the controller
-		cr.Status.Reason = "failed to get NetworkAttachmentDefinition"
 		return SyncStateError, errors.Wrap(err, "failed to get NetworkAttachmentDefinition")
 	}
 	cr.Status.MacvlanNetworkAttachmentDef = netAttDef.GetSelfLink()
```
adrianchiris, line 118: will you be handling this in the PR ?

so this status is also set in the controller ?

e0ne: I didn't catch what do you mean.  This Reason change will be handled as others in this PR

adrianchiris: this line changes `cr.Status`

all `cr` changes should be done in controller. so this is still WIP ?

e0ne: Sorry, I missed that. Will take a look on it

adrianchiris:
Added [WIP] prefix, just remove it when ready for reviews, thanks !!

pkg/state/state_macvlan_network.go
adrianchiris: @e0ne if further non-trivial work is required for this (e.g extending some Interfaces)
Then i think this can be done in a separate PR if needed.

e0ne:
retest-all

adrianchiris (approved):

Issue #89, opened 2021-01-20
SRIOV operator "unknown field "eSwitchMode""


e0ne:
Deploying an SriovNetworkNodePolicy with an eSwitchMode field raises "unknown field eSwitchMode" even though the image version does support this field.

PR #90, opened 2021-01-20
Add 'eSwitchMode' flag support

e0ne:
NIC Device Mode cpould be set to  "legacy" or "switchdev" during
sriov-network-operator configuration.

Fixes: Issue #89

### DIFF
--- a/deployment/network-operator/charts/sriov-network-operator/crds/sriovnetwork.openshift.io_sriovnetworknodepolicies_crd.yaml
+++ b/deployment/network-operator/charts/sriov-network-operator/crds/sriovnetwork.openshift.io_sriovnetworknodepolicies_crd.yaml
@@ -52,6 +52,12 @@ spec:
                 - netdevice
                 - vfio-pci
                 type: string
+              eSwitchMode:
+                description: NIC Device Mode. Allowed value "legacy","switchdev".
+                enum:
+                - legacy
+                - switchdev
+                type: string
               isRdma:
                 description: RDMA mode. Defaults to false.
                 type: boolean


deployment/network-operator/charts/sriov-network-operator/crds/sriovnetwork.openshift.io_sriovnetworknodepolicies_crd.yaml
```diff
@@ -73,7 +79,11 @@ spec:
                 properties:
                   deviceID:
                     description: The device hex code of SR-IoV device. Allowed value
-                      "158b", "1015", "1017".
+                      "0d58", "1572", "158b", "1013", "1015", "1017", "101b".
```
moshe010, line 82: I think we should remove the intel devices https://pci-ids.ucw.cz/read/PC/8086

e0ne: @moshe010 should we add 1013 ConnectX-4?

adrianchiris: thats just the description string in the CRD.

this is a direct copy from sriov-network-operator.

if we want network-operator to define a different set of CRDs that sriov-network-operator will work with then:

1. they need to be fully compatible with sriov-network-operator different controllers so they operate as expected
2. maintain these crds in their own folder in the project, these should be then mirrored to the chart we maintain under `deployment`

in addition we may also want to explicitly define some config maps sriov-network-operator uses to customize its controllers which are created by default when its deployed.

adrianchiris:
does it work when deploying via helm? 

no need to update sriov-network-operator images in : https://github.com/Mellanox/network-operator/blob/f12f5b6ee1ded1d1d75658b0308575944e030ae3/deployment/network-operator/charts/sriov-network-operator/values.yaml#L44

e0ne:
> does it work when deploying via helm?
> 
> is there no need to update sriov-network-operator images ?
> 
> https://github.com/Mellanox/network-operator/blob/f12f5b6ee1ded1d1d75658b0308575944e030ae3/deployment/network-operator/charts/sriov-network-operator/values.yaml#L44

We've already got the latest version v4.8

e0ne:
> does it work when deploying via helm?
> 
> is there no need to update sriov-network-operator images ?
> 
> https://github.com/Mellanox/network-operator/blob/f12f5b6ee1ded1d1d75658b0308575944e030ae3/deployment/network-operator/charts/sriov-network-operator/values.yaml#L44

I tested deployment via helm chart from master branch and sriovnetworknodepolicies.sriovnetwork.openshift.io CRD doesn't contain eSwitchMode option

adrianchiris:
I see, so the images we use already support this configuration.

adrianchiris (approved):

moshe010 (approved):

PR #93, opened 2021-02-02
Add helm chart tests

Mmduh-483:
This PR bring 2 helm chart tests for network operator that tests device plugin resources and rping for deployment

### DIFF
--- a/deployment/network-operator/README.md
+++ b/deployment/network-operator/README.md
@@ -100,6 +100,23 @@ $ helm install --devel --set nfd.enabled=false -n network-operator --create-name
 custom resources. The user is required to create it later with configuration matching the cluster or use
 chart parameters to deploy it together with the operator.
 
+## Helm Tests
+
+Network Operator has Helm tests to verify deployment. To run tests it is required to set the following chart parameters on helm install/upgrade: `deployCR`, `devicePlugin`, `secondaryNetwork` as the test depends on `NicClusterPolicy` instance being deployed by Helm.
+Supported Tests:
+- Device Plugin Resource: This test creates a pod that requests the first resource in `devicePlugin.resources`
+- RDMA Traffic: This test creates a pod that test loopback RDMA traffic with `rping`
+
+Run the helm test with following command after deploying network operator with helm
+```
+$ helm test -n network-operator network-operator --timeout=5m
+```
+Notes:
+- Test will keeping running endlessly if pod creating failed so it is recommended to use `--timeout` which fails test after exceeding given timeout
+- Default PF to run test is `ens2f0` to override it add `--set test.pf=<pf_name>` to `helm install/upgrade`
+- Tests should be executed after `NicClusterPolicy` custom resource state is `Ready`
+- In case of a test failed it is possible to collect the logs with `kubectl logs -n <namespace> <test-pod-name>`
+
 ## Chart parameters
 
 In order to tailor the deployment of the network operator to your cluster needs
--- a/deployment/network-operator/templates/tests/test-dev-plugin.yaml
+++ b/deployment/network-operator/templates/tests/test-dev-plugin.yaml
@@ -1,0 +1,28 @@
+{{- if and .Values.deployCR .Values.devicePlugin.deploy  }}
+apiVersion: v1
+kind: Pod
+metadata:
+  name: "{{ .Release.Name }}-dev-plugin-test"
+  annotations:
+    helm.sh/hook: test
+    helm.sh/hook-delete-policy: before-hook-creation,hook-succeeded
+spec:
+  restartPolicy: Never
+  containers:
+    - image: mellanox/rping-test
+      name: "{{ .Release.Name }}-dev-plugin-test"
+      securityContext:
+        capabilities:
+          add: [ "IPC_LOCK" ]
+      resources:
+        requests:
+          rdma/{{ (index .Values.devicePlugin.resources 0).name }}: '1'
+        limits:
+          rdma/{{ (index .Values.devicePlugin.resources 0).name }}: '1'
+      command:
+        - sh
+        - -c
+        - |
+          ls -l /dev/infiniband /sys/class/net /sys/class/infiniband
+          if [[ "$(ls /dev/infiniband/ | wc -l)" -lt 4 ]]; then echo "Not all RDMA resouces mounted" >&2; exit 1; fi
+{{- end }}
--- a/deployment/network-operator/templates/tests/test-rping.yaml
+++ b/deployment/network-operator/templates/tests/test-rping.yaml
@@ -1,0 +1,49 @@
+{{- if and .Values.deployCR .Values.devicePlugin.deploy .Values.secondaryNetwork.deploy .Values.secondaryNetwork.multus.deploy .Values.secondaryNetwork.cniPlugins.deploy }}
+apiVersion: mellanox.com/v1alpha1
+kind: MacvlanNetwork
+metadata:
+  name: "{{ .Release.Name }}-rping-macvlan-network-test"
+  annotations:
+    helm.sh/hook: test
+    helm.sh/hook-delete-policy: before-hook-creation,hook-succeeded,hook-failed
+spec:
+  networkNamespace: "{{ .Release.Namespace }}"
+  mode: "bridge"
+  master: "{{ .Values.test.pf }}"
+  mtu: 1500
+  ipam: |
+    {
+      "type": "static",
+      "addresses": [{ "address": "10.10.0.1/24" }]
+    }
+---
+apiVersion: v1
+kind: Pod
+metadata:
+  name: "{{ .Release.Name }}-rping-test"
+  annotations:
+    helm.sh/hook: test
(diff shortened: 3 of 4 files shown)


deployment/network-operator/README.md
```diff
@@ -216,6 +216,20 @@ Specifies components to deploy in order to facilitate a secondary network in Kub
...
+- Device Plugin Resource: This test create a pod that requests the first resource in `devicePlugin.resources`
+- RDMA Traffic: This test creates a pod that run server and client with `rping`
+
+Run the helm test with following command after deploying network operator with helm
+```
+$ helm test -n network-operator network-operator
+```
+Notes:
+- Test will keeping running endlessly if pod creating failed so it is recommended to use `--timeout` which fails test after exceeding given timeout
+- Default PF to run test is `ens2f0` to override it add `--set test.pf=<pf_name>` to `helm install/upgrade`
```
moshe010, line 231: can we avoid the default and make it as mandatory args?

moshe010 (commented):
can we used it in CI?

deployment/network-operator/README.md
Mmduh-483: yes, when deployed without setting error will appear
```
Error: template: network-operator/templates/tests/test-rping.yaml:9:21: executing "network-operator/templates/tests/test-rping.yaml" at <.Values.test.pf>: nil pointer evaluating interface {}.pf
```

Mmduh-483:
> can we used it in CI?

yes, @AbdYsn can support it in the CI

deployment/network-operator/templates/tests/test-dev-plugin.yaml
```diff
@@ -0,0 +1,28 @@
+{{- if and .Values.deployCR .Values.devicePlugin.deploy  }}
+apiVersion: v1
+kind: Pod
+metadata:
+  name: "{{ .Release.Name }}-dev-plugin-test"
+  annotations:
+    helm.sh/hook: test-success
```
adrianchiris, line 7: we only support helm v3, you can use updated test hook annotation

https://helm.sh/docs/topics/chart_tests/

deployment/network-operator/templates/tests/test-rping.yaml
```diff
@@ -0,0 +1,47 @@
...
+spec:
+  networkNamespace: "{{ .Release.Namespace }}"
+  mode: "bridge"
+  master: "{{ .Values.test.pf }}"
+  mtu: 1500
+  ipam: |
+    {
+      "type": "static",
+      "addresses": [{ "address": "10.10.0.1/24" }],
+      "routes": [{ "dst": "0.0.0.0/0" }]
```
adrianchiris, line 15: routes is mandatory ?

deployment/network-operator/templates/tests/test-rping.yaml
```diff
@@ -0,0 +1,47 @@
+{{- if and .Values.deployCR .Values.devicePlugin.deploy .Values.secondaryNetwork.deploy .Values.secondaryNetwork.multus.deploy .Values.secondaryNetwork.cniPlugins.deploy }}
+apiVersion: mellanox.com/v1alpha1
+kind: MacvlanNetwork
```
adrianchiris, line 3: ATM This is created regardless if tests are run or not right ?

if you add the test hook annotation it will only be created when helm test is invoked however
you probably hit race where the pod was created but the macvlan net-attach-def was not created yet.

also i wonder when is this network deleted ?

deployment/network-operator/README.md
```diff
@@ -216,6 +216,20 @@ Specifies components to deploy in order to facilitate a secondary network in Kub
 | `ipamPlugin.repository` | string | `dougbtv` | IPAM CNI Plugin image repository  |
 | `ipamPlugin.version` | string | `v0.3` | IPAM CNI Plugin image version  |
 
+## Helm Tests
+
+Network Operator has helm tests to verify deployment, currently supported tests:
+- Device Plugin Resource: This test create a pod that requests the first resource in `devicePlugin.resources`
+- RDMA Traffic: This test creates a pod that run server and client with `rping`
```
adrianchiris, line 223: nit

> This test creates a pod that run server and client with `rping`

This test creates a pod that test loopback RDMA traffic with `rping`

deployment/network-operator/README.md
```diff
@@ -216,6 +216,20 @@ Specifies components to deploy in order to facilitate a secondary network in Kub
...
+
+Run the helm test with following command after deploying network operator with helm
+```
+$ helm test -n network-operator network-operator
+```
+Notes:
+- Test will keeping running endlessly if pod creating failed so it is recommended to use `--timeout` which fails test after exceeding given timeout
```
adrianchiris, line 230: >  Test will keeping running endlessly if pod creating failed so it is recommended to use `--timeout` which fails test after exceeding given timeout

so why not put --timeout in the command above ?

deployment/network-operator/README.md
adrianchiris: if we go with the mandatory approach, this comment should state that it is mandatory AND  example command above should reflect that as well.

adrianchiris (changes requested):

deployment/network-operator/README.md
Mmduh-483: It was decided not to go with the mandatory approach

deployment/network-operator/templates/tests/test-dev-plugin.yaml
Mmduh-483: changed to test

deployment/network-operator/templates/tests/test-rping.yaml
Mmduh-483: no, removed

deployment/network-operator/README.md
Mmduh-483: changed

deployment/network-operator/README.md
Mmduh-483: added

deployment/network-operator/templates/tests/test-rping.yaml
Mmduh-483: Changed to be deployed on test with delete policy for failure and success,
tested it for success and worked
for failure it is not removed, maybe issue with helm
https://helm.sh/docs/topics/charts_hooks/

deployment/network-operator/README.md
```diff
@@ -216,6 +216,20 @@ Specifies components to deploy in order to facilitate a secondary network in Kub
 | `ipamPlugin.repository` | string | `dougbtv` | IPAM CNI Plugin image repository  |
 | `ipamPlugin.version` | string | `v0.3` | IPAM CNI Plugin image version  |
 
+## Helm Tests
```
adrianchiris, line 219: i would prefer to put it under the quickstart section and not all the way at the end.

deployment/network-operator/README.md
```diff
@@ -216,6 +216,20 @@ Specifies components to deploy in order to facilitate a secondary network in Kub
...
+
+Network Operator has helm tests to verify deployment, currently supported tests:
```
adrianchiris, line 221: I think you need to mention that these test can work only if you deploy a custom resource instance with helm.
as you can also do `deployCR=false` and deploy it later, in this case the tests will not work.

Also each test will run under different conditions, you should mention that.

deployment/network-operator/README.md
```diff
@@ -216,6 +216,20 @@ Specifies components to deploy in order to facilitate a secondary network in Kub
...
+- Device Plugin Resource: This test creates a pod that requests the first resource in `devicePlugin.resources`
+- RDMA Traffic: This test creates a pod that test loopback RDMA traffic with `rping`
+
+Run the helm test with following command after deploying network operator with helm
+```
+$ helm test -n network-operator network-operator --timeout=5m
+```
+Notes:
```
adrianchiris, line 229: you should also note that tests should be executed after `nicclusterpolicy` custom resource state is `Ready`

adrianchiris (commented):
deployed this PR on my setup, it seems to be OK

a few nits

deployment/network-operator/README.md
```diff
@@ -100,6 +100,22 @@ $ helm install --devel --set nfd.enabled=false -n network-operator --create-name
...
+- Test will keeping running endlessly if pod creating failed so it is recommended to use `--timeout` which fails test after exceeding given timeout
+- Default PF to run test is `ens2f0` to override it add `--set test.pf=<pf_name>` to `helm install/upgrade`
+- Tests should be executed after `nicclusterpolicy` custom resource state is `Ready`
```
adrianchiris, line 117: add a note also on failure , that the test resources will not be cleaned, and how to collect logs for debug

Mmduh-483: updated

adrianchiris: nit: `nicclusterpolicy` -> `NicClusterPolicy`

adrianchiris (approved):

PR #96, opened 2021-02-11
Add status printer column for custom resources

adrianchiris:
This commit adds a `status` printer column which
reflects the global state of `NicClusterPolicy` and `MacvlanNetwork` CR instance.

Example:
```
$> kubectl get nicclusterpolicy
NAME                 STATUS
nic-cluster-policy   ready
```


e0ne:
Since we're changing it, let's do the same for MacvlanNetwork CR too

e0ne (changes requested):
Let's add the same for MacvlanNetwork CR.

adrianchiris:
>Let's add the same for MacvlanNetwork CR.

@e0ne  Done.

e0ne (approved):
Thanks for addressing my comments!

moshe010 (approved):

PR #98, opened 2021-02-11
Allow to override $GOPATH

e0ne:
Moving $GOPATH out of tree is useful for a several reasons during
development:
* it's faster to share directory without .gopath directory
* make targets work inside WSL2 instance - some issue with permissions
  occure so this workaround makes everything working without hours of
  troubleshooting


Makefile
```diff
@@ -17,7 +17,7 @@ BINARY_NAME=network-operator
 PACKAGE=network-operator
 ORG_PATH=github.com/Mellanox
 REPO_PATH=$(ORG_PATH)/$(PACKAGE)
-GOPATH=$(CURDIR)/.gopath
+GOPATH?=$(CURDIR)/.gopath
```
adrianchiris, line 20: make clean will delete it, is that desirable in case of externally provided GOPATH ?

e0ne: Good point. Honestly, I don't care about it at this point.

adrianchiris (approved):

moshe010 (approved):

PR #104, opened 2021-02-21
Helm deployment fail when set deployCR and all NicClusterPolicy components are unset

### LINKED ISSUE #99: Helm deployment fail when deployCR set true and all NicClusterPolicy Components are set false
Setting the values to `deplyCR=true`  `ofedDriver.deply=false`  `nvPeerDriver.deply=false` `devicePlugin.deploy=false`  `secondaryNetwork.deploy=false`, will cause install/upgrade to fail

```
$ helm install --devel -n network-operator network-operator ./
Error: unable to build kubernetes objects from release manifest: error validating "": error validating data: unknown object type "nil" in NicClusterPolicy.spec
```

Mmduh-483:
When set deployCR helm will render the NicClusterPolicy, when all components are unset the NicClusterPolicy spec will be empty which fails the deployment.

This PR make helm ensure that there is at least one component is required to deploy NicClusterPolicy

fixes #99

### DIFF
--- a/deployment/network-operator/templates/mellanox.com_v1alpha1_nicclusterpolicy_cr.yaml
+++ b/deployment/network-operator/templates/mellanox.com_v1alpha1_nicclusterpolicy_cr.yaml
@@ -13,7 +13,7 @@
   See the License for the specific language governing permissions and
   limitations under the License.
 */}}
-{{- if .Values.deployCR }}
+  {{- if and (or .Values.ofedDriver.deploy .Values.nvPeerDriver.deploy .Values.devicePlugin.deploy (and .Values.secondaryNetwork.deploy (or .Values.secondaryNetwork.cniPlugins.deploy .Values.secondaryNetwork.multus.deploy .Values.secondaryNetwork.ipamPlugin.deploy))) .Values.deployCR }}
 apiVersion: mellanox.com/v1alpha1
 kind: NicClusterPolicy
 metadata:


deployment/network-operator/templates/mellanox.com_v1alpha1_nicclusterpolicy_cr.yaml
```diff
@@ -13,7 +13,7 @@
   See the License for the specific language governing permissions and
   limitations under the License.
 */}}
-{{- if .Values.deployCR }}
+  {{- if and (or .Values.ofedDriver.deploy .Values.nvPeerDriver.deploy .Values.devicePlugin.deploy (and .Values.secondaryNetwork.deploy (or .Values.secondaryNetwork.cniPlugins.deploy .Values.secondaryNetwork.multus.deploy .Values.secondaryNetwork.ipamPlugin.deploy))) .Values.deployCR }}
```
moshe010, line 16: this look very complicate if which will be hard to maintain. 
Maybe we should deploy the CRD even if everything is false. 
will that work?

moshe010 (changes requested):
the commit message title is too long

deployment/network-operator/templates/mellanox.com_v1alpha1_nicclusterpolicy_cr.yaml
Mmduh-483: The spec is empty which cause an error

adrianchiris:
@Mmduh-483  PTAL at these commit message guidlines : https://www.kubernetes.dev/docs/guide/pull-requests/#commit-message-guidelines

deployment/network-operator/templates/mellanox.com_v1alpha1_nicclusterpolicy_cr.yaml
adrianchiris: IIRC, K8s will fail creating a CR with an empty spec even if the  object schema allows it.

we could define a "dummy" field but i dont see an added value except solving the problem above.

perhaps add this to [_helpers.tpl](https://github.com/Mellanox/network-operator/blob/master/deployment/network-operator/templates/_helpers.tpl) with comment  and use it here so the template is kept clean ?

we would still need to maintain though...

moshe010:
We will see how to address validation in generic way maybe using schema

Issue #105, opened 2021-02-22
[helm chart] Images configuration


e0ne:
In different production and dev/test deployments it's useful to be able to configure following attributes for each image:

- imagePullPolicy 
- allow to use private image registry

PR #106, opened 2021-02-22
Add ability to specify imagePullSecrets for Network Operator image

e0ne:
We need to imagePullSecrets while we use private or pulic image
registry which requires authorization.

Closes: #105


adrianchiris:
No need for `imagePullSecrets` to propagate to the manifests ?

you may want to use a custom image for multus, device plugin, whereabouts etc

e0ne:
@adrianchiris you're absolutely right. I was going to do it in a separate commit/PR, that's why I put 'Related issue' comment instead of 'Closes issue'

adrianchiris:
> @adrianchiris you're absolutely right. I was going to do it in a separate commit/PR, that's why I put 'Related issue' comment instead of 'Closes issue'

ack, so better to cover it in one PR IMO (different commit) as its all related.

config/crd/bases/mellanox.com_nicclusterpolicies.yaml
```diff
@@ -138,15 +150,19 @@ spec:
                 - version
                 type: object
               secondaryNetwork:
-                description: SecondaryNetwork describes configuration options for
-                  secondary network
+                description: SecondaryNetworkSpec describes configuration options
```
moshe010, line 153: it called secondaryNetwork so why you change it to SecondaryNetworkSpec?  maybe we can just remove it and write "describes configuration options .."

e0ne: Need to find a proper fix to make both kubebuilder and godoc happy

deployment/network-operator/README.md
```diff
@@ -167,7 +169,8 @@ Production cluster environment can deny direct access to the Internet and instea
 | `nvPeerDriver.deploy` | bool | `false` | deploy NVIDIA Peer memory driver container |
 | `nvPeerDriver.repository` | string | `mellanox` | NVIDIA Peer memory driver image repository |
 | `nvPeerDriver.image` | string | `nv-peer-mem-driver` | NVIDIA Peer memory driver image name  |
-| `nvPeerDriver.version` | string | `1.0-9` | Mellanox OFED driver version  |
+| `nvPeerDriver.version` | string | `1.0-9` | NVIDIA Peer memory driver version  |
```
moshe010, line 172: un releated to this PR but we update the nv peer mem to 1.1-0

moshe010 (approved):

PR #107, opened 2021-02-23
Update sriov-network-operator crd and images

Mmduh-483:
Update sriov-network-operator crd to support switchdev and update operator and daemon images to be from harbor.mellanox

### DIFF
--- a/deployment/network-operator/charts/sriov-network-operator/values.yaml
+++ b/deployment/network-operator/charts/sriov-network-operator/values.yaml
@@ -42,8 +42,8 @@ resources: {}
 
 # Image URIs for sriov-network-operator components
 images:
-  operator: quay.io/openshift/origin-sriov-network-operator:4.8
-  sriovConfigDaemon: quay.io/openshift/origin-sriov-network-config-daemon:4.8
+  operator: mellanox/sriov-operator:4.8
+  sriovConfigDaemon: mellanox/sriov-operator-daemon:4.8
   sriovCni: nfvpe/sriov-cni:v2.5
   ibSriovCni: mellanox/ib-sriov-cni:v1.0.0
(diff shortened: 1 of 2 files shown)


deployment/network-operator/charts/sriov-network-operator/values.yaml
```diff
@@ -42,8 +42,8 @@ resources: {}
 
 # Image URIs for sriov-network-operator components
 images:
-  operator: quay.io/openshift/origin-sriov-network-operator:4.8
-  sriovConfigDaemon: quay.io/openshift/origin-sriov-network-config-daemon:4.8
+  operator: harbor.mellanox.com/cloud-orchestration/sriov-operator:latest
```
e0ne, line 45: harbor.mellanox.com is available only inside Mellanox network. Can we use public registry here?

e0ne (changes requested):

deployment/network-operator/charts/sriov-network-operator/values.yaml
Mmduh-483: You have a point
@moshe010 should I change it to mellanox dockerhub

moshe010: yes let change to dockerhub

moshe010: also can you tag the images and not use latest

moshe010 (changes requested):

deployment/network-operator/charts/sriov-network-operator/values.yaml
Mmduh-483: updated

moshe010 (approved):

PR #108, opened 2021-02-23
Update node-feature-discovery deployment

e0ne:
1. Use helm chart updated from NFD repo
2. network-operator specifies the latest build of the NFD
3. nfd-master '--instance' option is set to 'nvidia.networking' be default

### DIFF
--- a/deployment/network-operator/charts/node-feature-discovery/templates/clusterrole.yaml
+++ b/deployment/network-operator/charts/node-feature-discovery/templates/clusterrole.yaml
@@ -15,7 +15,7 @@
 */}}
 
 {{- if .Values.rbac.create }}
-apiVersion: rbac.authorization.k8s.io/{{ .Values.rbac.apiVersion }}
+apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
   name: {{ include "node-feature-discovery.fullname" . }}
@@ -26,8 +26,12 @@ rules:
   - ""
   resources:
   - nodes
+  # when using command line flag --resource-labels to create extended resources
+  # you will need to uncomment "- nodes/status"
+  # - nodes/status
   verbs:
   - get
   - patch
   - update
+  - list
 {{- end }}
--- a/deployment/network-operator/charts/node-feature-discovery/templates/master.yaml
+++ b/deployment/network-operator/charts/node-feature-discovery/templates/master.yaml
@@ -61,9 +61,9 @@ spec:
           resources:
             {{- toYaml .Values.master.resources | nindent 12 }}
           args:
-            { { - if .Values.master.instance | empty | not } }
+            {{- if .Values.master.instance | empty | not }}
             - "--instance={{ .Values.master.instance }}"
-            { { - end } }
+            {{- end }}
 ## Enable TLS authentication
 ## The example below assumes having the root certificate named ca.crt stored in
 ## a ConfigMap named nfd-ca-cert, and, the TLS authentication credentials stored
--- a/deployment/network-operator/charts/node-feature-discovery/templates/master.yaml
+++ b/deployment/network-operator/charts/node-feature-discovery/templates/master.yaml
@@ -46,7 +46,7 @@ spec:
         - name: master
           securityContext:
             {{- toYaml .Values.master.securityContext | nindent 12 }}
-          image: "{{ .Values.image.repository }}:{{ .Chart.AppVersion }}"
+          image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
           imagePullPolicy: {{ .Values.image.pullPolicy }}
           ports:
           - containerPort: 8080
@@ -58,9 +58,12 @@ spec:
                 fieldPath: spec.nodeName
           command:
             - "nfd-master"
-          
           resources:
             {{- toYaml .Values.master.resources | nindent 12 }}
+          args:
+            { { - if .Values.master.instance | empty | not } }
+            - "--instance={{ .Values.master.instance }}"
+            { { - end } }
 ## Enable TLS authentication
 ## The example below assumes having the root certificate named ca.crt stored in
 ## a ConfigMap named nfd-ca-cert, and, the TLS authentication credentials stored
@@ -68,7 +71,6 @@ spec:
 ## Additional hardening can be enabled by specifying --verify-node-name in
 ## args, in which case every nfd-worker requires a individual node-specific
 ## TLS certificate.
-#          args:
 #            - "--ca-file=/etc/kubernetes/node-feature-discovery/trust/ca.crt"
 #            - "--key-file=/etc/kubernetes/node-feature-discovery/certs/tls.key"
 #            - "--cert-file=/etc/kubernetes/node-feature-discovery/certs/tls.crt"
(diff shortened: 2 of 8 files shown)


deployment/network-operator/charts/node-feature-discovery/templates/master.yaml
```diff
@@ -58,17 +58,19 @@ spec:
                 fieldPath: spec.nodeName
           command:
             - "nfd-master"
-          
           resources:
             {{- toYaml .Values.master.resources | nindent 12 }}
+          args:
+            { { - if .Values.master.instance | empty | not } }
+            - "--instance={{ .Values.master.instance }}"
```
moshe010, line 65: I don't think we want it to be configurable. you can hardcode the value nvidia.networking or something else

deployment/network-operator/charts/node-feature-discovery/templates/nfd-worker-conf.yaml
```diff
@@ -0,0 +1,25 @@
+{{/*
+   Copyright 2020 NVIDIA
```
moshe010, line 2: s/2020/2021

deployment/network-operator/values.yaml
```diff
@@ -24,6 +24,11 @@ sriovNetworkOperator:
 
 # Node Feature discovery chart related values
 node-feature-discovery:
+  image:
+    repository: gcr.io/k8s-staging-nfd/node-feature-discovery
+    tag: v0.6.0-233-g3e00bfb
```
moshe010, line 29: can we do it deafult in the app vesion

deployment/network-operator/values.yaml
```diff
@@ -24,6 +24,11 @@ sriovNetworkOperator:
...
+  master:
+    instance: "network-operator"
```
moshe010, line 31: let remove this to be hard coded

moshe010 (changes requested):

deployment/network-operator/charts/node-feature-discovery/templates/clusterrole.yaml
```diff
@@ -15,7 +15,7 @@
 */}}
 
 {{- if .Values.rbac.create }}
-apiVersion: rbac.authorization.k8s.io/{{ .Values.rbac.apiVersion }}
+apiVersion: rbac.authorization.k8s.io/v1
```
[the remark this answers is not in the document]
e0ne: rbac.authorization.k8s.io/v1beta1 ClusterRoleBinding is deprecated in v1.17+, unavailable in v1.22+, so only stable API is supported in NFD

deployment/network-operator/charts/node-feature-discovery/templates/master.yaml
```diff
@@ -46,7 +46,7 @@ spec:
         - name: master
           securityContext:
             {{- toYaml .Values.master.securityContext | nindent 12 }}
-          image: "{{ .Values.image.repository }}:{{ .Chart.AppVersion }}"
+          image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
```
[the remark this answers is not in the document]
e0ne: Let's keep it as exists in NFD but hardcode version

deployment/network-operator/charts/node-feature-discovery/templates/master.yaml
e0ne: Let's hardcode it in values

deployment/network-operator/values.yaml
e0ne: One we'll start using chart from NFD helm repository, we won't be able to hardcode it without network-operator and/or admission controller changes, so I)'m not sure that it's a good idea to do it

AbdYsn:
logs can be found [here](http://13.74.249.42/nic_operator_helm-ci/93/). it seems the pod was not labeled with pci-15b3.present or something similar

deployment/network-operator/charts/node-feature-discovery/templates/worker.yaml
```diff
@@ -39,30 +39,26 @@ spec:
...
+        resources:
+        {{- toYaml .Values.worker.resources | nindent 12 }}
         command:
         - "nfd-worker"
-        resources:
-          {{- toYaml .Values.worker.resources | nindent 12 }}
         args:
         - "--sleep-interval=60s"
         - "--server={{ include "node-feature-discovery.fullname" . }}-master:{{ .Values.master.service.port }}"
-        {{- with .Values.worker.options }}
```
e0ne, line 63: It should be restored

deployment/network-operator/charts/node-feature-discovery/templates/master.yaml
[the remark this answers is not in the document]
adrianchiris: > Let's keep it as exists in NFD but hardcode version

+1 
ideally no changes to the chart itself but only override values in network-operator values.yaml file

deployment/network-operator/charts/node-feature-discovery/templates/nfd-worker-conf.yaml
```diff
@@ -0,0 +1,25 @@
...
+
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ .Values.worker.configmapName }}
+  labels:
+  {{- include "node-feature-discovery.labels" . | nindent 4 }}
+data:
+  nfd-worker.conf: |
+  {{ .Values.worker.config | indent 4 }}
```
adrianchiris, line 25: we should actually use the configmap and  put our configs here and not under options WDYT ?

i.e override this value from the top level values.yaml in network-operator

deployment/network-operator/charts/node-feature-discovery/values.yaml
```diff
@@ -17,8 +17,11 @@
 # Declare variables to be passed into your templates.
 
 image:
-  repository: quay.io/kubernetes_incubator/node-feature-discovery
-  pullPolicy: IfNotPresent
+  repository: gcr.io/k8s-staging-nfd/node-feature-discovery
+  # This should be set to 'IfNotPresent' for released version
+  pullPolicy: Always
```
adrianchiris, line 22: change in top level values.yaml to IfNotPresent ?

adrianchiris (commented):
@e0ne i expect the nfd chart itself to be identical to what was merged to NFD with the exception of version override in `node-feature-discovery/Chart.yaml`

is that the case ?

In addition

deployment/network-operator/charts/node-feature-discovery/templates/nfd-worker-conf.yaml
e0ne: Good catch. Let me try it

deployment/network-operator/values.yaml
```diff
@@ -34,15 +38,16 @@ node-feature-discovery:
...
+    config: |
       sources:
-        pci:
-          deviceClassWhitelist:
-            - "02"
-            - "0200"
-            - "0207"
-          deviceLabelFields:
-            - vendor
+          pci:
```
e0ne, line 43: An indentation looks to big but it works as expected

adrianchiris:
In regards to commit message:
This is actually the first time we use HELM chart from NFD, so commit message should reflect that.

deployment/network-operator/values.yaml
moshe010: @e0ne  please open an issue in github and in our RM for this. 
I will merged as it has it doesn't effect functionality

moshe010 (approved):

PR #113, opened 2021-02-25
update charts to v0.3.0

moshe010:
update network operator app to v0.3.0
update network operator chart to v0.3.0
update nfd chart to v0.2.0
update Readme nfd to v0.6.0-233-g3e00bfb

### DIFF
--- a/deployment/network-operator/Chart.yaml
+++ b/deployment/network-operator/Chart.yaml
@@ -1,8 +1,8 @@
 apiVersion: v2
 name: network-operator
-version: 0.2.0-beta
+version: 0.3.0
 kubeVersion: '>= 1.17.0'
-appVersion: v0.2.0
+appVersion: v0.3.0
 description: Nvidia Mellanox network operator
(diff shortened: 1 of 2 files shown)


e0ne (commented):
Why do we need to bump NFD chart to 0.2.0?

e0ne (approved):

deployment/network-operator/Chart.yaml
```diff
@@ -15,12 +15,9 @@ dependencies:
 - condition: nfd.enabled
   name: node-feature-discovery
   repository: ""
-  version: 0.1.0
+  version: 0.2.0
```
adrianchiris, line 18: we should keep the same chart version as in NFD repo IMO

adrianchiris (commented):
multus CNI default version under `chart parameters` should be updated to v3.6.
in _deployment/network-operator/README.md_

adrianchiris:
#117

PR #114, opened 2021-02-26
Cleanup pkg/utils/utils.go

e0ne:
- functin IsCRD is defined but not used
- function GetEnv just a wrapper of the os.Getenv


moshe010 (approved):
LGTM. just waiting for CI to Pass

moshe010 (approved):

PR #128, opened 2021-03-01
Release Network-Operator v0.3.0

moshe010:


e0ne (approved):

adrianchiris (changes requested):
Looking at `index.yaml`, i think we should avoid re-creating old entires.

we should use : `>  > helm repo index . --url https://mellanox.github.io/network-operator/release --merge`
you mind re-creating this file using old index and --merge option ?

i will submit PR to fix templates.

adrianchiris:
Hmm, let me check that first. ill report back

adrianchiris (approved):

PR #135, opened 2021-03-10
Rename 'devicePlugin' to 'rdmaSharedDevicePlugin'

e0ne:
Since we will allow configure and install SR-IOV Network Device Plugin
we need to have more user-friendly name to not confuse users with a
'devicePlugin' option.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ CRD that defines a Cluster state for Mellanox Network devices.
 #### NICClusterPolicy spec:
 NICClusterPolicy CRD Spec includes the following sub-states/stages:
 - `ofedDriver`: [OFED driver container](https://github.com/Mellanox/ofed-docker) to be deployed on Mellanox supporting nodes.
-- `devicePlugin`: [RDMA shared device plugin](https://github.com/Mellanox/k8s-rdma-shared-dev-plugin)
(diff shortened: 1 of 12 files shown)


deployment/network-operator/templates/tests/test-dev-plugin.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if and .Values.deployCR .Values.devicePlugin.deploy  }}
+{{- if and .Values.deployCR .Values.rdmaSharedDevicePlugin.deploy  }}
```
adrianchiris, line 1: a comment about the name of the file, should we change to `test-rdma-shared-dev-plugin.yaml` ?
im fine either way, but thought its worth mentioning

adrianchiris (approved):

deployment/network-operator/templates/tests/test-dev-plugin.yaml
e0ne: Good catch, Adrian! I'll rename it once CI reports on it

AbdYsn:
@e0ne CI is passing with mellanox/kubernetes-ci#93

AbdYsn:
@moshe010 we implemented a new mechanism in jenkins to be able to checkout ci patchs to test api breaking patchs. it worked in this pr using `depends-on: kubernetes-ci#93` in the commit message.

moshe010 (approved):

PR #144, opened 2021-03-18
[helm] fix nfd worker tolerations value

moshe010:
the tolerations value is array therefore we need to pass
[] and not {}


e0ne:
@moshe010  Thanks for fixing it here and in NFD repository

Issue #133, opened 2021-03-09
Stop using SelfLink


e0ne:
**What would you like to be added**:
We need to add some link to the NetworkAttachDefinition in MacvlanNetwork. SelfLink can't be used anymore, so we have to implement some workaround like https://github.com/lensapp/lens/pull/1804/files

**Why is this needed**:
This was deprecated in Kubernetes v1.16 and will be removed in v.1.21. SelfLink is disabled by default in v.1.20: https://github.com/kubernetes/enhancements/issues/1164

PR #145, opened 2021-03-19
Stop using SelfLink

e0ne:
Objects SelfLink was deprecated in Kubernetes v1.16 and will be
removed in v.1.21. SelfLink is disabled by default in v.1.20.

This implementation is a simplified version of the original one [1]
and uses public API which isn't supposed to be changed soon.

This patch also moves MacvlanNetwork status update to controller.

[1] https://github.com/kubernetes/kubernetes/blob/54449be03127f883e52df48b95adef36925ae2cf/staging/src/k8s.io/apiserver/pkg/endpoints/handlers/namer.go#L105

Closes: #133
Closes: #59

### DIFF
--- a/controllers/macvlannetwork_controller.go
+++ b/controllers/macvlannetwork_controller.go
@@ -23,16 +23,20 @@ import (
 	"github.com/go-logr/logr"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 
+	netattdefv1 "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
+
 	mellanoxcomv1alpha1 "github.com/Mellanox/network-operator/api/v1alpha1"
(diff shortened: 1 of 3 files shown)


pkg/utils/utils.go
```diff
@@ -53,3 +56,9 @@ func GetFilesWithSuffix(baseDir string, suffixes ...string) ([]string, error) {
 	}
 	return files, nil
 }
+
+func GetNetworkAttachmentDefLink(netAttDef *netattdefv1.NetworkAttachmentDefinition) string {
+	link := fmt.Sprintf("%s/namespaces/%s/%s/%s",
```
moshe010, line 61: instead of link := you can just return
another option is to define in the signature
func GetNetworkAttachmentDefLink(netAttDef *netattdefv1.NetworkAttachmentDefinition) (link string)
{
link = fmt.Sprintf("%s/namespaces ...
}

e0ne: @moshe010 could you please add more details on  this option?

> another option is to define in the signature

moshe010: > @moshe010 could you please add more details on this option?
> 
> > another option is to define in the signature
sorry you still need to return. It called Named return values
https://tour.golang.org/basics/7

e0ne: Thank you, I'll update a PR

moshe010 (approved):

controllers/macvlannetwork_controller.go
```diff
@@ -89,10 +93,23 @@ func (r *MacvlanNetworkReconciler) updateCrStatus(cr *mellanoxcomv1alpha1.Macvla
 		cr.Status.Reason = syncError.Error()
 	}
 
+	netAttachDef := &netattdefv1.NetworkAttachmentDefinition{}
```
adrianchiris, line 96: I wonder if we should conditionally do this bit only if cr.Status is "Ready", else set the net-attach-def self link to empty.

This will save some un-needed API Calls.

Issue #134, opened 2021-03-09
CRDs are not synced 


e0ne:
**What happened**:
CRDs are not synced in these folders:

- https://github.com/Mellanox/network-operator/tree/master/deployment/network-operator/crds
- https://github.com/Mellanox/network-operator/tree/master/config/crd/bases

**What you expected to happen**:
CRDs should be the same for helm chart and manual operator deployment. We should sync them and create some CI checkto not introduce this bug in the future.

moshe010:
It seem partial-closes doesn't works in github. I will reopen this

e0ne:
Partial-closed added because there is no automation for this

PR #146, opened 2021-03-19
Sync CRDs between operator and helm chart

e0ne:
CRDs are synced automatically and CI check is added.

Closes: #134

### DIFF
--- a/scripts/check-manifests.sh
+++ b/scripts/check-manifests.sh
@@ -1,0 +1,24 @@
+#!/bin/bash
+# Copyright 2021 NVIDIA
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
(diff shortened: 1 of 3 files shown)


scripts/check-manifests.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+# Copyright 2020 NVIDIA
```
moshe010, line 2: s/2020/2021

e0ne: thanks, will fix it once I'll figure out why it doesn't work on CI

moshe010:
see like the CRD are not sync

e0ne:
> see like the CRD are not sync see The command "bash scripts/check-manifests.sh" exited with 1.

I need to make this script more accurate, it checks all files in a repo and I can't reproduce the issue locally

e0ne:
Travis CI failure is not related to the change

go.mod
```diff
@@ -7,6 +7,7 @@ require (
 	github.com/go-logr/logr v0.3.0
 	github.com/googleapis/gnostic v0.5.3 // indirect
 	github.com/k8snetworkplumbingwg/network-attachment-definition-client v1.1.0
+	github.com/mattn/goveralls v0.0.8 // indirect
```
moshe010, line 10: why you need goveralls? We don't integrated  Coveralls.io  in network operator at the moment

moshe010 (changes requested):

go.mod
e0ne: Need to investigate why Travis CI add it. Let's remove it for now

moshe010 (approved):

PR #147, opened 2021-03-21
update multus/sriov-network-operator/whereabouts-cni to k8snetworkplumbingwg

moshe010:


e0ne (changes requested):
Do we want to change the images that we use too?

moshe010:
> Do we want to change the images that we use too?

We don't but it will be separate PR

moshe010:
I will put different PR and close this one

PR #151, opened 2021-03-25
Fix error in lint-helm target in Makfile

ykulazhenkov:
Change helm installation method. Use get_helm.sh script instead of downloading a tarball.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -65,6 +65,7 @@ HADOLINT = $(TOOLSDIR)/hadolint
 HADOLINT_VER = v1.23.0
 
 HELM = $(TOOLSDIR)/helm
+GET_HELM = $(TOOLSDIR)/get_helm.sh
 HELM_VER = v3.5.3
 
 TIMEOUT = 15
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -124,7 +124,7 @@ $(HADOLINT): | $(TOOLSDIR) ; $(info  install hadolint...)
 
 $(HELM): | $(TOOLSDIR) ; $(info  install helm...)
 	$Q curl -sSfL https://get.helm.sh/helm-$(HELM_VER)-linux-amd64.tar.gz -o - | tar xz -C $(TOOLSDIR) --strip-components=1 linux-amd64/helm
```
moshe010, line 126: why not install helm with script?

$ curl -fsSL -o get_helm.sh https://raw.githubusercontent.com/helm/helm/master/scripts/get-helm-3
$ chmod 700 get_helm.sh
$ ./get_helm.sh

moshe010: also how did the first commit passed CI?

ykulazhenkov: The main reason why I decided to use tarball-based installation method is to prevent installing helm globally. My personal preference is to install all supplementary tools for project building/linting to the local path. 
I checked again, and it is possible to use `get_helm.sh` for installation to a local directory, so I updated my PR to use this method.

> also how did the first commit passed CI?
Execution bit is already set on the helm binary after it unpacked from the tarball.  The original version of `make lint-helm` was setting exec bit on the wrong binary (hadolint), but that didn't prevent helm installation from working if that wrong binary exists. Hadolint binary was already in place because `make lint-dockerfile` was called before.

e0ne (approved):

moshe010 (approved):

PR #152, opened 2021-03-26
Add additional nodeAffinity for nv-peer-mem-driver POD

ykulazhenkov:
We use `nvidia.com/gpu.present: "true"` label to select nodes on which we should deploy nv-peer-mem-driver POD. NFD manages this label, and it set label based on GPU PCI device presence. With current network-operator behavior, we can deploy `nv-peer-mem-driver` POD on nodes with no GPU driver, which will cause a crash of `nv-peer-mem-driver` POD.

The recent version of GPU operator put 'nvidia.com/gpu.count' label
on a node after initialization of GPU driver finished on it.
We can understand that node has GPU and GPU driver already deployed
on it by checking this label.

### DIFF
--- a/manifests/stage-nv-peer-mem-driver/0010_nv-peer-mem-driver-ds.yaml
+++ b/manifests/stage-nv-peer-mem-driver/0010_nv-peer-mem-driver-ds.yaml
@@ -99,3 +99,12 @@ spec:
       nodeSelector:
         feature.node.kubernetes.io/pci-15b3.present: "true"
         nvidia.com/gpu.present: "true"
+      affinity:
+        nodeAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+            - matchExpressions:
+              - key: nvidia.com/gpu.count
+                operator: Gt
+                values:
+                  - "0"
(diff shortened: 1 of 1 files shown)


manifests/stage-nv-peer-mem-driver/0010_nv-peer-mem-driver-ds.yaml
```diff
@@ -99,3 +99,12 @@ spec:
       nodeSelector:
         feature.node.kubernetes.io/pci-15b3.present: "true"
         nvidia.com/gpu.present: "true"
+      affinity:
+        nodeAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
```
moshe010, line 104: maybe we can also add init-container with the following validation for he nvidia driver [1] 
This is how the GPU operator does it 

[1] - https://github.com/NVIDIA/gpu-operator/blob/master/assets/state-container-toolkit/0400_container_toolkit.yml#L26-L34

ykulazhenkov: Yes, I agree. Such init container can be a good addition. I updated the PR.

moshe010 (approved):

moshe010:
LGTM waiting for CI to pass

e0ne (approved):

moshe010 (approved):

PR #155, opened 2021-03-30
Add support for host OFED to peer-mem POD

### LINKED ISSUE #76: nv-peer-mem driver container not starting if used with host OFED
### Issue
Deploying the nv-peer-mem driver alone paired with the host OFED fails to start because the Mellanox driver resources directory in case of host OFED is different from the one used with the ofed-driver container.

### Possible solution
Add a configuration to change the run-mlnx-ofed volume so that it can be changed if used with the host OFED.

ykulazhenkov:
Autoselect right path of OFED src for nv-peer-mem POD

Based on the fact that Mellanox OFED driver POD is deployed or not,
network-operator will select the right OFED src path for nv-peer-mem POD.
If OFED driver POD doesn't exist, OFED from the host will be used.

~~Add additional option `nvPeerDriver.ofedDriverSourcePath` to
NicClusterPolicy CR. Default value points to the location
in which `Mellanox OFED driver` POD will install the driver.
If the option set to an empty string, then Mellanox OFED driver
from the host OS will be used~~

Fixes #76

### DIFF
--- a/manifests/stage-nv-peer-mem-driver/0010_nv-peer-mem-driver-ds.yaml
+++ b/manifests/stage-nv-peer-mem-driver/0010_nv-peer-mem-driver-ds.yaml
@@ -78,8 +78,11 @@ spec:
             - name: NO_PROXY
               value: {{ .RuntimeSpec.NoProxy }}
           volumeMounts:
-            - name: run-mlnx-ofed
-              mountPath: /run/mellanox/drivers
+            - name: mlnx-ofed-usr-src
+              mountPath: /run/mellanox/drivers/usr/src
+              mountPropagation: HostToContainer
+            - name: mlnx-ofed-lib-modules
+              mountPath: /run/mellanox/drivers/lib/modules
               mountPropagation: HostToContainer
             - name: run-nvidia
               mountPath: /run/nvidia/drivers
@@ -110,11 +113,14 @@ spec:
       volumes:
         {{ $ofedPath := "/run/mellanox/drivers" }}
         {{ if .RuntimeSpec.UseHostOFED }}
-          {{ $ofedPath = "/" }}
+          {{ $ofedPath = "" }}
         {{end}}
-        - name: run-mlnx-ofed
+        - name: mlnx-ofed-usr-src
           hostPath:
-            path: {{ $ofedPath }}
+            path: {{ $ofedPath }}/usr/src
+        - name: mlnx-ofed-lib-modules
+          hostPath:
+            path: {{ $ofedPath }}/lib/modules
         - name: run-nvidia
           hostPath:
             path: {{ .CrSpec.GPUDriverSourcePath }}
--- a/manifests/stage-nv-peer-mem-driver/0010_nv-peer-mem-driver-ds.yaml
+++ b/manifests/stage-nv-peer-mem-driver/0010_nv-peer-mem-driver-ds.yaml
@@ -108,9 +108,13 @@ spec:
             initialDelaySeconds: 10
             failureThreshold: 1
       volumes:
+        {{ $ofedPath := "/run/mellanox/drivers" }}
+        {{ if .RuntimeSpec.UseHostOFED }}
+          {{ $ofedPath = "/" }}
+        {{end}}
         - name: run-mlnx-ofed
           hostPath:
-            path: /run/mellanox/drivers
+            path: {{ $ofedPath }}
         - name: run-nvidia
           hostPath:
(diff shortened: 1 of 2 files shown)


deployment/network-operator/templates/mellanox.com_v1alpha1_nicclusterpolicy_cr.yaml
```diff
@@ -31,6 +31,7 @@ spec:
     repository: {{ .Values.nvPeerDriver.repository }}
     version: {{ .Values.nvPeerDriver.version }}
     gpuDriverSourcePath: {{ .Values.nvPeerDriver.gpuDriverSourcePath }}
+    ofedDriverSourcePath: {{ .Values.nvPeerDriver.ofedDriverSourcePath }}
```
moshe010, line 34: you already have the .Values.ofedDriver.deploy which indicate if we deploy OFED so why we need this options as well?

manifests/stage-nv-peer-mem-driver/0010_nv-peer-mem-driver-ds.yaml
```diff
@@ -78,8 +78,11 @@ spec:
             - name: NO_PROXY
               value: {{ .RuntimeSpec.NoProxy }}
           volumeMounts:
-            - name: run-mlnx-ofed
-              mountPath: /run/mellanox/drivers
+            - name: ofed-usr-src
```
moshe010, line 81: you alway mount the OFED on host?

deployment/network-operator/templates/mellanox.com_v1alpha1_nicclusterpolicy_cr.yaml
ykulazhenkov: I agree. We can remove this option from helm values. That is not what we want to expose right now. I will update PR

manifests/stage-nv-peer-mem-driver/0010_nv-peer-mem-driver-ds.yaml
ykulazhenkov: Could you please clarify what you mean? We always mount driver sources to `/run/mellanox/drivers/` within nvPeerDriver POD. If we want to use the driver from the host OS we mount:
HOST `/usr/src` -> nv-peer-mem POD `/run/mellanox/drivers/usr/src`
HOST `/lib/modules` -> nv-peer-mem POD `/run/mellanox/drivers/lib/modules`
If we want to use driver deployed by `Mellanox OFED` POD (it deployes driver to /run/mellanox/drivers path on host) we mount:
HOST `/run/mellanox/drivers/usr/src` -> nv-peer-mem POD `/run/mellanox/drivers/usr/src`
HOST `/run/mellanox/drivers/lib/modules` -> nv-peer-mem POD `/run/mellanox/drivers/lib/modules`

moshe010: sorry this change it not relevant for the fix it accurate mount point right? can you do it in different PR or commit ?

manifests/stage-nv-peer-mem-driver/0010_nv-peer-mem-driver-ds.yaml
```diff
@@ -108,9 +111,12 @@ spec:
...
             failureThreshold: 1
       volumes:
-        - name: run-mlnx-ofed
+        - name: ofed-usr-src
+          hostPath:
+            path: {{ .CrSpec.OFEDDriverSourcePath }}/usr/src
+        - name: ofed-lib-modules
           hostPath:
-            path: /run/mellanox/drivers
+            path: {{ .CrSpec.OFEDDriverSourcePath }}/lib/modules
```
moshe010, line 119: so like we said you can know what should be the value per of OFED deploy flag

manifests/stage-nv-peer-mem-driver/0010_nv-peer-mem-driver-ds.yaml
ykulazhenkov: I moved this change to separate commit

manifests/stage-nv-peer-mem-driver/0010_nv-peer-mem-driver-ds.yaml
ykulazhenkov: done

moshe010 (approved):

Issue #111, opened 2021-02-25
Fix NFD worker configuration indentation 


e0ne:
An indentation in values.yaml [1] has additional spaces. It should be aligned  and work as expected

[1] https://github.com/Mellanox/network-operator/blob/2dcf59f17d737b11632da2f41ed640aaf8323ee5/deployment/network-operator/values.yaml#L43

Issue #68, opened 2020-12-10
use NFD helm charts from NFD repo


adrianchiris:
Apparently there are charts at:

https://github.com/k8s-at-home/charts/tree/master/charts/node-feature-discovery

perhaps we should use-them, 

Should ping nfd folks ask if thats the official place they recommend to take charts from  before swapping

adrianchiris:
Update: ive pushed a PR to kubernetes-sigs/node-feature-discovery  : https://github.com/kubernetes-sigs/node-feature-discovery/pull/423

PR #160, opened 2021-04-02
Use upstream NFD chart

e0ne:
NFD charts are merged to NFD repo now so we don't need to duplicate
effort of maintaining charts.

Closes: #68
Closes: #111

depends-on: kubernetes-ci#104


moshe010 (commented):
Why we still need the folder  under deployment/network-operator. I thought all the point that we will have one chart in nfd isn't it?

e0ne:
CI passed. I'll remove WIP once https://github.com/Mellanox/kubernetes-ci/pull/104 will be merged to not break other PRs

e0ne:
Need to add minimum helm version

Chart.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: v2
+name: network-operator
+version: 0.4.0
+kubeVersion: '>= 1.17.0'
+appVersion: v0.4.0
+description: Nvidia Mellanox network operator
```
moshe010, line 6: you can remove the Mellanox

Chart.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: v2
```
moshe010, line 1: why we need a new Chart.yaml. Don't we have this already in https://github.com/Mellanox/network-operator/blob/master/deployment/network-operator/Chart.yaml

e0ne: It was a part of WIP patch. I'll remove it

moshe010 (approved):

PR #166, opened 2021-04-12
Increase initialDelaySeconds for MOFED POD liveness probe

ykulazhenkov:
Before version 1.21, Kubernetes used startupProbe for "fresh" POD
starts and restarts (caused by liveness check failed as an example).
Starting from v1.21, startupProbe applied to "fresh" starts only.
To prevent a crash loop after MOFED POD restarts, we should grant
enough time to POD to fully boot before we start liveness checking.

### DIFF
--- a/manifests/stage-ofed-driver/0010_ofed-driver-ds.yaml
+++ b/manifests/stage-ofed-driver/0010_ofed-driver-ds.yaml
@@ -78,7 +78,10 @@ spec:
               command:
                 [sh, -c, 'lsmod | grep mlx5_core']
             periodSeconds: 30
-            initialDelaySeconds: 30
+            # starting from  v1.21, Kubernetes doesn't use startupProbe during POD restarts
+            # to prevent crash loop after POD restarts, we should grant enough time to POD to
+            # fully boot before we start liveness checking
+            initialDelaySeconds: 570
             failureThreshold: 1
             successThreshold: 1
           readinessProbe:


e0ne (approved):

manifests/stage-ofed-driver/0010_ofed-driver-ds.yaml
```diff
@@ -78,7 +78,10 @@ spec:
               command:
                 [sh, -c, 'lsmod | grep mlx5_core']
             periodSeconds: 30
-            initialDelaySeconds: 30
+            # starting from  v1.21, Kubernetes doesn't use startupProbe during POD restarts
+            # to prevent crash loop after POD restarts, we should grant enough time to POD to
+            # fully boot before we start liveness checking
+            initialDelaySeconds: 570
```
moshe010, line 84: who did you come up with this number?

ykulazhenkov: We discussed this with @AbdYsn. We use the same delay as for `startupProbe` 10 minutes. If I remember correctly, @AbdYsn mentioned that choosing `startupProbe` delays is based on experiments on multipile different ENVs.

moshe010: @ykulazhenkov  can you point me to the kubernetes commit that change the behaviour?

ykulazhenkov: I think this PR changed behavior of startupProbe https://github.com/kubernetes/kubernetes/pull/98376

ykulazhenkov: I will ask community. Maybe this is some regression in Kuberntes, because I can't find any mentions that behavior of startupProbe should change.

ykulazhenkov:
Need to check with the community that the current behavior of startupProbe is expected.

ykulazhenkov:
https://github.com/kubernetes/kubernetes/issues/101064

ykulazhenkov:
Issue in Kubernetes confirmed. Fix in progress. No need to change timeouts in network-operator repo. Closed.

Issue #167, opened 2021-04-14
SR-IOV network operator image is built for OpenShift


e0ne:
**What happened**:
SR-IOV network operator default image is supposed to be running on top of OpenShift instead of Kubernetes

**What you expected to happen**:
SR-IOV network operator should not call OpenShift speciffic API 

**Environment**:
- Network Operator v0.4.0

PR #168, opened 2021-04-14
Use Kubernetes-specific image for SR-IOV Network operator by default

e0ne:
Closes: #167

### DIFF
--- a/deployment/network-operator/charts/sriov-network-operator/values.yaml
+++ b/deployment/network-operator/charts/sriov-network-operator/values.yaml
@@ -42,7 +42,7 @@ resources: {}
 
 # Image URIs for sriov-network-operator components
 images:
-  operator: mellanox/sriov-operator:7ad0f8e
+  operator: mellanox/sriov-operator:7ad0f8e-rel1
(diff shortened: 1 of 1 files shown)


deployment/network-operator/charts/sriov-network-operator/values.yaml
```diff
@@ -42,7 +42,7 @@ resources: {}
 
 # Image URIs for sriov-network-operator components
 images:
-  operator: mellanox/sriov-operator:7ad0f8e
+  operator: mellanox/sriov-operator:7ad0f8e-k8s
```
moshe010, line 45: maybe <commit>-rel1

moshe010 (changes requested):

deployment/network-operator/charts/sriov-network-operator/values.yaml
e0ne: I don't have strong opinion on it. We can push new tag for the image. I added 'k8s' suffix to make it self-descriptive.

moshe010 (approved):

PR #170, opened 2021-04-26
Add initContainers for device plugins

e0ne:
Both SR-IOV Device Plugin and RDMA Shared Device Plugin should
wait until OFED container is started if it's supposed to be
deployed.

During cluster scale-up Kubernetes schedules DaemonSet to new nodes
and we don't configure an order of pods in this case. To wait until
MOFED containder is ready we'll add initContainer with the same logic
as MOFED container has.

This will prevent us to run device plugin before MOFED started during
scale-up but it doesn't fix an issue if node was rebooted.

### DIFF
--- a/manifests/stage-rdma-device-plugin/0060_rdma-shared-dev-plugin-ds.yaml
+++ b/manifests/stage-rdma-device-plugin/0060_rdma-shared-dev-plugin-ds.yaml
@@ -45,6 +45,14 @@ spec:
       - key: nvidia.com/gpu
         operator: Exists
         effect: NoSchedule
+{{if .DeployInitContainer}}
+      initContainers:
+        - name: ofed-driver-validation
+          image: {{ .CrSpec.Repository }}/{{ .CrSpec.Image }}:{{ .CrSpec.Version }}
+          imagePullPolicy: IfNotPresent
+          command: [ 'sh', '-c' ]
+          args: [ "until lsmod | grep mlx5_core; do echo waiting for OFED drivers to be loaded; sleep 30; done" ]
+{{end}}
       containers:
       - image: {{ .CrSpec.Repository }}/{{ .CrSpec.Image }}:{{ .CrSpec.Version }}
         name: rdma-shared-dp
--- a/manifests/stage-sriov-device-plugin/0030-sriov-dp-daemonset.yml
+++ b/manifests/stage-sriov-device-plugin/0030-sriov-dp-daemonset.yml
@@ -45,6 +45,14 @@ spec:
           operator: Exists
           effect: NoSchedule
       serviceAccountName: sriov-device-plugin
+{{if .DeployInitContainer}}
+      initContainers:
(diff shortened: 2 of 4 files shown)


manifests/stage-rdma-device-plugin/0060_rdma-shared-dev-plugin-ds.yaml
```diff
@@ -45,6 +45,14 @@ spec:
       - key: nvidia.com/gpu
         operator: Exists
         effect: NoSchedule
+{{if .DeployInitContainer}}
+      initContainers:
+        - name: ofed-driver-validation
+          image: {{ .CrSpec.Repository }}/{{ .CrSpec.Image }}:{{ .CrSpec.Version }}
+          imagePullPolicy: IfNotPresent
+          command: [ 'sh', '-c' ]
+          args: [ "until lsmod | grep mlx5_core; do echo waiting for OFED drivers to be loaded; sleep 30; done" ]
```
moshe010, line 54: I am not sure this validation is good enough? did you talk to OFED team?
don't we need to check the rdma ulp modules are loaded?

e0ne: We do the same checks in the OFED container [1]. If we need more checks, we need fix them everywhere.

[1] https://github.com/Mellanox/network-operator/blob/e73e49b99ff9d2f0eb92ab95143a9b4890f96d07/manifests/stage-ofed-driver/0050_ofed-driver-ds.yaml#L90

moshe010: I see, let first talk to ofed and see what they recommend

moshe010: by the way why we need the sleep?

e0ne: Since there are neither initialDelaySeconds nor periodSecond for init containers, we need to have some timeout between checks

e0ne: I tested init container with `test -e /run/mellanox/drivers/.driver-ready` script few dozens time. it looks less stable rather than kernel module check. e.g.: after I stopped MOFED pod I've got the following state:

```
 root@node-1:~# lsmod | grep mlx
 root@node-1:~# ls /run/mellanox/drivers/.driver-ready
 /run/mellanox/drivers/.driver-ready
```

It seems to be a MOFED issue which should be troubleshooted and fixed separately. I prefer to check for kernel module now

moshe010: ok let keep it for now and open issue in github to track it.

moshe010 (approved):

PR #172, opened 2021-04-29
Use labels to wait for MOFED pods

e0ne:
Operator will add nvidia.com/ofed.wait=true label to node,
so device plugins pods (SR-IOV Device Plugin, RDMA Shared
Device Plugin) won't be cheduled to these nodes. All running
device plugins pods will be terminated.

Once OFED reach Running state nodes will have
nvidia.com/ofed.wait=false label, to allow Kubernetes schedule
device plugins pods to these nodes.

In case if OFED deployment is disabled label
nvidia.com/ofed.wait=false will be added to the nodes.

This patch will fix an issue with pods initializing orded after
node reboot.

### DIFF
--- a/controllers/nicclusterpolicy_controller.go
+++ b/controllers/nicclusterpolicy_controller.go
@@ -22,9 +22,12 @@ import (
 	"time"
 
 	"github.com/go-logr/logr"
+	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
+	apiErrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
@@ -60,7 +63,7 @@ func (r *NicClusterPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Req
 	instance := &mellanoxv1alpha1.NicClusterPolicy{}
 	err := r.Get(context.TODO(), req.NamespacedName, instance)
 	if err != nil {
-		if errors.IsNotFound(err) {
+		if apiErrors.IsNotFound(err) {
 			// Request object not found, could have been deleted after reconcile request.
 			// Owned objects are automatically garbage collected. For additional cleanup logic use finalizers.
 			// Return and don't requeue
@@ -106,6 +109,12 @@ func (r *NicClusterPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Req
 	if err != nil {
 		return reconcile.Result{}, err
 	}
+
+	err = r.updateNodeLabels(instance)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
 	if managerStatus.Status != state.SyncStateReady {
 		return reconcile.Result{
 			RequeueAfter: time.Duration(config.FromEnv().Controller.RequeueTimeSeconds) * time.Second,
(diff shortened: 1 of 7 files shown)


controllers/nicclusterpolicy_controller.go
```diff
@@ -115,6 +122,39 @@ func (r *NicClusterPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
+		var pods corev1.PodList
+		appLabel := "mofed-" + attrs[0].Attributes[nodeinfo.AttrTypeOSName] + attrs[0].Attributes[nodeinfo.AttrTypeOSVer]
+		_ = r.Client.List(context.TODO(), &pods, client.MatchingLabels{"app": appLabel})
+		for _, pod := range pods.Items {
+			labelValue := "true"
+			if pod.Status.Phase != "Running" {
+				labelValue = "false"
+			}
+			node := &corev1.Node{}
+			_ = r.Client.Get(context.TODO(), client.ObjectKey{
```
adrianchiris, line 144: you could do it in one PATCH call

controllers/nicclusterpolicy_controller.go
```diff
@@ -115,6 +122,39 @@ func (r *NicClusterPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
+	// TODO: support case if MOFED is not configured
+	nodeInfo := infoCatalog.GetNodeInfoProvider()
+	if nodeInfo == nil {
+		return errors.New("catalog does not provide node information")
+	}
+	if cr.Spec.OFEDDriver != nil {
+		attrs := nodeInfo.GetNodesAttributes(
+			nodeinfo.NewNodeLabelFilterBuilder().WithLabel(nodeinfo.NodeLabelMlnxNIC, "true").Build())
+		var pods corev1.PodList
+		appLabel := "mofed-" + attrs[0].Attributes[nodeinfo.AttrTypeOSName] + attrs[0].Attributes[nodeinfo.AttrTypeOSVer]
```
adrianchiris, line 136: technically you could define a new label dentifying a pod as a "mofed driver container" regardless of os and arch 
(which is used for the daemonset to own these pods in case of mixed environments which we dont support atm :))

then you would not need to fetch nodes and just use the fixed label.

eg label  `driver-pod: mofed-<version>` or similar

controllers/nicclusterpolicy_controller.go
```diff
@@ -115,6 +122,39 @@ func (r *NicClusterPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
+		_ = r.Client.List(context.TODO(), &pods, client.MatchingLabels{"app": appLabel})
+		for _, pod := range pods.Items {
+			labelValue := "true"
+			if pod.Status.Phase != "Running" {
```
adrianchiris, line 140: Will a pod be in running state when the driver container is building or during driver load ?

controllers/nicclusterpolicy_controller.go
```diff
@@ -106,6 +107,12 @@ func (r *NicClusterPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Req
 	if err != nil {
 		return reconcile.Result{}, err
 	}
+
+	err = r.updateNodeLabels(instance, sc)
```
adrianchiris, line 111: A comment explaining what we are doing here would be great

controllers/nicclusterpolicy_controller.go
e0ne: Good catch, we need t check ContainerStatuses

controllers/nicclusterpolicy_controller.go
e0ne: Let's do it! It will simplify the code without any side effects

controllers/nicclusterpolicy_controller.go
e0ne: Yes, I added TODO for it. Will fix it once I resolve CI issue

pkg/nodeinfo/attributes.go
```diff
@@ -35,6 +35,7 @@ const (
 	NodeLabelCPUArch       = "kubernetes.io/arch"
 	NodeLabelMlnxNIC       = "feature.node.kubernetes.io/pci-15b3.present"
 	NodeLabelNvGPU         = "nvidia.com/gpu.present"
+	NodeLabelWaitOFED      = "nvidia.com/mofed.wait"
```
adrianchiris, line 38: I suspect we are going to have additional network-operator related labels in the future.

maybe we should go with a naming scheme that partitions  the network operator related labels.

e.g :  `network.nvidia.com/operator.mofed.wait` ? or similar

WDYT?

controllers/nicclusterpolicy_controller.go
```diff
@@ -115,6 +124,52 @@ func (r *NicClusterPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
+// Set nvidia.com/ofed.wait=false if OFED is not deployed.
+func (r *NicClusterPolicyReconciler) updateNodeLabels(cr *mellanoxv1alpha1.NicClusterPolicy) error {
+	if cr.Spec.OFEDDriver != nil {
+		pods := &corev1.PodList{}
+		podLabel := "mofed-" + cr.Spec.OFEDDriver.Version
+		_ = r.Client.List(context.TODO(), pods, client.MatchingLabels{"driver-pod": podLabel})
+		for i := range pods.Items {
+			pod := pods.Items[i]
+			labelValue := "true"
+			// We assume that OFED pod contains only one contain to simplify the logic.
```
adrianchiris, line 137: nit: contain -> container

adrianchiris (commented):
a couple of nits, overall looks good

pkg/nodeinfo/attributes.go
e0ne: Good idea! I didn't thought about it but it seems reasonable

controllers/nicclusterpolicy_controller.go
e0ne: ack

adrianchiris (approved):
LGTM, just need CI to pass before merge

Issue #184, opened 2021-05-14
Failure to deploy mofed on a single node in cluster will halt deployment progress 


e0ne:
If one or more mofed pods fail to deploy or are restarted the deployment will not proceed to the following stages,

In a large cluster this will be an issue since at least one of the nodes will have some sort of a problem.

We should progress deployment stages on a per node basis instead of a complete stage status

PR #185, opened 2021-05-14
Do not wait state group deployment in manager

e0ne:
We manage deployment order in a manager only for NicClusterPolicy now.
It's not needed anymore since we use labeling mechanism to wait until
MOFED will be deployed.

There should not be any issues for parallel deployment of networking
stack (CNI plugins, Multus, Whereabouts, etc) and MOFED because they
don't depend on each other.

This patch improves deployment speed on a large environment and fixes
an issue if one MOFED pod block any other deployment on the other nodes.

Closes: #184

### DIFF
--- a/pkg/state/manager.go
+++ b/pkg/state/manager.go
@@ -85,6 +85,7 @@ func (smgr *stateManager) SyncState(customResource interface{}, infoCatalog Info
 	managerResult := Results{
 		Status: SyncStateNotReady,
 	}
+	statesReady := true
 
 	for i, stateGroup := range smgr.stateGroups {
 		log.V(consts.LogLevelInfo).Info("Sync State group", "index", i)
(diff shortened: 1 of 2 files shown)


pkg/state/manager.go
```diff
@@ -83,7 +83,7 @@ func (smgr *stateManager) SyncState(customResource interface{}, infoCatalog Info
 	// Sync groups of states, transition from one group to the other when a group finishes
 	log.V(consts.LogLevelInfo).Info("Syncing system state")
 	managerResult := Results{
-		Status: SyncStateNotReady,
+		Status: SyncStateReady,
```
moshe010, line 86: I prefer to init it with SyncStateNotReady and in line 99 check if  done change to SyncStateReady

e0ne: Ok, I’ll rollback this change. It was done to remove one extra “if statement” and/or assignment but you’re right, let’s keep code more readable

pkg/state/manager.go
```diff
@@ -96,14 +97,20 @@ func (smgr *stateManager) SyncState(customResource interface{}, infoCatalog Info
...
+		} else {
+			statesReady = true
+			log.V(consts.LogLevelInfo).Info("Sync Completed successfully for State group", "index", i)
 		}
-		log.V(consts.LogLevelInfo).Info("Sync Completed successfully for State group", "index", i)
 	}
-	// Done Syncing CR
-	managerResult.Status = SyncStateReady
-	log.V(consts.LogLevelInfo).Info("Sync Done for custom resource")
+	if statesReady {
```
moshe010, line 109: don't you overwrite the statesReady  in the for loop? so here you are checking according to the last one?

pkg/state/manager.go
```diff
@@ -96,14 +97,20 @@ func (smgr *stateManager) SyncState(customResource interface{}, infoCatalog Info
 			log.V(consts.LogLevelError).Info("Error while syncing states", "Error:", err)
 			return managerResult, err
 		}
+
 		if !done {
+			statesReady = false
```
moshe010, line 102: can we move it to start of the loop?

moshe010 (approved):

PR #190, opened 2021-05-31
Add nvPeerDriver.fromGPUDriver option to NicClusterPolicy CR

ykulazhenkov:
Add nvPeerDriver.fromGPUDriver option to NicClusterPolicy CR

This option is passed as parameter to nvPeerDriver container.
If "fromGPUDriver" is true, nvPeerDriver container will try to use
nvidia_peermem module which is a part of
the GPU driver distro (in 465.x branch) instead of nv_peer_mem module
compiled from sources. If the option is true, but  GPU driver doesn't
provide nvidia_peermem module, then nv_peer_mem will be used.

Default value for the option is false, which means to use
nv_peer_mem module compiled from sources.

PR in ofed-docker repo which adds required logic to nvPeerMem container https://github.com/Mellanox/ofed-docker/pull/25

### DIFF
--- a/api/v1alpha1/nicclusterpolicy_types.go
+++ b/api/v1alpha1/nicclusterpolicy_types.go
@@ -72,6 +72,8 @@ type NVPeerDriverSpec struct {
 	ImageSpec `json:""`
 	// GPU driver sources path - Optional
 	GPUDriverSourcePath string `json:"gpuDriverSourcePath,omitempty"`
+	// Allow to use builtin nvidia_peermem module from GPU driver
+	FromGPUDriver bool `json:"fromGPUDriver,omitempty"`
 }
 
 // DevicePluginSpec describes configuration options for device plugin
--- a/config/crd/bases/mellanox.com_nicclusterpolicies.yaml
+++ b/config/crd/bases/mellanox.com_nicclusterpolicies.yaml
@@ -47,6 +47,10 @@ spec:
                 description: NVPeerDriverSpec describes configuration options for
                   NV Peer Memory driver
                 properties:
+                  fromGPUDriver:
+                    description: Allow to use builtin nvidia_peermem module from GPU
+                      driver
+                    type: boolean
                   gpuDriverSourcePath:
                     description: GPU driver sources path - Optional
                     type: string
--- a/deployment/network-operator/crds/mellanox.com_nicclusterpolicies.yaml
+++ b/deployment/network-operator/crds/mellanox.com_nicclusterpolicies.yaml
@@ -47,6 +47,10 @@ spec:
                 description: NVPeerDriverSpec describes configuration options for
                   NV Peer Memory driver
                 properties:
+                  fromGPUDriver:
+                    description: Allow to use builtin nvidia_peermem module from GPU
+                      driver
+                    type: boolean
                   gpuDriverSourcePath:
                     description: GPU driver sources path - Optional
                     type: string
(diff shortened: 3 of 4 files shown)


moshe010 (commented):
why we need to deploy nv_peer_mem container in case nvidia_permem is loaded?

ykulazhenkov:
Ok, I will try to briefly cover other options, so we can choose which fits best.

To successfully load **nvidia_peermem**(part of GPU driver distro) or **nv_peer_mem** modules, we need OFED and GPU drivers to be on a node. To load the module at the right time operator or operators should include logic to wait for OFED and GPU driver readiness.

We have several options of how to implement support for using nvidia_peermem module shipped with GPU driver. 
If we summarize and drop details, we can highlight two options:

1. nv_peer_mem/nvidia_peermem deployment logic concentrated in network-operator (current implementation in PR)
2. nv_peer_mem/nvidia_peermem deployment logic split between network and GPU operators

It depends on our final goal which option is better to use. 
The main question is: Do we plan to finally drop support of nv_peer_mem deployment by network-operator soon? 

**Option 1**
The current implementation in this PR is assuming that answer to the question is no, that we plan to support both modules for a quite long period. In this case, I think it can be reasonable to put all deployment logic in network-operator to concentrate dependency management logic in one component and finally simplify rollout and upgrade flows.

_Pros for all option 1 variants:_
- deployment logic for nv_peer_mem/nvidia_peermem modules  in one place
- no changes required in gpu-operator

_Cons for all option 1 variants:_

- network-operator manages loading of the nvidia_peermem module which is a part of GPU driver distro. gpu-operator has no control over this module, this can cause some problems for gpu-operator upgrade

**Option 1.A  (current)** nv_peer_mem container contains everything to built/load both modules (nv_peer_mem/nvidia_peermem). It also check for nvidia_peermem module presence, so there is no need to implement version check logic on the network-operator level.

_Pros:_ 

- minimal API impact
- no need to add new images/manifests to network-operator
- no need to implement chek for nvidia_peermem module presence (GPU driver version check) on network-operator level

_Cons:_

- we need to deploy a large nv_peer_mem image on a node even if we plan to use a builtin nvidia_peermem module

**Option 1.B** nv_peer_mem container  built/load nv_peer_mem module. A small "loader" POD managed by network-operator is used to load nvidia_peermem module.
Cons:

- NicClusterPolicy should be extended to support configuration options(repo, image, tag) for the "loader" image
- network-operator should contain logic to check GPU driver version  to decide which image to use nv_peer_mem or "loader"

**Option 2**
Deployment logic distributed between network and GPU operators. I think this is a good option if we finally plan to drop support of nv_peer_mem container in network-operator in the future.
In this case, network-operator will deploy nv_peer_mem container only if GPU driver version is less than /_first version which supports nvidia_peermem deployment/_. GPU operator is responsible to deploy nvidia_peermem module after OFED driver is ready.
**Required changes:**

- Network-operator should have logic to check GPU driver version and to prevent nv_peer_mem deployment when the version is greater than some value.
- Network-operator should have a new component that will expose information about OFED version/state as an annotation on node, when OFED is loaded and ready
- GPU-operator should have logic to deploy "loader" POD for nvidia_peermem module when OFED is ready 

_Pros for Option 2:_

-  gpu-operator manges nvidia_peermem module which is a part of GPU driver

_Cons for Option 2:_

- require changes in both operators
- deployment logic split between two operators and two teams, need to support it in both places.
- possible upgrade problems when transiting from nv_peer_mem to nvidia_peermem

ykulazhenkov:
After discussion, it was decided to take a different approach. The work will continue in a separate PR

adrianchiris:
I think that Option 2 is our end goal here as discussed.
we just need to give GPU operator folks heads up on this behaviour so they can adjust before releasing a version of the gpu operator with a GPU driver that contains nvidia peer mem module.

nv_peer_mem driver daemonset should only be deployed on nodes that have  `nvidia.com/cuda.driver.major` less than the version that supports nvidia peer mem.

PR #191, opened 2021-06-01
Add additional nodeAffinity rule for nv-peer-mem-driver POD

ykulazhenkov:
This rule prevents driver from deploying on node
which has NVIDIA GPU driver with version >= 465.
Starting from v465 NVIDIA GPU driver includes a built-in nvidia_peermem
module which is a replacement for nv_peer_mem module.
NVIDIA GPU operator manages nvidia_peermem module loading.

Signed-off-by: Yury Kulazhenkov <ykulazhenkov@nvidia.com>

### DIFF
--- a/manifests/stage-nv-peer-mem-driver/0050_nv-peer-mem-driver-ds.yaml
+++ b/manifests/stage-nv-peer-mem-driver/0050_nv-peer-mem-driver-ds.yaml
@@ -146,3 +146,10 @@ spec:
                 operator: Gt
                 values:
                   - "0"
+              # cuda driver starting from v465 should include builtin nvidia_peermem module
+              # which should be used instead nv_peer_mem
+              # prevent scheduling nv-peer-mem-driver POD on nodes with cuda driver version >= 465
+              - key: nvidia.com/cuda.driver.major
+                operator: Lt
+                values:
+                  - "{{ .RuntimeSpec.MaxCudaVersion }}"
--- a/pkg/nodeinfo/attributes.go
+++ b/pkg/nodeinfo/attributes.go
@@ -28,25 +28,30 @@ var log = logf.Log.WithName("nodeinfo")
 
 // Node labels used by nodeinfo package
 const (
-	NodeLabelOSName        = "feature.node.kubernetes.io/system-os_release.ID"
-	NodeLabelOSVer         = "feature.node.kubernetes.io/system-os_release.VERSION_ID"
-	NodeLabelKernelVerFull = "feature.node.kubernetes.io/kernel-version.full"
-	NodeLabelHostname      = "kubernetes.io/hostname"
-	NodeLabelCPUArch       = "kubernetes.io/arch"
-	NodeLabelMlnxNIC       = "feature.node.kubernetes.io/pci-15b3.present"
-	NodeLabelNvGPU         = "nvidia.com/gpu.present"
-	NodeLabelWaitOFED      = "network.nvidia.com/operator.mofed.wait"
+	NodeLabelOSName           = "feature.node.kubernetes.io/system-os_release.ID"
+	NodeLabelOSVer            = "feature.node.kubernetes.io/system-os_release.VERSION_ID"
+	NodeLabelKernelVerFull    = "feature.node.kubernetes.io/kernel-version.full"
+	NodeLabelHostname         = "kubernetes.io/hostname"
+	NodeLabelCPUArch          = "kubernetes.io/arch"
+	NodeLabelMlnxNIC          = "feature.node.kubernetes.io/pci-15b3.present"
+	NodeLabelNvGPU            = "nvidia.com/gpu.present"
+	NodeLabelWaitOFED         = "network.nvidia.com/operator.mofed.wait"
+	NodeLabelCudaVersionMajor = "nvidia.com/cuda.driver.major"
 )
 
 type AttributeType int
 
 // Attribute type Enum, add new types before Last and update the mapping below
 const (
+	// required attrs
 	AttrTypeHostname = iota
 	AttrTypeCPUArch
 	AttrTypeOSName
 	AttrTypeOSVer
-	AttrTypeLast
+	// optional attrs
+	AttrTypeCudaVersionMajor
+
+	OptionalAttrsStart = AttrTypeCudaVersionMajor
 )
 
 var attrToLabel = []string{
@@ -58,6 +63,8 @@ var attrToLabel = []string{
 	NodeLabelOSName,
 	// AttrTypeOSVer
 	NodeLabelOSVer,
+	// AttrTypeCudaVersionMajor
+	NodeLabelCudaVersionMajor,
 }
 
 // NodeAttributes provides attributes of a specific node
@@ -91,8 +98,9 @@ func newNodeAttributes(node *corev1.Node) NodeAttributes {
 	nLabels := node.GetLabels()
 	for attrType, label := range attrToLabel {
 		err = attr.fromLabel(AttributeType(attrType), nLabels, label)
-		if err != nil {
-			log.V(consts.LogLevelWarning).Info("Warning:", "cannot create NodeAttribute(%x), %v", attrType, err)
+		if err != nil && attrType < OptionalAttrsStart {
+			log.V(consts.LogLevelWarning).Info("Cannot create NodeAttribute",
+				"attribute", attrType, "error:", err.Error())
 		}
 	}
 	return attr
--- a/pkg/nodeinfo/filter.go
+++ b/pkg/nodeinfo/filter.go
@@ -82,3 +82,62 @@ func (b *NodeLabelFilterBuilder) Reset() *NodeLabelFilterBuilder {
 	b.filter = newNodeLabelFilter()
 	return b
 }
+
+// A node label filter which ignores label value. use NewNodeLabelNoValFilterBuilder to create instances
+type nodeLabelNoValFilter struct {
+	labels map[string]struct{}
+}
+
+// addLabel adds a label to nodeLabelFilter
+func (nlf *nodeLabelNoValFilter) addLabel(key string) {
+	nlf.labels[key] = struct{}{}
+}
+
+// Apply Filter on Nodes
+func (nlf *nodeLabelNoValFilter) Apply(nodes []*corev1.Node) (filtered []*corev1.Node) {
+NextIter:
+	for _, node := range nodes {
+		nodeLabels := node.GetLabels()
+		for k := range nlf.labels {
+			if _, ok := nodeLabels[k]; ok {
+				continue
+			}
+			// label not found on node or label value missmatch
+			continue NextIter
+		}
+		filtered = append(filtered, node)
+	}
+	return filtered
+}
+
+// newNodeLabelNoValFilter creates a new nodeLabelNoValFilter
+func newNodeLabelNoValFilter() nodeLabelNoValFilter {
+	return nodeLabelNoValFilter{labels: make(map[string]struct{})}
+}
+
+// NodeLabelNoValFilterBuilder is a builder for nodeLabelFilter
+type NodeLabelNoValFilterBuilder struct {
+	filter nodeLabelNoValFilter
+}
+
+// NewNodeLabelNoValFilterBuilderr returns a new NodeLabelNoValFilterBuilder
+func NewNodeLabelNoValFilterBuilderr() *NodeLabelNoValFilterBuilder {
+	return &NodeLabelNoValFilterBuilder{filter: newNodeLabelNoValFilter()}
+}
+
+// WithLabel adds a label for the Build process of the Label filter
+func (b *NodeLabelNoValFilterBuilder) WithLabel(key string) *NodeLabelNoValFilterBuilder {
+	b.filter.addLabel(key)
+	return b
+}
+
+// Build the Filter
+func (b *NodeLabelNoValFilterBuilder) Build() Filter {
+	return &b.filter
+}
+
+// Reset NodeLabelNoValFilterBuilder
+func (b *NodeLabelNoValFilterBuilder) Reset() *NodeLabelNoValFilterBuilder {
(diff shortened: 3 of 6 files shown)


manifests/stage-nv-peer-mem-driver/0050_nv-peer-mem-driver-ds.yaml
```diff
@@ -146,3 +146,10 @@ spec:
                 operator: Gt
                 values:
                   - "0"
+              # cuda driver starting from v465 should include builtin nvidia_peermem module
+              # which should be used instead nv_peer_mem
+              # prevent scheduling nv-peer-mem-driver POD on nodes with cuda driver version >= 465
+              - key: nvidia.com/cuda.driver.major
```
adrianchiris, line 152: A layman question here: when cuda driver get updated and this condition is updated to be no longer true, will the pod get unscheduled from the node ?

the term `requiredDuringSchedulingIgnoredDuringExecution` has me a bit worried.

ykulazhenkov: > will the pod get unscheduled from the node ?

Yes

adrianchiris: hmm so i wonder why we have `IgnoredDuringExecution`.

according to k8s documentation it shouldnt evict the pod

>Node affinity is conceptually similar to nodeSelector -- it allows you to constrain which nodes your pod is eligible to be scheduled on, based on labels on the node.
>
>There are currently two types of node affinity, called requiredDuringSchedulingIgnoredDuringExecution and preferredDuringSchedulingIgnoredDuringExecution. You can think of them as "hard" and "soft" respectively, in the sense that the former specifies rules that must be met for a pod to be scheduled onto a node (similar to nodeSelector but using a more expressive syntax), while the latter specifies preferences that the scheduler will try to enforce but will not guarantee.

And here is the more relevant part:

> The "IgnoredDuringExecution" part of the names means that, similar to how nodeSelector works, if labels on a node change at runtime such that the affinity rules on a pod are no longer met, the pod continues to run on the node. In the future we plan to offer requiredDuringSchedulingRequiredDuringExecution which will be identical to requiredDuringSchedulingIgnoredDuringExecution except that it will evict pods from nodes that cease to satisfy the pods' node affinity requirements.

ykulazhenkov: nv-peer-mem-driver is a DaemonSet.
DaemonSet controller manage PODs creation and removal 
Check this doc section: https://kubernetes.io/docs/concepts/workloads/controllers/daemonset/#updating-a-daemonset

AbdYsn:
The niv_operator CI failure is weird, the nic cluster policy is ready while the nv-peer-mem still is not ready. @ykulazhenkov do you think this PR can cause that?

ykulazhenkov:
> The niv_operator CI failure is weird, the nic cluster policy is ready while the nv-peer-mem still is not ready. @ykulazhenkov do you think this PR can cause that?

I don't think this is directly related to this change. 
I see exact same issue with rdma-shared-dp-ds POD (POD start time is after NicClusterPolicy was ready)

manifests/stage-nv-peer-mem-driver/0050_nv-peer-mem-driver-ds.yaml
adrianchiris: nice ! thanks.

manifests/stage-nv-peer-mem-driver/0050_nv-peer-mem-driver-ds.yaml
```diff
@@ -146,3 +146,10 @@ spec:
...
+                operator: Lt
+                values:
```
e0ne, line 154: What value will be in nic-cluster-policy state in case if  version is 465 or above?

ykulazhenkov: If all nodes have GPU drivers with version => 465 then from the network-operator point of view this is equal to situation when there are no nodes with GPU in the cluster. In this case, state-NV-Peer will be notReady.

moshe010: I think that misleading to the user and also according to [1] it will keep trying to reconcile. 
can we workaround it that it won't try to reconcile when GPU drivers with version => 465?

[1] - https://github.com/Mellanox/network-operator/blob/e73e49b99ff9d2f0eb92ab95143a9b4890f96d07/pkg/state/state_nv_peer.go#L89-L93

adrianchiris: @moshe010  IIRC , this will happen if nvidia.com/gpu.present is not set on node.
in this case i believe it will be set and the gpu driver annotation will be >=465.
this will cause the DS to be created but fail on readiness in : https://github.com/Mellanox/network-operator/blob/e73e49b99ff9d2f0eb92ab95143a9b4890f96d07/pkg/state/state_skel.go#L188

adrianchiris: There is an issue where the order of thing is not really assured after initial deployment however its not related to this PR.

ykulazhenkov:
I updated PR.
I added shouldIgnoreStatus check to stateNVPeer.Sync()
Please, check comments in code for details

pkg/state/state_nv_peer.go
```diff
@@ -159,3 +172,36 @@ func (s *stateNVPeer) getManifestObjects(
...
 }
+
+// shouldIgnoreStatus check if nvPeerMem DS status should be ignored
+// Workaround to support switching from nv_peer_mem module which is managed by network-operator
+// to nvidia_peermem module which is a part of the GPU driver starting from v465.
+// We should ignore nvPeerMem status in case all nodes in a cluster have GPU driver version > 465
+// if some nodes still using GPU driver < 465 or there are no nodes with GPU driver, we should continue sync attempts
+// we don't remove DS, just ignore its status. This is required to be able to recovery nv-peer-mem in situations when
+// GPU driver version was downgraded or when we have an environment with mixed versions of the GPU drivers
+// and node with older version appear after node with never.
```
adrianchiris, line 183: > node with never.

node with a newer version

pkg/nodeinfo/attributes.go
```diff
@@ -36,17 +36,22 @@ const (
 	NodeLabelMlnxNIC       = "feature.node.kubernetes.io/pci-15b3.present"
 	NodeLabelNvGPU         = "nvidia.com/gpu.present"
 	NodeLabelWaitOFED      = "network.nvidia.com/operator.mofed.wait"
+	NodeLabelCudaVersion   = "nvidia.com/cuda.driver.major"
 )
 
 type AttributeType int
```
adrianchiris, line 42: An alternative would be to extend this to be a struct where we are able to mark an attribute as optional. current approach is less intrusive though.

pkg/nodeinfo/filter.go
```diff
@@ -72,6 +75,13 @@ func (b *NodeLabelFilterBuilder) WithLabel(key, val string) *NodeLabelFilterBuil
 	return b
 }
 
+// WithLabelIgnoreValue adds a label for the Build process of the Label filter
+// use special constant as value, which mean that nodeLabelFilter will ignore value when processing nodeLabel
+func (b *NodeLabelFilterBuilder) WithLabelIgnoreValue(key string) *NodeLabelFilterBuilder {
```
adrianchiris, line 80: `GetNodeAttributes()` accepts multiple filters.

you can define a new filter `nodeLabelNoValFilter` which will filter according to label presence without the value.

then you would not change logic to existing filter and avoid using special value.

pkg/state/state_nv_peer.go
```diff
@@ -103,6 +108,13 @@ func (s *stateNVPeer) Sync(customResource interface{}, infoCatalog InfoCatalog)
 	if err != nil {
 		return SyncStateNotReady, errors.Wrap(err, "failed to create/update objects")
 	}
+
+	if s.shouldIgnoreStatus(nodeInfo) {
```
adrianchiris, line 112: nit: can you add a short one-liner comment why this is called.
(i know that this method description explains it, but a gist as a reminder here would be helpful to avoid looking it up in where the method is defined is helpful.)

pkg/state/state_nv_peer.go
```diff
@@ -159,3 +172,36 @@ func (s *stateNVPeer) getManifestObjects(
...
+// shouldIgnoreStatus check if nvPeerMem DS status should be ignored
+// Workaround to support switching from nv_peer_mem module which is managed by network-operator
+// to nvidia_peermem module which is a part of the GPU driver starting from v465.
+// We should ignore nvPeerMem status in case all nodes in a cluster have GPU driver version > 465
+// if some nodes still using GPU driver < 465 or there are no nodes with GPU driver, we should continue sync attempts
+// we don't remove DS, just ignore its status. This is required to be able to recovery nv-peer-mem in situations when
+// GPU driver version was downgraded or when we have an environment with mixed versions of the GPU drivers
+// and node with older version appear after node with never.
+// nvPeerMem config should be removed from NicClusterPolicy explicitly to remove DS
+func (s *stateNVPeer) shouldIgnoreStatus(nodeInfo nodeinfo.Provider) bool {
```
adrianchiris, line 185: maybe:

`cudaDriverNewer(cudaVer int)`

Then move the comment to where this is called ?

pkg/nodeinfo/attributes.go
```diff
@@ -36,17 +36,22 @@ const (
 	NodeLabelMlnxNIC       = "feature.node.kubernetes.io/pci-15b3.present"
 	NodeLabelNvGPU         = "nvidia.com/gpu.present"
 	NodeLabelWaitOFED      = "network.nvidia.com/operator.mofed.wait"
+	NodeLabelCudaVersion   = "nvidia.com/cuda.driver.major"
```
adrianchiris, line 39: nit: `NodeLabelCudaVersionMajor` ?

pkg/state/state_nv_peer.go
```diff
@@ -21,6 +22,9 @@ import (
 const stateNVPeerName = "state-NV-Peer"
 const stateNVPeerDescription = "Nvidia Peer Memory driver deployed in the cluster"
 
+// starting from this GPU driver version nvPeerMem driver should not deploy
+const maxCudaVersion = 465
```
adrianchiris, line 26: nit: `maxCudaVersionMajor` ?

pkg/state/state_nv_peer.go
ykulazhenkov: done

pkg/nodeinfo/attributes.go
ykulazhenkov: Yes, I was also thinking about a similar approach which using stuct to hold information about is the option is required or not. The approach with struct is much wordy and for me, it decreases readability a bit. I prefer to stick with the current implementation.

pkg/nodeinfo/filter.go
ykulazhenkov: Done. I moved this logic to separate `nodeLabelNoValFilter` filter

pkg/nodeinfo/attributes.go
ykulazhenkov: done

pkg/state/state_nv_peer.go
ykulazhenkov: done

pkg/state/state_nv_peer.go
ykulazhenkov: I think that `shouldIgnoreStatus` function name is accurate because it describes what the function intends to do.
This logic is required to workaround the specific nv_peer_mem upgrade problem, so for me, it looks like a good idea to isolate workaround logic in a separate function and not to put it in the main `Sync` function. This will simplify this logic removal if required in the future.

pkg/nodeinfo/filter.go
adrianchiris: Thanks!

pkg/state/state_nv_peer.go
adrianchiris: i see, OK

adrianchiris (commented):
small nit, but otherwise LGTM

pkg/state/state_nv_peer.go
```diff
@@ -103,6 +108,16 @@ func (s *stateNVPeer) Sync(customResource interface{}, infoCatalog InfoCatalog)
 	if err != nil {
 		return SyncStateNotReady, errors.Wrap(err, "failed to create/update objects")
 	}
+
+	// check if nvPeerMem status should be ignored
+	// workaround to support upgrade from nv_peer_mem > nvidia_peermem module
```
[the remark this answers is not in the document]
ykulazhenkov: Yep, Make sense. Fixed

adrianchiris:
@e0ne can you review and merge ?

pkg/nodeinfo/attributes.go
```diff
@@ -91,8 +98,9 @@ func newNodeAttributes(node *corev1.Node) NodeAttributes {
 	nLabels := node.GetLabels()
 	for attrType, label := range attrToLabel {
 		err = attr.fromLabel(AttributeType(attrType), nLabels, label)
-		if err != nil {
-			log.V(consts.LogLevelWarning).Info("Warning:", "cannot create NodeAttribute(%x), %v", attrType, err)
+		if err != nil && attrType < OptionalAttrsStart {
```
e0ne, line 101: Do we need this change? We'll miss logs if operator fails to create optional node attribute

e0ne: LGTM, thanks for the clarification

e0ne (approved):

PR #193, opened 2021-06-07
Sign in to dockerhub in travis before building image

AbdYsn:
Added a before_script step for travis CI to sign in
to dockerhub to fix the issue with dockerhub pull rate
limit.

Signed-off-by: abdallahyas <abdallahyas@mellanox.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ env:
 
 before_script:
   - go get -u github.com/mattn/goveralls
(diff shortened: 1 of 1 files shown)


AbdYsn:
seeing https://docs.travis-ci.com/user/pull-requests/#pull-requests-and-security-restrictions, it seems this will not work. will need to find another solution.

PR #200, opened 2021-06-13
WIP. Move image build step to 'deploy' phase on Travis CI

e0ne:
We'll build image on Jenkins on each pull request.

Travis CI will build image only during deploy phase to not face
DockerHub image pull limit.

depends-on: kubernetes-ci#131

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,6 @@ script:
   - make lint-helm
   - make test-coverage
     #  - goveralls -coverprofile=network-operator.cover -service=travis-ci
-  - make image
 
 before_deploy:
   - docker login -u "$REGISTRY_USER" -p "$REGISTRY_PASS"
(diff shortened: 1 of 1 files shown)


example/rdma-test-pod1.yml
```diff
@@ -1,4 +1,4 @@
-apiVersion: v1
+ apiVersion: v1
```
moshe010, line 1: why this change?

.travis.yml
```diff
@@ -31,7 +30,9 @@ deploy:
   # Push image to Dockerhub on tag
   - provider: script
     skip_cleanup: true
-    script: bash scripts/docker-push.sh $IMAGE_NAME $TRAVIS_TAG $TRAVIS_CPU_ARCH
+    script:
+      - make image
```
moshe010, line 34: how will we know that the image build is passing before merging PR?

e0ne: It will be tested in https://github.com/Mellanox/kubernetes-ci/pull/131

moshe010:
I think we agree that this is not needed

PR #201, opened 2021-06-18
Remove '--devel' from 'helm install' command

e0ne:
We don't have '-alpha' and '-beta' versions now so we don't need
to add '--devel' flag to install Network Operator chart.


moshe010 (approved):
Please make sure also the nvidia doc is update as well.

PR #218, opened 2021-07-13
Enable GitHub Actions

e0ne:


.github/workflows/build.yml
```diff
@@ -0,0 +1,52 @@
+name: "Build"
+on: [push, pull_request]
+
+jobs:
+  lint-code:
+    name: golangci-lint
+    runs-on: ubuntu-latest
```
moshe010, line 7: do we want to pin the ubuntu version? So we wont' break on stuff change in ubuntuOS? 
Currently we support Ubuntu 20.04 so maybe it better to pin to that version?

e0ne: it uses ubuntu 20.04.2 [1].  According to [2] there is no plan to upgrade Ubuntu before the next LTS.

[1] https://github.com/Mellanox/network-operator/pull/218/checks?check_run_id=3058567841#step:1:4
[2] https://github.com/actions/virtual-environments/

moshe010: I understand that what I mean is that I want to use "ubuntu-20.04" and not "ubuntu-latest" so when they will change it won't effect us.

e0ne: I don't have strong opinion here. Both approaches will work. Using pinned version is more stable. From other side, it's a version of OS to run unit-tests and it's different than we've got for production environment: there is no RedHat-based images at the moment.

moshe010:
Please fix DCO Action required after 1s — DCO

moshe010 (approved):

PR #229, opened 2021-08-05
Add Helm Chart values example do deploy at NVIDIA DGX SYSTEMS

e0ne:


moshe010 (commented):
I think it better to move it to network-operator/deployment/network-operator.
so we will have values.yaml and dgx values yaml. What do you think?

moshe010:
> I think it better to move it to network-operator/deployment/network-operator.
> so we will have values.yaml and dgx values yaml. What do you think?

Ok let keep it here until it approved by QA

PR #238, opened 2021-09-09
Updated MOFED to 5.4-2.4.1.3

e0ne:


moshe010 (commented):
we should update doc and what we support with this new OFED

e0ne:
> we should update doc and what we support with this new OFED

let's do these updates once we test new MOFED image with RH

moshe010 (approved):

Issue #65, opened 2020-11-25
Add support for nodeSelector


MartinForReal:
We need to select target node using labelSelector so that we can define multiple policy for cluster if we have nodes which have different mellanox nics.

adrianchiris:
@MartinForReal can you elaborate more on your use-case ?

MartinForReal:
There are connect-3/4/5 nics available in our cluster and we can define multiple nicclusterpolicy to install differenct version of ofed on target node.  And one clusterpolicy is defined to configure multus and device plugin.

adrianchiris:
This sounds interesting, although the project does not officially support ConnectX-3.
We are planning to tag a release in December, and have limited resources which are directed towards finalizing content/testing.
However this will be considered as an enhancement and candidate for future release.

As its an opensource project, code contributions are welcome :)

PR #248, opened 2021-10-01
Propagate nodeSelector and nodeAffinity properties from the NicClusterPolicy to network operator dependencies

almaslennikov:


AbdYsn:
i am seeing the issue with the sriov device plugin, can you rebase the patch to master to fix it?

moshe010:
why don't you use you nviidia mail ? (I see combat0705@yandex.ru)? You can configure github to have both mails. 
Also can you sqush some of the commits? )(e.g. 
- Reverted go.sum
- Reverted examples
- Linter fix

adrianchiris:
Can you provide some additional information on the issue in this PR ?
what is the issue, its affects and proposed solution.

api/v1alpha1/nicclusterpolicy_types.go
```diff
@@ -101,16 +102,24 @@ type SecondaryNetworkSpec struct {
...
 type NicClusterPolicySpec struct {
 	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
 	// Important: Run "make" to regenerate code after modifying this file
 
-	OFEDDriver             *OFEDDriverSpec       `json:"ofedDriver,omitempty"`
-	NVPeerDriver           *NVPeerDriverSpec     `json:"nvPeerDriver,omitempty"`
-	RdmaSharedDevicePlugin *DevicePluginSpec     `json:"rdmaSharedDevicePlugin,omitempty"`
-	SriovDevicePlugin      *DevicePluginSpec     `json:"sriovDevicePlugin,omitempty"`
-	SecondaryNetwork       *SecondaryNetworkSpec `json:"secondaryNetwork,omitempty"`
+	ClusterNodeAffinity     *v1.NodeAffinity `json:"nodeAffinity,omitempty"`
```
adrianchiris, line 116: why do we need both ?

api/v1alpha1/nicclusterpolicy_types.go
```diff
@@ -101,16 +102,24 @@ type SecondaryNetworkSpec struct {
...
+	ClusterNodeSelectorSpec `json:""`
```
adrianchiris, line 117: no need to specify its field ? `nodeSelector` in the json annotation ?

api/v1alpha1/nicclusterpolicy_types.go
```diff
@@ -101,16 +102,24 @@ type SecondaryNetworkSpec struct {
 	IpamPlugin *ImageSpec `json:"ipamPlugin,omitempty"`
 }
 
+type ClusterNodeSelectorSpec struct {
```
adrianchiris, line 105: nit: just NodeSelector ?

api/v1alpha1/nicclusterpolicy_types.go
adrianchiris: also this is already in NicClusterPolicy struct, so perhaps just name it `NodeAffinity` ?

api/v1alpha1/nicclusterpolicy_types.go
almaslennikov: As the struct is just a wrapper over `map[string]string` that represents the nodeSelector, it doesn't map to any json field directly. `nodeSelector` annotation is defined for the struct member.

api/v1alpha1/nicclusterpolicy_types.go
almaslennikov: We need both `nodeSelector` and `nodeAffinity` as they're both supported by kubernetes

api/v1alpha1/nicclusterpolicy_types.go
almaslennikov: Renamed

api/v1alpha1/nicclusterpolicy_types.go
adrianchiris: ah thats embedded struct, missed that :)

api/v1alpha1/nicclusterpolicy_types.go
adrianchiris: and why do we need to support both in the policy ?

I just dont fully understand the problem you are trying to solve hence my (possibly dumb) questions.

moshe010:
is this commit still draft? also please make sure all sighoff commits are with nvidia e-mail.

moshe010:
> Can you provide some additional information on the issue in this PR ? what is the issue, its affects and proposed solution.

@adrianchiris  - the thought was to give alex to implement the requested feature in https://github.com/Mellanox/network-operator/issues/65

manifests/stage-nv-peer-mem-driver/0050_nv-peer-mem-driver-ds.yaml
```diff
@@ -153,3 +154,23 @@ spec:
...
+        {{- end }}
+      {{- end }}
+      {{- end }}
+      serviceAccountName: multus
+      {{- if .CrSpec.ImagePullSecrets }}
+      imagePullSecrets:
+      {{- range .CrSpec.ImagePullSecrets }}
+        - name: {{ . }}
+      {{- end }}
+      {{- end }}
```
adrianchiris, line 176: nit: missing newline

manifests/stage-nv-peer-mem-driver/0050_nv-peer-mem-driver-ds.yaml
```diff
@@ -153,3 +154,23 @@ spec:
...
+      {{- end}}
+      {{- if (len .NicClusterPolicy.ClusterNodeSelector) }}
+      nodeSelector:
+      {{- range $key, $value := .NicClusterPolicy.ClusterNodeSelector }}
+        {{ $key }}: {{ $value }}
+        {{- end }}
+      {{- end }}
+      {{- end }}
+      serviceAccountName: multus
+      {{- if .CrSpec.ImagePullSecrets }}
```
adrianchiris, line 171: this is not really related to this PR right ?

api/v1alpha1/nicclusterpolicy_types.go
adrianchiris: > @adrianchiris - the thought was to give alex to implement the requested feature in #65

So we dont really need both, but just provide the user with an option to apply different policy to different nodes.
we can use just node affinity IMO as it also provides same functionality as node selector (and more)
https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#node-affinity

that said, if it offers better user-experience im OK with having both

e0ne: Here is issue we're trying to address: #252. Since nodeSelector requires additional logic to not break current workflows I suggest to split it to several PRs. Let's implement node affinity support first

pkg/state/state_cni_plugins.go
```diff
@@ -114,7 +115,8 @@ func (s *stateCNIPlugins) GetWatchSources() map[string]*source.Kind {
 func (s *stateCNIPlugins) getManifestObjects(
 	cr *mellanoxv1alpha1.NicClusterPolicy) ([]*unstructured.Unstructured, error) {
 	renderData := &CNIPluginsManifestRenderData{
-		CrSpec: cr.Spec.SecondaryNetwork.CniPlugins,
+		CrSpec:           cr.Spec.SecondaryNetwork.CniPlugins,
+		NicClusterPolicy: cr.Spec,
 		RuntimeSpec: &runtimeSpec{
```
e0ne, line 120: We should not pass the whole NicClusterPolicy object to template. Let's pass only NodeAffinity or move it under RuntimeSpec field

e0ne (changes requested):

PR #255, opened 2021-10-06
Update RDMA Shared device plugin to v1.2.0

e0ne:


AbdYsn:
@e0ne i am seeing 
```
sh: lsmod: command not found
waiting for OFED drivers to be loaded
``` 
in the logs of the rdma-shared-device-plugin init container for the failing CIs, we don't see this in the kind CI because it does not have a test that deploys the rdma-shared-device-plugin.

e0ne:
We need to hold it until https://github.com/Mellanox/k8s-rdma-shared-dev-plugin/pull/43 will be released in v.1.2.1 or newer

PR #272, opened 2021-11-03
Accept resource names for hostdevicenetwork either with or without prefix

almaslennikov:
Currently network operator doesn't allow deploying host device networks if the resource name start with "nvidia.com" prefix. The solution is to normalize all resource names for hostdevicenetwork state to start with a common prefix.

### DIFF
--- a/pkg/state/state_hostdevice_network_test.go
+++ b/pkg/state/state_hostdevice_network_test.go
@@ -17,6 +17,8 @@ limitations under the License.
 package state
 
 import (
+	"strings"
+
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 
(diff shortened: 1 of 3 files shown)


pkg/state/state_hostdevice_network_test.go
```diff
@@ -29,6 +31,14 @@ import (
 	"github.com/Mellanox/network-operator/pkg/utils"
 )
 
+func checkResourceNameAnnotation(obj *unstructured.Unstructured) {
+	annotations := obj.Object["metadata"].(map[string]interface{})["annotations"].(map[string]interface{})
+	resourceName := annotations["k8s.v1.cni.cncf.io/resourceName"].(string)
+
+	Expect(strings.HasPrefix(resourceName, resourceNamePrefix)).To(Equal(true))
```
ykulazhenkov, line 38: nit: Equal(true) -> BeTrue()

pkg/state/state_hostdevice_network_test.go
```diff
@@ -29,6 +31,14 @@ import (
...
+	Expect(strings.Count(resourceName, resourceNamePrefix)).To(Equal(1))
```
ykulazhenkov, line 39: nit: Expect(strings.Count(resourceName, resourceNamePrefix)).To(Equal(1)) > Expect(resourceName).To(HavePrefix(resourceNamePrefix)) ?

moshe010 (approved):

PR #276, opened 2021-11-08
Update node labels regardless of state sync status

almaslennikov:
This fixes the issue with mofed.wait label not being set to false when other issues in deployment occur
The proposed solution is to update node labels regardless of state sync status

### DIFF
--- a/controllers/nicclusterpolicy_controller.go
+++ b/controllers/nicclusterpolicy_controller.go
@@ -117,11 +117,12 @@ func (r *NicClusterPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Req
 	}
 	// Create manager
 	managerStatus, err := r.stateManager.SyncState(instance, sc)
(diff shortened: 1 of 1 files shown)


controllers/nicclusterpolicy_controller.go
```diff
@@ -116,13 +116,9 @@ func (r *NicClusterPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Req
 		sc.Add(state.InfoTypeNodeInfo, infoProvider)
 	}
 	// Create manager
-	managerStatus, err := r.stateManager.SyncState(instance, sc)
+	managerStatus, _ := r.stateManager.SyncState(instance, sc)
```
moshe010, line 119: maybe we should just log it as a warning instead of ignoring it.

moshe010 (approved):
@e0ne  PTAL

e0ne (approved):

PR #285, opened 2021-11-30
Support extended selectors for sriov-device-plugin with helm deployment

almaslennikov:
Currently the SR-IOV Device Plugin resource definition can only select devices based on vendor ID. This creates problems in heterogeneous cluster deployments.
The proposed solution is to support all selectors from sriov-device-plugin.
The fix is intentionally aligned with the current implementation and not made more generic as that would require changing `values.yaml` and ensuring backwards compatibility with different versions of network-operator

### DIFF
--- a/deployment/network-operator/README.md
+++ b/deployment/network-operator/README.md
@@ -393,6 +393,13 @@ to be associated with the resource. Refer to [SR-IOV Network Device Plugin Selec
 resources:
     - name: hostdev
       vendors: [15b3]
+    - name: ethernet_rdma
+      vendors: [15b3]
+      linkTypes: [ether]
+    - name: sriov_rdma
+      vendors: [15b3]
+      devices: [1018]
+      drivers: [mlx5_ib]
 ``` 
 
 >__Note__: The parameter listed are non-exhaustive, for the full list of chart parameters refer to
--- a/deployment/network-operator/templates/mellanox.com_v1alpha1_nicclusterpolicy_cr.yaml
+++ b/deployment/network-operator/templates/mellanox.com_v1alpha1_nicclusterpolicy_cr.yaml
@@ -84,6 +84,12 @@ spec:
             "resourceName": {{ $element.name | quote }},
             "selectors": {
               "vendors": {{ $element.vendors | default list | toJson }},
+              "devices": {{ $element.devices | default list | toJson }},
+              "drivers": {{ $element.drivers | default list | toJson }},
+              "pfNames": {{ $element.pfNames | default list | toJson }},
+              "pciAddresses": {{ $element.pciAddresses | default list | toJson }},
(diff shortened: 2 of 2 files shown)


AbdYsn:
There was a power issue in the labs, retesting the CIs /retest-nic_operator_helm /retest-nic_operator_kind

e0ne (commented):
LGTM. Let's wait until CI passes and we cut v1.1.0 release

deployment/network-operator/README.md
```diff
@@ -393,6 +393,13 @@ to be associated with the resource. Refer to [SR-IOV Network Device Plugin Selec
 resources:
     - name: hostdev
       vendors: [15b3]
+    - name: infiniband_rdma
+      vendors: [15b3]
+      linkTypes: [infiniband]
```
adrianchiris, line 398: lets keep this example with ethernet

deployment/network-operator/values.yaml
```diff
@@ -158,6 +158,12 @@ sriovDevicePlugin:
   resources:
     - name: hostdev
       vendors: [15b3]
+    - name: infiniband_rdma
```
adrianchiris, line 161: This affects the default configuration right ? i dont think we want that.

also these 3 resources now collide 

1 includes 2 and 2 
2 may include 3

adrianchiris:
@almaslennikov also please fix commit message

deployment/network-operator/values.yaml
almaslennikov: Let's keep the default configuration as is and only change the examples in readme

deployment/network-operator/templates/mellanox.com_v1alpha1_nicclusterpolicy_cr.yaml
```diff
@@ -84,6 +84,10 @@ spec:
             "resourceName": {{ $element.name | quote }},
             "selectors": {
               "vendors": {{ $element.vendors | default list | toJson }},
+              "devices": {{ $element.devices | default list | toJson }},
+              "drivers": {{ $element.drivers | default list | toJson }},
+              "ifNames": {{ $element.ifNames | default list | toJson }},
```
adrianchiris, line 89: sriov network device plugin does not have this selector:  https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin#extended-selectors-for-device-type-netdevice

adrianchiris: https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin#device-selectors

adrianchiris: we should align to have the following selectors.

vendors, devices, drivers, pfNames, pciAddresses, rootDevices, linkTypes

Issue #304, opened 2022-02-08
dont use scheduler.alpha.kubernetes.io annotation


adrianchiris:
**What happened**:

scheduler.alpha.kubernetes.io is deprecated in k8s v1.16+. this emits a warning when deploying components which have this annotation

```
Warning: spec.template.metadata.annotations[scheduler.alpha.kubernetes.io/critical-pod]: non-functional in v1.16+; use the "priorityClassName" field instead
```

**What you expected to happen**:

No warning to be emitted, need to use `priorityClassName` instead

rollandf:
@adrianchiris can we close this issue?

adrianchiris:
Done. thx for working on it

PR #307, opened 2022-02-09
Set priorityClassName instead of annotation

rollandf:
scheduler.alpha.kubernetes.io is deprecated in k8s v1.16+.
This emits a warning when deploying components which have
this annotation:

Warning: spec.template.metadata.annotations
[scheduler.alpha.kubernetes.io/critical-pod]: non-functional in v1.16+;
use the "priorityClassName" field instead

Closes issue: #304


rollandf:
Updated according to https://kubernetes.io/docs/tasks/administer-cluster/guaranteed-scheduling-critical-addon-pods/

adrianchiris (approved):

PR #318, opened 2022-03-10
Change default startup probe time for ofed driver

adrianchiris:
In VM environments we see it can take up to 12 minutes
to complete driver compilation. Increase default stratup probe
retry period to 20s to allow for a total of 20 min before moving
the pod to failed state


e0ne (approved):

abeltre1:
@e0ne, I've encountered an issue where deploying a NicClusterPolicy to work alongside the network operator results in immediate failure within my OpenShift cluster. Is this behavior to be expected, considering the discussions in the comments above? Here's the configuration I'm using:

# Configuration 

```
startupProbe:
  initialDelaySeconds: 10
  periodSeconds: 20
livenessProbe:
  initialDelaySeconds: 30
  periodSeconds: 30
readinessProbe:
  initialDelaySeconds: 10
  periodSeconds: 30
```

Could this issue be related to the probe configurations, or is there another underlying cause I should be aware of?"


# Issue

Startup probe failed: ls: cannot access '/run/mellanox/drivers/.driver-ready': No such file or directory

PR #319, opened 2022-03-14
Add a script to update Network Operator dependencies

e0ne:


scripts/update-deps.sh
```diff
@@ -0,0 +1,126 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o pipefail
+set +e
+usage () {
+	#cat << EOF
```
rollandf, line 19: I think this is needed, no?

scripts/update-deps.sh
```diff
@@ -0,0 +1,126 @@
...
+	Usage: $this [-h] [--mofed] [--nvpeer] [--shared-dp] [--sriov-dp] [--cni-plugins] [--multus] [--whereabouts] <version>
```
rollandf, line 20: Does the `$this` works?

scripts/update-deps.sh
```diff
@@ -0,0 +1,126 @@
...
+Options:
+  -h         show this help and exit
```
rollandf, line 22: Maybe add indentation the same as the others

scripts/update-deps.sh
```diff
@@ -0,0 +1,126 @@
...
+  --modef          update MOFED image version
```
rollandf, line 23: Typo, should be `--mofed`

rollandf (changes requested):

scripts/update-deps.sh
e0ne: fixed

scripts/update-deps.sh
e0ne: fixed

scripts/update-deps.sh
e0ne: Such indentation is needed to be shown in CLI correctly

scripts/update-deps.sh
e0ne: thanks, fixed

scripts/update-deps.sh
```diff
@@ -0,0 +1,128 @@
...
+  --name "$(basename "$0")" \
+  --options "" \
+  -- "$@"
+)
+
+eval set --$opts
+
+
+new_version=$2
+while [[ $# -gt 0 ]]; do
```
adrianchiris, line 91: do you need the while loop ?
you can just use case statement as you are not shifting and exiting in every switch

perhaps check that $# equals 2 or fail otherwise first.

unless you want to support update of multiple components in the same call, then you need to modify this shell script to support it.

scripts/update-deps.sh
```diff
@@ -0,0 +1,128 @@
...
+
+opts=$(getopt \
+  --longoptions "$(printf "%s:," "${ARGUMENT_LIST[@]}")" \
+  --name "$(basename "$0")" \
+  --options "" \
+  -- "$@"
+)
+
+eval set --$opts
+
```
adrianchiris, line 88: nit: too many empty lines

adrianchiris:
@e0ne  once comments addressed can be merged IMO

@rollandf please give it an LGTM once you are happy

scripts/update-deps.sh
```diff
@@ -0,0 +1,126 @@
...
+	Usage: $this [-h] [--mofed] [--nvpeer] [--shared-dp] [--sriov-dp] [--cni-plugins] [--multus] [--whereabouts] <version>
+Options:
+  -h         show this help and exit
+  --mofed          update MOFED image version
+  --nvpeer         update nv-peer-mem image version
+  --shared-dp      update RDMA Shared Device Plugin image version
+  --sriov-dp       update SR-IOV Device Plugin image version
+  --cni-plugins    update CNI Plugins image version
+  --multus         update Multus CNI image version
+  --whereabouts    update Whereabouts IPAM Plugin image version
```
adrianchiris, line 31: @e0ne could you add also an example section with a run example just for clarity

scripts/update-deps.sh
```diff
@@ -0,0 +1,126 @@
...
+EOF
+}
+
+
+updateVersion() {
+	old=`grep $1 deployment/network-operator/values.yaml  -A 4 | grep version | awk '{ print $2 }'`
+	new=$2
+	grep -l -r $old ./deployment ./example | xargs sed -i 's/'"$old"'/'"$new"'/g'
```
adrianchiris, line 39: we could leverage kustomize to do the grunt work

e0ne:
Will be implemented in #450

PR #321, opened 2022-03-15
Use single namespace

rollandf:
In case the Network Operator is installed in OpenShift
via Operator Hub, the nvidia-network-operator-resources namespace
was not created automatically as with Helm.

With this patch, the Network Operator will create the underlying
resources in the same namespace it runs in.


e0ne (approved):
LGTM. deployment and upgrade verified manually

adrianchiris (approved):

adrianchiris:
LGTM !

adrianchiris:
@rollandf could you add a bit of a description in commit message on why this is needed ? 

then we can merge IMO

rollandf:
> @rollandf could you add a bit of a description in commit message on why this is needed ?
> 
> then we can merge IMO

@adrianchiris done

PR #326, opened 2022-03-29
Add SR-IOV Network Operator chart into the Network Operator repo

e0ne:
To fix [1] we need to use internal version of the SR-IOV Network
Operator chart until there will be a new release.

[1] https://github.com/k8snetworkplumbingwg/sriov-network-operator/issues/267


adrianchiris:
@rollandf PTAL

PR #338, opened 2022-04-15
Do not deploy SR-IOV Device Plugin by default

e0ne:
Macvlan Network with RDMA Shared Device Plugin should be deployed
by default.

HostDevice Network is be disabled in the default configuration now.


e0ne:
@adrianchiris @rollandf could you please take a look on it?

PR #340, opened 2022-04-17
Keep secrets for Service Account

rollandf:
In the reconcile loop, when applying Service Accounts
the secrets where deleted.
Then the service-account-token was recreated, causing
a leak of secrets object.


pkg/state/state_skel.go
```diff
@@ -135,6 +134,44 @@ func (s *stateSkel) createOrUpdateObjs(
...
+	gvk := updated.GroupVersionKind()
+	if gvk.Group == "" && gvk.Kind == "ServiceAccount" {
+		return s.mergeServiceAccount(updated, current)
+	}
+	return nil
+}
+
+// For Service Account, keep secrets if exists
+func (s *stateSkel) mergeServiceAccount(updated, current *unstructured.Unstructured) error {
+	curSecrets, ok, err := unstructured.NestedSlice(current.Object, "secrets")
```
adrianchiris, line 153: nit: we could save some code dupl if a `mergeNestedSlice(updated, current, sliceFieldName)` is defined.

not critical at this point, but if we end up adding more, this would need to be refactored.

adrianchiris (approved):
@e0ne do you think this is needed for current release ?

e0ne (approved):
LGTM

PR #341, opened 2022-04-29
Automatic driver upgrade

almaslennikov:
The PR adds an automatic driver upgrade feature to Network Operator.
If user makes changes to driver configuration in NicClusterPolicy, operator will start the process of automatic upgrade managed by user-set Upgrade Policy.

`upgrade-state` annotation is introduced to k8s node and helps to control the upgrade process.
UpgradeStateManager, based on the node's state, schedules different operation in the upgrade pipeline for that node.


controllers/upgrade_controller.go
```diff
@@ -0,0 +1,230 @@
...
+	err := r.List(ctx, nicClusterPolicyList)
+	if err != nil {
+		return nil, err
+	}
+
+	if (len(nicClusterPolicyList.Items)) == 0 {
+		return nil, errors.New("nicClusterPolicy not found")
+	}
+
+	nicClusterPolicy := nicClusterPolicyList.Items[0]
```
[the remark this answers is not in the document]
almaslennikov: I think that the flow would be more generic if we just look for driver daemon sets with specific labels. Not sure that this is needed, though. WDYT?

pkg/upgrade/upgrade_state.go
```diff
@@ -0,0 +1,243 @@
...
+	manager.PodRestartManager = NewPodRestartManager(k8sClient, log)
+
+	go manager.WatchForChanges()
+
+	return manager
+}
+
+func (m *ClusterUpgradeStateManager) WatchForChanges() {
+	for {
+		currentState := <-m.StateChannel
```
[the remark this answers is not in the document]
almaslennikov: My main goal was to do as little work as possible inside the reconcile loop. State manager processes each node and may also do some API calls using runtime client. WDYT, is this enough motivation to keep them separated via a channel?

api/v1alpha1/nicclusterpolicy_types.go
```diff
@@ -110,6 +110,41 @@ type PSPSpec struct {
...
+	// Enable indicates if node draining is allowed during upgrade
+	// +optional
+	// +kubebuilder:default:=true
+	Enable bool `json:"enable,omitempty"`
+	// Force indicates if force draining is allowed
+	// +optional
+	// +kubebuilder:default:=false
+	Force bool `json:"force,omitempty"`
+	// +optional
+	PodSelector string `json:"podSelector,omitempty"`
```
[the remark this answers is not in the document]
almaslennikov: @ykulazhenkov `drain.Helper` structure that is used in drain API, receives `PodSelector` as a single string. https://pkg.go.dev/k8s.io/kubectl/pkg/drain#Helper

pkg/upgrade/upgrade_state.go
```diff
@@ -0,0 +1,243 @@
...
+}
+
+func (m *ClusterUpgradeStateManager) ProcessReadyOrUnknownNodes(
+	currentClusterState *ClusterUpgradeState, nodeStateName string) {
+	m.Log.V(consts.LogLevelInfo).Info("ProcessReadyOrUnknownNodes")
+
+	for _, nodeState := range currentClusterState.NodeStates[nodeStateName] {
+		if GetPodGeneration(nodeState.DriverPod, m.Log) != currentClusterState.DriverDaemonSet.GetGeneration() {
+			_ = ChangeNodeUpgradeState(nodeState.Node, UpgradeRequired, m.K8sClient, m.Log)
+			return
```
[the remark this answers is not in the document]
almaslennikov: Thanks! It should be `continue`, not `return`

pkg/upgrade/upgrade_state.go
```diff
@@ -0,0 +1,243 @@
...
+	return manager
+}
+
+func (m *ClusterUpgradeStateManager) WatchForChanges() {
+	for {
+		currentState := <-m.StateChannel
+
+		m.Log.V(consts.LogLevelInfo).Info("State Manager, got state update")
+
+		if m.lastProcessedTimestamp.After(currentState.Timestamp) {
```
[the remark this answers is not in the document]
almaslennikov: While `StateManager` processes a single state, reconciler may yield several updates which may become outdated after the `StateManager` is done. By introducing timestamps, we can only process the relevant states

api/v1alpha1/nicclusterpolicy_types.go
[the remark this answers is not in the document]
ykulazhenkov: > @ykulazhenkov `drain.Helper` structure that is used in drain API, receives `PodSelector` as a single string. https://pkg.go.dev/k8s.io/kubectl/pkg/drain#Helper

Ok, thanks for the clarification, then it is totally fine to keep it as a string

pkg/upgrade/upgrade_state.go
```diff
@@ -0,0 +1,226 @@
...
+		DrainManager:      drainManager,
+		PodRestartManager: podRestartManager,
+		Log:               log,
+		K8sClient:         k8sClient,
+	}
+
+	return manager
+}
+
+func (m *ClusterUpgradeStateManager) ApplyState(ctx context.Context, currentState *ClusterUpgradeState) {
```
ykulazhenkov, line 70: This function should return error. Also, you should not ignore errors in ClusterUpgradeStateManager methods, for example when you failed to update node upgrade state in k8s API.

api/v1alpha1/nicclusterpolicy_types.go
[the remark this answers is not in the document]
adrianchiris: my main concern here is that its not clear what is the format of PodSelector.
if its well defined then we should use a structured type (and convert to string later) so it will be less error prone.

should it be of type[1] ?
or does the drain API only supports matching on labels ?

also i see DrainSpec.PodSelector is not being used in Drain manager. is that on purpose or just something left out ?


[1] https://github.com/k8snetworkplumbingwg/multi-networkpolicy/blob/220baf5d60c179716ed02d44cb0e646c76c1b605/vendor/k8s.io/apimachinery/pkg/apis/meta/v1/types.go#L1117

pkg/upgrade/upgrade_state.go
```diff
@@ -0,0 +1,226 @@
...
+	m.Log.V(consts.LogLevelInfo).Info("State Manager, got state update")
+
+	policy := currentState.Policy
+	if !policy.AutoUpgrade {
```
ykulazhenkov, line 74: Should we wipe upgrade annotations from all nodes in this case?

pkg/upgrade/upgrade_state.go
```diff
@@ -0,0 +1,226 @@
...
+		m.Log.V(consts.LogLevelInfo).Info("Driver auto upgrade is disabled, skipping")
+		return
+	}
+
+	m.Log.V(consts.LogLevelInfo).Info("Node states:",
```
ykulazhenkov, line 79: I think  we should add state validation logic at the beginning of this function. This logic should validate that annotation which was read from the Node object is really reflect current status. In this validation handler it should be possible to directly move node between states and update node annotation if we detect that information is outdated. Node annotation can contain outdated information for example if operator restarts or if we failed to update node annotation on the previous cycle.

pkg/upgrade/upgrade_state.go
```diff
@@ -0,0 +1,226 @@
...
+
+	upgradesAvailable := policy.MaxParallelUpgrades - upgradesInProgress
+
+	m.Log.V(consts.LogLevelInfo).Info("Upgrades in progress",
+		"currently in progress", upgradesInProgress,
+		"max parallel upgrades", policy.MaxParallelUpgrades,
+		"upgrade slots available", upgradesAvailable)
+
+	// First, check if unknown or ready nodes need to be upgraded
+	m.ProcessReadyOrUnknownNodes(currentState, Unknown)
```
ykulazhenkov, line 99: Please, do not ignore errors in all Process handlers. I thinks we should rise all critical errors up to Reconcile loop and rely on controller-runtime error handling logic which will schedule retry for us.

api/v1alpha1/nicclusterpolicy_types.go
```diff
@@ -110,18 +110,58 @@ type PSPSpec struct {
...
-	SecondaryNetwork       *SecondaryNetworkSpec `json:"secondaryNetwork,omitempty"`
-	PSP                    *PSPSpec              `json:"psp,omitempty"`
+	NodeAffinity           *v1.NodeAffinity       `json:"nodeAffinity,omitempty"`
+	OFEDDriver             *OFEDDriverSpec        `json:"ofedDriver,omitempty"`
+	NVPeerDriver           *NVPeerDriverSpec      `json:"nvPeerDriver,omitempty"`
+	RdmaSharedDevicePlugin *DevicePluginSpec      `json:"rdmaSharedDevicePlugin,omitempty"`
+	SriovDevicePlugin      *DevicePluginSpec      `json:"sriovDevicePlugin,omitempty"`
+	SecondaryNetwork       *SecondaryNetworkSpec  `json:"secondaryNetwork,omitempty"`
+	PSP                    *PSPSpec               `json:"psp,omitempty"`
+	OfedUpgradePolicy      *OfedUpgradePolicySpec `json:"upgradePolicy,omitempty"`
```
adrianchiris, line 164: should this be here or nested in OFEDDriverSpec ?

pkg/upgrade/util.go
```diff
@@ -0,0 +1,59 @@
...
+
+func ChangeNodeUpgradeState(
+	node *v1.Node, newNodeState string, k8sClient client.Client, log logr.Logger) error {
+	log.V(consts.LogLevelInfo).Info("Updating node state",
+		"node", node.Name,
+		"new state:", newNodeState)
+
+	node.SetManagedFields(nil)
+	node.Annotations[UpgradeStateAnnotation] = newNodeState
+	err := k8sClient.Update(context.TODO(), node)
```
ykulazhenkov, line 37: I think it is better to receive context as a ChangeNodeUpgradeState function parameter and use it for k8s API calls.

pkg/upgrade/util.go
```diff
@@ -0,0 +1,59 @@
...
+	"github.com/Mellanox/network-operator/pkg/consts"
+)
+
+func ChangeNodeUpgradeState(
+	node *v1.Node, newNodeState string, k8sClient client.Client, log logr.Logger) error {
+	log.V(consts.LogLevelInfo).Info("Updating node state",
+		"node", node.Name,
+		"new state:", newNodeState)
+
+	node.SetManagedFields(nil)
```
ykulazhenkov, line 35: Why we need to remove managed fields? Removing of the content of this field can be problematic, because this can break node object update logic in other controllers.

main.go
```diff
@@ -104,6 +104,17 @@ func main() {
 		setupLog.Error(err, "unable to create controller", "controller", "HostDeviceNetwork")
 		os.Exit(1)
 	}
+
+	upgradeLogger := ctrl.Log.WithName("controllers").WithName("Upgrade")
+
+	if err = (&controllers.UpgradeReconciler{
+		Client: mgr.GetClient(),
+		Log:    upgradeLogger,
```
adrianchiris: nit: can you directly create the log here ? for consistency with other controllers

pkg/upgrade/consts.go
```diff
@@ -0,0 +1,29 @@
...
+limitations under the License.
+*/
+
+package upgrade
+
+const (
+	UpgradeStateAnnotation     = "nvidia.com/ofed-upgrade-state"
+	PodTemplateGenerationLabel = "pod-template-generation"
+
+	Unknown         = ""
```
adrianchiris, line 23: These are the different upgrade states a node may be in right ?

can you add a comment explaining it  ? 

maybe prefix it with UpgradeState<the-state> ?

controllers/upgrade_controller.go
```diff
@@ -0,0 +1,231 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	for i := range podList.Items {
+		pod := &podList.Items[i]
+		r.Log.V(consts.LogLevelDebug).Info("Pod", "pod", pod.Name, "owner", pod.OwnerReferences[0].Name)
+
+		if pod.OwnerReferences == nil || len(pod.OwnerReferences) < 1 {
+			r.Log.V(consts.LogLevelWarning).Info("OFED Driver Pod has no owner DaemonSet", "pod", pod)
```
adrianchiris, line 125: can this happen for ofed pod ?

controllers/upgrade_controller.go
```diff
@@ -0,0 +1,231 @@
...
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	reqLogger.V(consts.LogLevelDebug).Info("Built upgrade state", "state", state)
+
+	return ctrl.Result{}, nil
+}
+
+func (r *UpgradeReconciler) BuildState(
```
adrianchiris, line 97: Nit: can you add docstring to the methods ?
(general comment for the entire PR)

controllers/upgrade_controller.go
```diff
@@ -86,7 +88,8 @@ func (r *UpgradeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		return ctrl.Result{}, err
 	}
```
adrianchiris, line 90: nit, maybe add debug log for the generated state ?

main.go
```diff
@@ -106,11 +107,12 @@ func main() {
 	}
 
 	upgradeLogger := ctrl.Log.WithName("controllers").WithName("Upgrade")
-
+	clusterUpdateStateManager := upgrade.NewClusterUpdateStateManager(upgradeLogger, mgr.GetClient())
```
adrianchiris: you could pass here : upgradeLogger.WithName("manager")
so its clear from logs which component did the logging. WDYT?

controllers/upgrade_controller.go
```diff
@@ -86,7 +88,8 @@ func (r *UpgradeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
 
-	reqLogger.V(consts.LogLevelDebug).Info("Built upgrade state", "state", state)
+	reqLogger.V(consts.LogLevelInfo).Info("Propagate state to state manager")
+	r.StateManager.ApplyState(ctx, state)
```
adrianchiris, line 92: this never fails ?

pkg/upgrade/upgrade_state.go
adrianchiris: same question would also apply if the policy in the CRD is nil

adrianchiris: perhaps cleanup should be called at the controller level in this case

if autoupgrade is disabled or policy is nil

pkg/upgrade/upgrade_state.go
```diff
@@ -38,3 +44,167 @@ type ClusterUpgradeState struct {
...
+				// TODO uncordon
+			}
+			_ = ChangeNodeUpgradeState(nodeState.Node, Ready, m.K8sClient, m.Log)
+		}
+	}
+}
+
+func GetPodGeneration(pod *v1.Pod, log logr.Logger) int64 {
+	generation, err := strconv.ParseInt(pod.Labels[PodTemplateGenerationLabel], 10, 0)
+	if err != nil {
```
adrianchiris, line 200: error here is really unlikely to happen right ?

because if it does i think we will start draining the node

pkg/upgrade/util.go
```diff
@@ -40,3 +43,17 @@ func ChangeNodeUpgradeState(
 
 	return err
 }
+
+func CreateK8sInterface() (kubernetes.Interface, error) {
```
adrianchiris, line 47: i think its fine to either move it to main.go or to a top level util package.

pkg/upgrade/util.go
```diff
@@ -40,3 +43,17 @@ func ChangeNodeUpgradeState(
...
+	config, err := rest.InClusterConfig()
```
adrianchiris, line 48: you can use controller runtime ctrl.GetConfigOrDie()
which allows to override kubeconfig via flag or env variable

main.go
almaslennikov: The idea was to create a single logger for all upgrade-related entities for easy logs filtering. That way the logger is created in main and propagated to all other components

pkg/upgrade/util.go
almaslennikov: I've added it to reduce the rate at which "Operation cannot be fulfilled on nodes the object has been modified" occurs but now I've changed the solution to re-requesting the object and trying again if it happens. So `node.SetManagedFields(nil)` code is removed

controllers/upgrade_controller.go
almaslennikov: It shouldn't but I'd still check for this case to avoid potential access violation

controllers/upgrade_controller.go
```diff
@@ -0,0 +1,234 @@
...
+
+		if pod.OwnerReferences == nil || len(pod.OwnerReferences) < 1 {
+			r.Log.V(consts.LogLevelWarning).Info("OFED Driver Pod has no owner DaemonSet", "pod", pod)
+			continue
+		}
+		ownerDaemonSet, ok := daemonSets[pod.OwnerReferences[0].UID]
+
+		if !ok {
+			r.Log.V(consts.LogLevelWarning).Info("OFED Driver Pod is not owned by an OFED Driver DaemonSet",
+				"pod", pod, "owner", ownerDaemonSet)
```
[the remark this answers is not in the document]
almaslennikov: Thanks, makes sense

api/v1alpha1/nicclusterpolicy_types.go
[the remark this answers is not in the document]
adrianchiris: ack, can you add a comment on it ? so its clear what is the expected content of this field

main.go
```diff
@@ -107,8 +108,15 @@ func main() {
...
+	k8sInterface, err := utils.CreateK8sInterface()
+	if err != nil {
+		setupLog.Error(err, "unable to create k8s interface", "controller", "Upgrade")
+		os.Exit(1)
+	}
+	drainManager := upgrade.NewDrainManager(mgr.GetClient(), k8sInterface, upgradeLogger.WithName("drainManager"))
+	podRestartManager := upgrade.NewPodRestartManager(mgr.GetClient(), upgradeLogger.WithName("podRestartManager"))
 	clusterUpdateStateManager := upgrade.NewClusterUpdateStateManager(
-		upgradeLogger.WithName("manager"), mgr.GetClient())
+		drainManager, podRestartManager, upgradeLogger.WithName("manager"), mgr.GetClient(), k8sInterface)
```
adrianchiris, line 119: nice ! good job partitioning the logs like this !

pkg/upgrade/upgrade_state.go
```diff
@@ -0,0 +1,279 @@
...
+
+	pods := make([]*v1.Pod, 0, len(currentClusterState.NodeStates[UpgradeStatePodRestart]))
+	for _, nodeState := range currentClusterState.NodeStates[UpgradeStatePodRestart] {
+		if GetPodGeneration(nodeState.DriverPod, m.Log) != nodeState.DriverDaemonSet.GetGeneration() {
+			// Pods should only be scheduled for restart if they are not terminating or restarting already
+			if nodeState.DriverPod.Status.Phase == "Running" {
+				pods = append(pods, nodeState.DriverPod)
+			}
+		} else if nodeState.DriverPod.Status.Phase == "Running" {
+			err := ChangeNodeUpgradeState(ctx, nodeState.Node, UpgradeStateReady, m.K8sClient, m.Log)
```
adrianchiris, line 219: maybe im missing something, but are we missing an upgrade state here (driverPodReadyUncordonNeeded or similar) ?

if a pod is restarted and now matches the generation of the daemonset, and it is now running. when will uncordon happen ? since here we change the upgrade state of the node to ready.

since in ProcessUncordon, we only uncordon pod restart and drain failed states.

controllers/upgrade_controller.go
```diff
@@ -0,0 +1,277 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	mellanoxv1alpha1 "github.com/Mellanox/network-operator/api/v1alpha1"
+	"github.com/Mellanox/network-operator/pkg/consts"
+	"github.com/Mellanox/network-operator/pkg/upgrade"
+)
+
+const OfedDriverLabel = "nvidia.com/ofed-driver"
```
ykulazhenkov, line 44: Should we add this label to the OFED DaemonSet template?

almaslennikov: Yes we should both for OFED DaemonSet and Pods

ykulazhenkov: I mean do we need to set it here, in the template which is used to create DS? https://github.com/almaslennikov/network-operator/blob/818565321afeafc00da8794683c8f7d39bc4b182/manifests/stage-ofed-driver/0050_ofed-driver-ds.yaml#L29

Maybe I miss something but I can't see where you set these labels in the current PR version.

deployment/values.yaml
```diff
@@ -0,0 +1,185 @@
+# Copyright 2020 NVIDIA
```
e0ne, line 1: this file shouldn't be pushed to the repo

controllers/upgrade_controller.go
```diff
@@ -0,0 +1,285 @@
...
```
e0ne, line 44: Should we move it to upgrade/consts.go?

e0ne (changes requested):

pkg/upgrade/upgrade_state.go
ykulazhenkov: Was discussed offline. We agreed that each stage can optionally include extra steps to validate current state before applying any actions. No need for generic validation function.

pkg/upgrade/upgrade_state.go
```diff
@@ -0,0 +1,328 @@
...
+
+	return nil
+}
+
+// ProcessDrainNodes schedules UpgradeStateDrain nodes for drain.
+// If drain is disabled by upgrade policy, moves the nodes straight to UpgradeStatePodRestart state.
+func (m *ClusterUpgradeStateManager) ProcessDrainNodes(
+	ctx context.Context, currentClusterState *ClusterUpgradeState, drainSpec *v1alpha1.DrainSpec) error {
+	m.Log.V(consts.LogLevelInfo).Info("ProcessDrainNodes")
+	if !drainSpec.Enable {
```
ykulazhenkov, line 222: I think drainSpec can be nil here? Is it?

main.go
```diff
@@ -104,6 +106,31 @@ func main() {
...
+	if err != nil {
+		setupLog.Error(err, "unable to create k8s interface", "controller", "Upgrade")
+		os.Exit(1)
+	}
+	nodeUpgradeStateProvider := upgrade.NewNodeUpgradeStateProvider(
+		mgr.GetClient(), upgradeLogger.WithName("nodeUpgradeStateProvider"))
+	drainManager := upgrade.NewDrainManager(
+		k8sInterface, nodeUpgradeStateProvider, upgradeLogger.WithName("drainManager"))
+	podRestartManager := upgrade.NewPodRestartManager(mgr.GetClient(), upgradeLogger.WithName("podRestartManager"))
+	clusterUpdateStateManager := upgrade.NewClusterUpdateStateManager(
```
ykulazhenkov, line 121: As far as I can see you don't use k8sInterface in ClusterUpdateStateManager. Please, remove if it is not required.

pkg/upgrade/pod_restart_manager.go
```diff
@@ -0,0 +1,61 @@
...
+// SchedulePodsRestart receives a list of pods and schedules to delete them
+func (m *PodRestartManagerImpl) SchedulePodsRestart(ctx context.Context, pods []*corev1.Pod) error {
+	m.Log.V(consts.LogLevelInfo).Info("Starting Pod Restart")
+	if len(pods) == 0 {
+		m.Log.V(consts.LogLevelInfo).Info("Pod Restart Manager, no driver pods scheduled to restart")
+		return nil
+	}
+	for _, pod := range pods {
+		pod := pod
+		go func() {
```
ykulazhenkov, line 47: Maybe we can run Delete command for all PODs in the main loop? I think that with current implementation we may spawn goroutine for the same Pod more then once (new goroutine will be created by each following Reconcile cycle until Pod will be in ready state)

main.go
almaslennikov: It's used for node uncordon

pkg/upgrade/pod_restart_manager.go
almaslennikov: Pods will only be scheduled for deletion if they are outdated and are not in "Terminating" state. The pod deletion can take quite a long time so I would prefer to still run it in a separate goroutine

ykulazhenkov: Delete call to k8s API will return immediately after DeletionTimestamp is set for resource, it will not wait for resource termination or complete removal, so it should return quite fast.

main.go
ykulazhenkov: ack

deployment/network-operator/values.yaml
```diff
@@ -128,6 +128,23 @@ ofedDriver:
   readinessProbe:
     initialDelaySeconds: 10
     periodSeconds: 30
+  upgradePolicy:
+    # global switch for automatic upgrade feature
+    # if set to false all other options are ignored
+    autoUpgrade: true
```
adrianchiris, line 134: lets have the feature disabled by default.

once we get some milage on it we can enable

adrianchiris:
@almaslennikov lets also add documentation for this feature under `docs/`
then reference it in the main README under upgrade section.

pkg/upgrade/upgrade_state.go
```diff
@@ -0,0 +1,344 @@
...
+		}
+
+		// And each container is ready
+		return true
+	}
+
+	return false
+}
+
+func GetPodGeneration(pod *v1.Pod, log logr.Logger) int64 {
```
adrianchiris, line 337: GetPodTemplateGeneration ?

controllers/upgrade_controller.go
```diff
@@ -0,0 +1,277 @@
...
+		r.Log.V(consts.LogLevelError).Error(err, "Failed to apply cluster upgrade state")
+		return ctrl.Result{}, err
+	}
+
+	return ctrl.Result{}, nil
+}
+
+// resetNodeUpgradeAnnotations iterates over nodes in the cluster and removes upgrade.UpgradeStateAnnotation annotation
+// It is used for cleanup when autoUpgrade feature gets disabled
+func (r *UpgradeReconciler) resetNodeUpgradeAnnotations(ctx context.Context) error {
```
adrianchiris, line 105: nit: removeNodeUpgradeAnnotations ?

pkg/upgrade/upgrade_state.go
almaslennikov: Changed the function to also return an error to be on the safe side

pkg/upgrade/util.go
```diff
@@ -0,0 +1,175 @@
...
+func (m *KeyedMutex) Lock(key string) func() {
+	value, _ := m.mutexes.LoadOrStore(key, &sync.Mutex{})
+	mtx := value.(*sync.Mutex)
+	mtx.Lock()
+	return func() { mtx.Unlock() }
+}
+
+// NodeUpgradeStateProvider allows for synchronized operations on node objects and ensures that the node,
+// got from the provider, always has the up-to-date upgrade state
+type NodeUpgradeStateProvider struct {
```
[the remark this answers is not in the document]
almaslennikov: I don't think this would bring much benefit: we don't need to synchronize access to node operations in other parts of the project and since it's only used for the upgrade flow I think more explicit API is better for code readability

pkg/upgrade/util.go
```diff
@@ -0,0 +1,175 @@
...
+		if nodeState != newNodeState {
+			p.Log.V(consts.LogLevelDebug).Info("upgrade state annotation for node doesn't match the expected",
+				"node", node.Name, "expected", newNodeState, "actual", nodeState)
+			return false, nil
+		}
+		return true, nil
+	}, timeoutCtx.Done())
+
+	if err != nil {
+		p.Log.V(consts.LogLevelError).Error(err, "Timeout exceeded on waiting to apply node state to operator cache",
```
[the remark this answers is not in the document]
almaslennikov: Thanks, changed it

pkg/upgrade/pod_restart_manager.go
```diff
@@ -0,0 +1,74 @@
...
+
+	"github.com/Mellanox/network-operator/pkg/consts"
+)
+
+type PodRestartManagerImpl struct {
+	InputPodChannel chan []*corev1.Pod
+	K8sClient       client.Client
+
+	Log            logr.Logger
+	internalCtx    context.Context
```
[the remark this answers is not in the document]
almaslennikov: Dropped internalCtx entirely and doing deletes synchronously now

pkg/upgrade/upgrade_state.go
almaslennikov: Added UncordonRequired state

pkg/upgrade/upgrade_state.go
```diff
@@ -0,0 +1,328 @@
...
+}
+
+// ProcessDrainFailedNodes processes UpgradeStateDrainFailed nodes and checks whether the driver pod on the node
+// has been successfully restarted. If the pod is in Ready state - moves the node to UpgradeStateUncordonRequired state.
+func (m *ClusterUpgradeStateManager) ProcessDrainFailedNodes(
+	ctx context.Context, currentClusterState *ClusterUpgradeState) error {
+	m.Log.V(consts.LogLevelInfo).Info("ProcessDrainFailedNodes")
+
+	for _, nodeState := range currentClusterState.NodeStates[UpgradeStateDrainFailed] {
+		if GetPodGeneration(nodeState.DriverPod, m.Log) == nodeState.DriverDaemonSet.GetGeneration() &&
```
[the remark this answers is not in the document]
almaslennikov: Done

pkg/upgrade/upgrade_state.go
almaslennikov: Added check

pkg/upgrade/pod_restart_manager.go
almaslennikov: Done

api/v1alpha1/nicclusterpolicy_types.go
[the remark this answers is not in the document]
almaslennikov: See the updated description

api/v1alpha1/nicclusterpolicy_types.go
```diff
@@ -110,6 +110,41 @@ type PSPSpec struct {
...
+	Enable bool `json:"enable,omitempty"`
+	// Force indicates if force draining is allowed
+	// +optional
+	// +kubebuilder:default:=false
+	Force bool `json:"force,omitempty"`
+	// +optional
+	PodSelector string `json:"podSelector,omitempty"`
+	// +optional
+	// +kubebuilder:default:=0
+	TimeoutSecond int `json:"timeoutSeconds,omitempty"`
```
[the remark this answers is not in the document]
almaslennikov: See the updated description

api/v1alpha1/nicclusterpolicy_types.go
```diff
@@ -110,6 +110,41 @@ type PSPSpec struct {
...
+	// +optional
+	// +kubebuilder:default:=false
+	DeleteEmptyDir bool `json:"deleteEmptyDir,omitempty"`
```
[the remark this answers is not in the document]
almaslennikov: See the updated description

controllers/upgrade_controller.go
```diff
@@ -0,0 +1,230 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	if (len(nicClusterPolicyList.Items)) == 0 {
+		return nil, errors.New("nicClusterPolicy not found")
+	}
+
+	nicClusterPolicy := nicClusterPolicyList.Items[0]
+	upgradePolicy := nicClusterPolicy.Spec.UpgradePolicy
```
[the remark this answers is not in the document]
almaslennikov: Added check for UpgradePolicy existence and whether it's enabled

controllers/upgrade_controller.go
```diff
@@ -0,0 +1,230 @@
...
+	}
+
+	return &upgrade.NodeUpgradeState{Node: &node, DriverPod: pod}, nil
+}
+
+func (r *UpgradeReconciler) getDriverDaemonSet(ctx context.Context) (*appsv1.DaemonSet, error) {
+	// Get list of driver pods
+	daemonSetList := &appsv1.DaemonSetList{}
+
+	err := r.List(ctx, daemonSetList, client.InNamespace(consts.NetworkOperatorResourceNamespace))
```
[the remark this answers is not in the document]
almaslennikov: Done, thanks

pkg/state/state_ofed.go
```diff
@@ -118,6 +124,42 @@ func (s *stateOFED) Sync(customResource interface{}, infoCatalog InfoCatalog) (S
...
+		podTemplateGeneration, err := utils.GetPodTemplateGeneration(pod, log)
+		if err != nil {
+			return SyncStateNotReady, errors.Wrap(err, "failed to get pod template generation")
+		}
+
+		// If pod and ds generations do not match, it means that pod is not up-to-date
+		if podTemplateGeneration != ds.GetGeneration() {
+			return SyncStateNotReady, nil
+		}
+	}
```
almaslennikov, line 162: Please take a look at this change

pkg/upgrade/drain_manager.go
```diff
@@ -0,0 +1,134 @@
...
+	Log logr.Logger
+}
+
+// DrainManager is an interface that allows to schedule nodes drain based on DrainSpec
+type DrainManager interface {
+	ScheduleNodesDrain(ctx context.Context, drainConfig *DrainConfiguration) error
+}
+
+// ScheduleNodesDrain receives DrainConfiguration and schedules drain for each node in the list.
+// When the node gets scheduled, it's marked as being drained and therefore will not be scheduled for drain twice
```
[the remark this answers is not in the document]
almaslennikov: Yes, so that we don't run drain if node is already draining

pkg/upgrade/upgrade_state.go
almaslennikov: Yes, thanks

pkg/upgrade/pod_restart_manager.go
```diff
@@ -0,0 +1,74 @@
...
+		m.Log.V(consts.LogLevelInfo).Info("Starting Pod Restart")
+		if len(pods) == 0 {
+			m.Log.V(consts.LogLevelInfo).Info("Pod Restart Manager, no driver pods scheduled to restart")
+			continue
+		}
+		for _, pod := range pods {
+			pod := pod
+			go func() {
+				m.Log.V(consts.LogLevelInfo).Info("Deleting pod", "pod", pod.Name)
+				err := m.K8sClient.Delete(m.internalCtx, pod)
```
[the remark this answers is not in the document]
almaslennikov: Yes it will and the reconcile will be rescheduled. Although, the chance is not too high as the pod would be recreated shortly after deletion

deployment/network-operator/values.yaml
almaslennikov: Makes sense, done

controllers/upgrade_controller.go
```diff
@@ -0,0 +1,234 @@
...
+				"pod", pod, "owner", ownerDaemonSet)
+			continue
+		}
+
+		nodeState, err := r.buildNodeUpgradeState(ctx, pod, ownerDaemonSet)
+		if err != nil {
+			return nil, err
+		}
+		nodeStateAnnotation := nodeState.Node.Annotations[upgrade.UpgradeStateAnnotation]
+		upgradeState.NodeStates[nodeStateAnnotation] = append(
```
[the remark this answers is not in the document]
ykulazhenkov: not applicable anymore, feel free to mark resolved

deployment/network-operator/values.yaml
almaslennikov: Oops, thanks for noticing - now it's changed

pkg/state/state_ofed.go
almaslennikov: Thanks for the suggestion, I've checked that it works and applied to stateSkel

deployment/network-operator/values.yaml
ykulazhenkov: Now it is fine. Thanks

ykulazhenkov:
We have upgrade section in helm charts documentation https://github.com/Mellanox/network-operator/blob/master/deployment/network-operator/README.md#upgrade. I think we need to add some cross references between this doc and your.

ykulazhenkov (approved):

almaslennikov:
> We have upgrade section in helm charts documentation https://github.com/Mellanox/network-operator/blob/master/deployment/network-operator/README.md#upgrade. I think we need to add some cross references between this doc and your.

@ykulazhenkov I've added a section on the automatic OFED upgrade to this file already. Please, take a look

ykulazhenkov:
> > We have upgrade section in helm charts documentation https://github.com/Mellanox/network-operator/blob/master/deployment/network-operator/README.md#upgrade. I think we need to add some cross references between this doc and your.
> 
> @ykulazhenkov I've added a section on the automatic OFED upgrade to this file already. Please, take a look

ha, now I see. I missed this change somehow. Thanks for the clarification

pkg/upgrade/upgrade_state.go
```diff
@@ -0,0 +1,385 @@
...
+			return err
+		}
+		if podTemplateGeneration != nodeState.DriverDaemonSet.GetGeneration() {
+			err := m.NodeUpgradeStateProvider.ChangeNodeUpgradeState(ctx, nodeState.Node, UpgradeStateUpgradeRequired)
+			if err != nil {
+				m.Log.V(consts.LogLevelError).Error(
+					err, "Failed to change node upgrade state", "state", UpgradeStateUpgradeRequired)
+				return err
+			}
+			return nil
```
adrianchiris, line 196: you are returning on the first  node which has its mofed pod generation different than the daemonset

is that the desired behavior ? i think you would want to process all nodes in the specified state.

pkg/upgrade/upgrade_state.go
```diff
@@ -0,0 +1,385 @@
...
+		}
+
+		if nodeStateName == UpgradeStateUnknown {
```
adrianchiris, line 199: else if ?

pkg/upgrade/upgrade_state.go
```diff
@@ -0,0 +1,344 @@
...
+			}
+		} else if nodeStateName == UpgradeStateUnknown {
+			err := m.NodeUpgradeStateProvider.ChangeNodeUpgradeState(ctx, nodeState.Node, UpgradeStateDone)
+			if err != nil {
+				m.Log.V(consts.LogLevelError).Error(
+					err, "Failed to change node upgrade state", "state", UpgradeStateDone)
+				return err
+			}
+		}
+		m.Log.V(consts.LogLevelInfo).Info("Node in done state, upgrade not required",
```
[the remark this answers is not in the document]
adrianchiris: in regards to logging, what id like to see is log message for node that changed state in this method.

pkg/upgrade/upgrade_state.go
adrianchiris: or replace return with continue (and add logs) in L196

pkg/upgrade/upgrade_suit_test.go
```diff
@@ -0,0 +1,175 @@
...
+		On("ChangeNodeUpgradeState", mock.Anything, mock.Anything, mock.Anything).
+		Return(func(ctx context.Context, node *corev1.Node, newNodeState string) error {
+			node.Annotations[upgrade.UpgradeStateAnnotation] = newNodeState
+			return nil
+		})
+
+	drainManager = mocks.DrainManager{}
+	drainManager.
+		On("ScheduleNodesDrain", mock.Anything, mock.Anything).
+		Return(func(context.Context, *upgrade.DrainConfiguration) error {
```
adrianchiris, line 98: Return(nil) doesnt work here ?

the mocks support either returning specific values or using method as you did here.
if there is no special logic you want to do, then you can simplify and just use the desired return value explicitly.

I see you did that across all tests.

docs/automatic-ofed-upgrade.md
```diff
@@ -0,0 +1,73 @@
...
+* `drain-failed` is set when drain on the node has failed. Manual interaction is required at this stage. See [Troubleshooting](#node-is-in-drain-failed-state) section for more details.
+* `uncordon-required` is set when OFED POD on the node is up-to-date and has "Ready" status. After uncordone the state is changed to `upgrade-done`
+
+#### State change diagram
+
+![State change diagram](ofed-upgrade-state-change-diagram.png)
+
+### Troubleshooting
+#### Node is in `drain-failed` state
+* Drain the node manually by running `kubectl drain <node_name> --ignore-daemonsets`
```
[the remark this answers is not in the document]
almaslennikov: `kubectl drain` also cordons the node

deployment/network-operator/values.yaml
```diff
@@ -156,6 +156,23 @@ ofedDriver:
   readinessProbe:
     initialDelaySeconds: 10
     periodSeconds: 30
+  upgradePolicy:
```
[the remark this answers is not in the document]
almaslennikov: Done

pkg/upgrade/node_upgrade_state_provider.go
```diff
@@ -0,0 +1,119 @@
+/*
```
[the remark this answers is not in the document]
almaslennikov: It was added previously but it was located in `util_test.go`. For clarity, moved UT from there to their respective files

pkg/upgrade/uncordon_manager.go
```diff
@@ -0,0 +1,45 @@
+/*
```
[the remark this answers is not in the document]
almaslennikov: same as above

api/v1alpha1/nicclusterpolicy_types.go
```diff
@@ -110,6 +112,47 @@ type PSPSpec struct {
...
+}
+
+// OfedUpgradePolicySpec describes policy configuration for automatic upgrades
+type OfedUpgradePolicySpec struct {
+	// AutoUpgrade is a global switch for automatic upgrade feature
+	// if set to false all other options are ignored
+	// +optional
+	// +kubebuilder:default:=true
```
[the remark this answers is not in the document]
almaslennikov: Yep, thanks

docs/automatic-ofed-upgrade.md
```diff
@@ -0,0 +1,73 @@
...
+apiVersion: mellanox.com/v1alpha1
+kind: NicClusterPolicy
+metadata:
+  name: nic-cluster-policy
+  namespace: nvidia-network-operator
+spec:
+  ofedDriver:
+    image: mofed
+    repository: mellanox
+    version: 5.3-1.0.0.1
```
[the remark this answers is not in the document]
almaslennikov: Changed it to the latest one

pkg/upgrade/upgrade_state.go
```diff
@@ -0,0 +1,385 @@
...
+	}
+	// Start upgrade process for upgradesAvailable number of nodes
+	err = m.ProcessUpgradeRequiredNodes(ctx, currentState, upgradesAvailable)
+	if err != nil {
+		m.Log.V(consts.LogLevelError).Error(
+			err, "Failed to process nodes", "state", UpgradeStateUpgradeRequired)
+		return err
+	}
+	// Schedule nodes for drain
+	err = m.ProcessDrainNodes(ctx, currentState, upgradePolicy.DrainSpec)
```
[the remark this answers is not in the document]
almaslennikov: The reconcile loop would probably be triggered by node/ds update but for safety I added a periodic reconcile every 2 minutes with an explaining comment in `controllers/upgrade_controller.go`

pkg/upgrade/upgrade_state.go
almaslennikov: No, thanks for noticing. Replaced the `return` with `continue`

pkg/upgrade/upgrade_state.go
almaslennikov: Added more logs

pkg/upgrade/upgrade_suit_test.go
almaslennikov: Thanks, it works

controllers/upgrade_controller.go
adrianchiris: @almaslennikov  shall we change this ?

docs/automatic-ofed-upgrade.md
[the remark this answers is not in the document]
adrianchiris: right !

pkg/upgrade/node_upgrade_state_provider.go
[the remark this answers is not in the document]
adrianchiris: super thx !

pkg/upgrade/upgrade_state.go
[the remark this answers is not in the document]
adrianchiris: ds will not update in this case, we will be depentent on external triggers.

thx for adding the periodic reconcile. i think it may also solve issues that were not found in the review :)

pkg/upgrade/upgrade_state.go
```diff
@@ -0,0 +1,389 @@
...
+			err := m.NodeUpgradeStateProvider.ChangeNodeUpgradeState(ctx, nodeState.Node, UpgradeStateDone)
+			if err != nil {
+				m.Log.V(consts.LogLevelError).Error(
+					err, "Failed to change node upgrade state", "state", UpgradeStateDone)
+				return err
+			}
+			m.Log.V(consts.LogLevelInfo).Info("Changed node state to UpgradeDOne",
+				"node", nodeState.Node.Name)
+		}
+		m.Log.V(consts.LogLevelInfo).Info("Node in UpgradeDone state, upgrade not required",
```
adrianchiris, line 211: can you change log level to debug here ?

also maybe add `continue` after L209 so you dont log the same info twice in case node state was unknown and its state was changed to done

almaslennikov: Changed the log level and added continue in the previous if statement

adrianchiris (approved):

PR #352, opened 2022-05-19
Update OCP bundle for Network Operator v1.2.0

e0ne:
This patch also deletes v1.1.0 bundle since we don't
need it anymore after introducing release branches.


e0ne:
We don't test OCP bundle in our CI so don't need to wait for it

PR #353, opened 2022-05-23
Use global imagePullSecrets values param for NicClusterPolicy

e0ne:


rollandf:
Looks good.
Can we an example/description of the new param in the Helm README?
https://github.com/Mellanox/network-operator/blob/master/deployment/network-operator/README.md

rollandf:
Looks good.
Can we an example/description of the new param in the Helm README?
https://github.com/Mellanox/network-operator/blob/master/deployment/network-operator/README.md

e0ne:
> Looks good. Can we add an example/description of the new param in the Helm README? https://github.com/Mellanox/network-operator/blob/master/deployment/network-operator/README.md

This patch doesn't introduce new params. It fixes usage of current

rollandf:
LGTM

deployment/network-operator/README.md
```diff
@@ -302,16 +302,17 @@ We have introduced the following Chart parameters.
...
-| `operator.imagePullSecrets` | list | `[]` | An optional list of references to secrets to use for pulling any of the Network Operator image |
-| `deployCR` | bool | `false` | Deploy `NicClusterPolicy` custom resource according to provided parameters |
+| Name | Type | Default | description                                                                                                          |
+| ---- | ---- | ------- |----------------------------------------------------------------------------------------------------------------------|
+| `nfd.enabled` | bool | `True` | deploy Node Feature Discovery                                                                                        |
+| `sriovNetworkOperator.enabled` | bool | `False` | deploy SR-IOV Network Operator                                                                                       |
+| `psp.enabled` | bool | `False` | deploy Pod Security Policy                                                                                           |
+| `imagePullSecrets` | list | `[]` | An optional list of references to secrets to use for pulling any of the Network Operator image if it's not overrided |
```
adrianchiris, line 310: can you also add a short example in the description ?

deployment/network-operator/templates/_helpers.tpl
```diff
@@ -65,3 +65,9 @@ Selector labels
 app.kubernetes.io/name: {{ include "network-operator.name" . }}
 app.kubernetes.io/instance: {{ .Release.Name }}
 {{- end }}
+
+{{- define "network-operator.imagePullSecret" }}
+{{- if .Values.imagePullSecrets }}
+imagePullSecrets: {{ .Values.imagePullSecrets }}
+{{- end }}
+{{- end }}
```
adrianchiris, line 73: nit: missing newline

deployment/network-operator/templates/_helpers.tpl
```diff
@@ -65,3 +65,9 @@ Selector labels
 app.kubernetes.io/name: {{ include "network-operator.name" . }}
 app.kubernetes.io/instance: {{ .Release.Name }}
 {{- end }}
+
+{{- define "network-operator.imagePullSecret" }}
```
adrianchiris, line 69: im thinking we should have a defined template for every component that supports image pull secret.
then in the relevant template (e.g operator, nicclusterpolicy) we just use the respective defined template

e.g

`network-operator.operator.imagePullSecrets`
`network-operator.sriovDevicePlugin.imagePullSecrets`
`network-operator.secondaryNetwork.multus.imagePullSecrets`

each will look roughly the same:

```
{{- define "network-operator.sriovDevicePlugin.imagePullSecrets" }}
{{- if .Values.sriovDevicePlugin.imagePullSecrets }}
.Values.sriovDevicePlugin.imagePullSecrets
{{- else }}
.Values.imagePullSecrets
{{- end}}
{{- end }}
```

then you just use that in the templates
e.g in nicclusterpolicy template, sriovDevicePlugin section
```
  sriovDevicePlugin:
    image: {{ .Values.sriovDevicePlugin.image }}
    repository: {{ .Values.sriovDevicePlugin.repository }}
    version: {{ .Values.sriovDevicePlugin.version }}
    imagePullSecrets: {{ include "network-operator.sriovDevicePlugin.imagePullSecrets" . | nindent 4 }}
```

if both the global and the per resource imagePullSecrets is not specified it would default to empty list (as expected)

note: didnt test the above so some tweaks may be needed :)

WDYT ?

deployment/network-operator/values.yaml
```diff
@@ -106,6 +106,8 @@ operator:
   # tag, if defined will use the given image tag, else Chart.AppVersion will be used
   # tag
 
+imagePullSecrets: []
```
adrianchiris, line 109: we need to support image pull secrets in sriov network operator as well

adrianchiris: This can be achieved (if imagePullSecrets cannot be easily set in sriov-network-operator and its config daemon)
by adding imagePullSecrets to service account:

https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#add-imagepullsecrets-to-a-service-account

e0ne: NFD already implemented this https://github.com/kubernetes-sigs/node-feature-discovery/blob/566957624036d172a0d39081d10193cc09cfdeb2/deployment/helm/node-feature-discovery/values.yaml#L7

deployment/network-operator/templates/_helpers.tpl
e0ne: it moves all logic from CR template to some functions. It simplifies CR template which is more actively changing, so I like this proposal

deployment/network-operator/README.md
e0ne: will add it asap

deployment/network-operator/templates/_helpers.tpl
```diff
@@ -65,3 +65,117 @@ Selector labels
 app.kubernetes.io/name: {{ include "network-operator.name" . }}
 app.kubernetes.io/instance: {{ .Release.Name }}
 {{- end }}
+
+{{/*
+imagePullSecrets helpers
+*/}}
+{{- define "network-operator.imagePullSecrets" }}
```
adrianchiris, line 72: nit: network-operator.operator.imagePullSecrets

adrianchiris (approved):
a couple of nits otherwise LGTM

deployment/network-operator/templates/_helpers.tpl
e0ne: done

adrianchiris (approved):

PR #360, opened 2022-06-01
Add image push CD for pushing images ubon code merge and tag creation

AbdYsn:
Added a Jenkinsfile to build and push the projects images to NVIDIA
internal container registry.

### DIFF
--- a/ci/image-push.Jenkinsfile
+++ b/ci/image-push.Jenkinsfile
@@ -1,0 +1,42 @@
+#!/usr/bin/env groovy
+
+library identifier: 'cloudo-shared-lib@master', retriever: modernSCM(
+  [$class: 'GitSCMSource',
+   remote: 'https://gitlab-master.nvidia.com/cloud-orchestration/cicd/jenkins-confs.git',
+   credentialsId: '184e759a-c526-4e1c-b41c-ed7d409efdbb'])
+
+def ref = "${env.gitlabBranch}"
+def versionPrefix = 'refs/tags/v'
+def tag = ''
+
+if (ref.startsWith(versionPrefix)) {
+    tag = ref.substring(versionPrefix.size()-1)
+} else if (ref == 'master') {
+    tag = 'latest'
+}
+
+if (tag == '') {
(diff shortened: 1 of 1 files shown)


AbdYsn:
@adrianchiris @e0ne , PTAL,
the ci will use the internal gitlab mirror of the project to trigger `push` and `tag` events, and it will be using the jenkinsfile added by this PR.
thanks.

ci/image-push.Jenkinsfile
```diff
@@ -0,0 +1,42 @@
...
+  [$class: 'GitSCMSource',
+   remote: 'https://gitlab-master.nvidia.com/cloud-orchestration/cicd/jenkins-confs.git',
+   credentialsId: '184e759a-c526-4e1c-b41c-ed7d409efdbb'])
+
+def ref = "${env.gitlabBranch}"
+def versionPrefix = 'refs/tags/v'
+def tag = ''
+
+if (ref.startsWith(versionPrefix)) {
+    tag = ref.substring(versionPrefix.size())
```
adrianchiris, line 13: we want a tag like : network-operator:1.0.0 or network-operator:v1.0.0 ? i think we want the latter.

this will extract version without the "v" prefix

adrianchiris: @AbdYsn ping :)

AbdYsn: Sorry got caught up in other stuff, and did not have the time to fix the issue, i agree with you, my only question here would be if you want to keep the logic of only pushing images for tags staring with v?

AbdYsn: done

adrianchiris (approved):

ci/image-push.Jenkinsfile
adrianchiris: >Sorry got caught up in other stuff, and did not have the time to fix the issue, i agree with you, my only question here would be if you want to keep the logic of only pushing images for tags staring with v?

i think its ok for now.

PR #364, opened 2022-06-07
Add support for env variables for OFED container

almaslennikov:
This commit extends NicClusterPolicy to allow to configure environment variables for OFED container.
It also streamlines the process of setting env variables during helm deployment.


deployment/network-operator/README.md
```diff
@@ -334,6 +334,7 @@ Production cluster environment can deny direct access to the Internet and instea
 | `ofedDriver.image` | string | `mofed` | Mellanox OFED driver image name |
 | `ofedDriver.version` | string | `5.5-1.0.3.2` | Mellanox OFED driver version |
 | `ofedDriver.imagePullSecrets` | list | `[]` | An optional list of references to secrets to use for pulling any of the Mellanox OFED driver image |
+| `ofedDriver.env` | list | `[]` | An optional list of environment variables passed the Mellanox OFED driver image                    |
```
rollandf, line 337: `An optional list of environment variables passed to the Mellanox OFED driver container` instead of
`An optional list of environment variables passed the Mellanox OFED driver image`

api/v1alpha1/nicclusterpolicy_types.go
```diff
@@ -65,6 +65,8 @@ type OFEDDriverSpec struct {
 	LivenessProbe *PodProbeSpec `json:"livenessProbe,omitempty"`
 	// Pod readiness probe settings
 	ReadinessProbe *PodProbeSpec `json:"readinessProbe,omitempty"`
+	// List of environment variables to set in the OFED container.
```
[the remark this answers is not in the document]
rollandf: @e0ne If we have a separate CRD, won't we have a backport compatibility issue?

deployment/network-operator/README.md
adrianchiris: please add a link to : https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.22/#envvar-v1-core

or similar so the format is clear

deployment/network-operator/templates/mellanox.com_v1alpha1_nicclusterpolicy_cr.yaml
```diff
@@ -24,6 +24,13 @@ spec:
     image: {{ .Values.ofedDriver.image }}
     repository: {{ .Values.ofedDriver.repository }}
     version: {{ .Values.ofedDriver.version }}
+    {{- if .Values.ofedDriver.env }}
```
adrianchiris, line 27: in general, v1.env you use supports also valueFrom key which can be used to reference values from object or from configmap/secret

so, to allow providing those you could just put the entire env section as is:
```
{{- if .Values.ofedDriver.env }}
  env: {{ toYaml .Values.ofedDriver.env | nindent 4 }}
{{- end }}
```

note: nindent number may be wrong :)

manifests/stage-ofed-driver/0050_ofed-driver-ds.yaml
```diff
@@ -66,6 +66,12 @@ spec:
               value: {{ .RuntimeSpec.HTTPSProxy }}
             - name: NO_PROXY
               value: {{ .RuntimeSpec.NoProxy }}
+          {{- if .CrSpec.Env }}
```
adrianchiris, line 69: v1.env you use supports also valueFrom key which can be used to reference values from object or from configmap/secret

to support those as well you can just append to the list each entry in `.CrSpec.Env`

adrianchiris (commented):
first use-case is to enable NFSoRDMA for GPU Direct storage use-case.
once we have that up and running please add an example for it as well.

adrianchiris:
nit: in general i would split to 2 or 3 commits 

1. changes to CRDs / manifests 
2. Changes to helm chart + helm chart documentation
3. doc/examples updates

PR is small so not really mandatory.

adrianchiris (commented):
@almaslennikov  LMK when this can be merged

e0ne:
CI failures are related to these changes

e0ne (changes requested):

manifests/stage-ofed-driver/0050_ofed-driver-ds.yaml
```diff
@@ -68,6 +68,9 @@ spec:
...
+            {{ toYaml .Values.ofedDriver.env | nindent 12 }}
```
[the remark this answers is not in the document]
almaslennikov: Yep, thanks, changed it to a usage of a local function

almaslennikov:
@adrianchiris @e0ne PTAL

e0ne (approved):
LGTM

adrianchiris (approved):

PR #366, opened 2022-06-15
Release Network Operator v1.3.0-beta.1

e0ne:

### DIFF
--- a/deployment/network-operator/values.yaml
+++ b/deployment/network-operator/values.yaml
@@ -128,7 +128,7 @@ operator:
                 values: [ "" ]
   repository: nvcr.io/nvidia/cloud-native
   image: network-operator
-  imagePullSecrets: []
+  # imagePullSecrets: []
   nameOverride: ""
   fullnameOverride: ""
   # tag, if defined will use the given image tag, else Chart.AppVersion will be used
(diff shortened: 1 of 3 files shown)


deployment/network-operator/values.yaml
```diff
@@ -126,15 +126,15 @@ operator:
               - key: "node-role.kubernetes.io/control-plane"
                 operator: In
                 values: [ "" ]
-  repository: nvcr.io/nvidia/cloud-native
+  repository: nvcr.io/nvidia/mellanox
```
adrianchiris, line 129: revert to previous value

deployment/network-operator/values.yaml
```diff
@@ -126,15 +126,15 @@ operator:
...
   image: network-operator
-  imagePullSecrets: []
+  #imagePullSecrets: []
   nameOverride: ""
   fullnameOverride: ""
   # tag, if defined will use the given image tag, else Chart.AppVersion will be used
   # tag
 
-imagePullSecrets: []
+#imagePullSecrets: []
```
adrianchiris, line 137: @e0ne does providing no value in image pull secrets here and providing an empty list equivalent when deploying the chart ?

it should be IMO but i want to make sure

adrianchiris:
once comments addressed im LGTM

PR #369, opened 2022-06-16
Update MOFED driver image format

adrianchiris:
starting with MOFED driver version  5.7
the mofed driver container image name format
will change from:
  `<image-repo>/<image-name>-<driver-ver>:<os-name><os-ver>-<cpu-arch>`
to:
  `<image-repo>/<image-name>:<driver-ver>-<os-name><os-ver>-<cpu-arch>`

- Support both formats depending on the driver version
  in state ofed.
- Add unit test for the newly added logic

### DIFF
--- a/pkg/state/state_ofed.go
+++ b/pkg/state/state_ofed.go
@@ -17,8 +17,10 @@ limitations under the License.
 package state
 
 import (
+	"fmt"
 	"os"
 
+	"github.com/Masterminds/semver/v3"
 	"github.com/pkg/errors"
 	appsv1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
@@ -36,8 +38,23 @@ import (
 	"github.com/Mellanox/network-operator/pkg/utils"
 )
 
-const stateOFEDName = "state-OFED"
-const stateOFEDDescription = "OFED driver deployed in the cluster"
+const (
+	stateOFEDName        = "state-OFED"
+	stateOFEDDescription = "OFED driver deployed in the cluster"
+
+	// newMofedImageFormatVersion is the first mofed driver container version
+	// which has a new image name format
+	// TODO: update with new version
+	newMofedImageFormatVersion = "5.7-0.1.2.0"
+	// mofedImageNewFormat is the new mofed driver container image name format
+	// format: <repo>/<image-name>:<driver-version>-<os-name><os-ver>-<cpu-arch>
+	// e.x: nvcr.io/nvidia/mellanox/mofed:5.7-0.1.2.0-ubuntu20.04-amd64
+	mofedImageNewFormat = "%s/%s:%s-%s%s-%s"
+	// mofedImageOldFormat is the old mofed driver container image name format
+	// format: <repo>/<image-name>-<driver-version>:<os-name><os-ver>-<cpu-arch>
+	// e.x: nvcr.io/nvidia/mellanox/mofed-5.6-1.0.3.3:ubuntu20.04-amd64
+	mofedImageOldFormat = "%s/%s-%s:%s%s-%s"
+)
 
 // NewStateOFED creates a new OFED driver state
 func NewStateOFED(k8sAPIClient client.Client, scheme *runtime.Scheme, manifestDir string) (State, error) {
@@ -63,12 +80,13 @@ type stateOFED struct {
 
 type ofedRuntimeSpec struct {
 	runtimeSpec
-	CPUArch    string
-	OSName     string
-	OSVer      string
-	HTTPProxy  string
-	HTTPSProxy string
-	NoProxy    string
+	CPUArch        string
+	OSName         string
+	OSVer          string
+	MOFEDImageName string
+	HTTPProxy      string
+	HTTPSProxy     string
+	NoProxy        string
 }
 
 type ofedManifestRenderData struct {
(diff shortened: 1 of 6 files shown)


pkg/state/state_ofed.go
```diff
@@ -36,8 +37,13 @@ import (
...
 )
 
-const stateOFEDName = "state-OFED"
-const stateOFEDDescription = "OFED driver deployed in the cluster"
+const (
+	stateOFEDName        = "state-OFED"
+	stateOFEDDescription = "OFED driver deployed in the cluster"
+
+	// TODO: update with new version
+	newMofedImageFormatVersion = "5.7-0.0.8.1"
```
ykulazhenkov, line 45: What about version with two digits, e.g. `5.10-0.0.8.1`?

pkg/state/state_ofed.go
```diff
@@ -195,3 +204,32 @@ func (s *stateOFED) getManifestObjects(
 	log.V(consts.LogLevelDebug).Info("Rendered", "objects:", objs)
 	return objs, nil
 }
+
+// getMofedDriverImageName generates MOFED driver image name based on the driver version specified in CR
+// TODO(adrianc): in Network-Operator v1.5.0, we should just use the new naming scheme
+func (s *stateOFED) getMofedDriverImageName(cr *mellanoxv1alpha1.NicClusterPolicy,
```
ykulazhenkov, line 210: nit: reduce typing
```
// getMofedDriverImageName generates MOFED driver image name based on the driver version specified in CR
// TODO(adrianc): in Network-Operator v1.5.0, we should just use the new naming scheme
func (s *stateOFED) getMofedDriverImageName(cr *mellanoxv1alpha1.NicClusterPolicy,
	nodeAttr map[nodeinfo.AttributeType]string) string {
	// format: <repo>/<image-name>:<driver-version>-<os-name><os-ver>-<cpu-arch>
	// e.x: nvcr.io/nvidia/mellanox/mofed:5.7-0.0.8.0-ubuntu20.04-amd64.
	mofedImageNameTmpl := "%s/%s:%s-%s%s-%s"
	if cr.Spec.OFEDDriver.Version < newMofedImageFormatVersion  {
		// format: <repo>/<image-name>-<driver-version>:<os-name><os-ver>-<cpu-arch>
		// e.x: nvcr.io/nvidia/mellanox/mofed-5.6-1.0.3.3:ubuntu20.04-amd64
		mofedImageNameTmpl = "%s/%s-%s:%s%s-%s"
	}
	return fmt.Sprintf(mofedImageNameTmpl,
		cr.Spec.OFEDDriver.Repository,
		cr.Spec.OFEDDriver.Image,
		cr.Spec.OFEDDriver.Version,
		nodeAttr[nodeinfo.AttrTypeOSName],
		nodeAttr[nodeinfo.AttrTypeOSVer],
		nodeAttr[nodeinfo.AttrTypeCPUArch])
}
```

pkg/state/state_ofed.go
adrianchiris: good catch !  i will use https://github.com/Masterminds/semver

pkg/state/state_ofed.go
adrianchiris: sure, ill simplify

adrianchiris: Done. LMK if thats OK.

this method now uses Masterminds/semver to parse the version so need to handle possible errors.
current flow will fallback to new image format and log in debug.

which will eventually lead to mofed daemonset to fail to pull image most likely like today.

pkg/state/state_ofed.go
adrianchiris: PTAL, also added UT to cover this

pkg/state/state_ofed.go
```diff
@@ -36,8 +38,19 @@ import (
...
 )
 
-const stateOFEDName = "state-OFED"
-const stateOFEDDescription = "OFED driver deployed in the cluster"
+const (
+	stateOFEDName        = "state-OFED"
+	stateOFEDDescription = "OFED driver deployed in the cluster"
+
+	// TODO: update with new version
+	newMofedImageFormatVersion = "5.7-0.0.8.1"
```
ykulazhenkov, line 46: nit: could you please add a short comment which describes what this version means.
something like this: starting from this OFED version we should use new OFED image name format

ykulazhenkov (approved):

pkg/state/state_ofed.go
ykulazhenkov: looks good, thanks

pkg/state/state_ofed.go
adrianchiris: Updated,

i left the TODO as we still need to finalize the new version

pkg/state/state_ofed.go
```diff
@@ -36,8 +38,23 @@ import (
...
-const stateOFEDName = "state-OFED"
-const stateOFEDDescription = "OFED driver deployed in the cluster"
+const (
+	stateOFEDName        = "state-OFED"
+	stateOFEDDescription = "OFED driver deployed in the cluster"
+
+	// newMofedImageFormatVersion is the first mofed driver container version
+	// which has a new image name format
+	// TODO: update with new version
+	newMofedImageFormatVersion = "5.7-0.0.8.1"
```
e0ne, line 48: Please, change version to 5.7-0.1.2.0

adrianchiris: Done.

adrianchiris:
rebased on top of master to pull in golang 1.18 changes

PR #373, opened 2022-06-20
Pass http_proxy and https_proxy environments to containers

e0ne:
There are no strict rules how these environment variables shold
be named: upper-case or lower-case. To support all scenarios
let's pass both of them into containers witch requires internet
access.

### DIFF
--- a/deployment/network-operator/templates/operator.yaml
+++ b/deployment/network-operator/templates/operator.yaml
@@ -74,6 +74,14 @@ spec:
             - name: HTTPS_PROXY
               value: {{ .Values.proxy.httpsProxy }}
             {{- end }}
+            {{- if .Values.proxy.httpProxy | empty | not }}
+            - name: http_proxy
+              value: {{ .Values.proxy.httpProxy }}
+            {{- end }}
+            {{- if .Values.proxy.httpsProxy | empty | not }}
+            - name: https_proxy
+              value: {{ .Values.proxy.httpsProxy }}
+            {{- end }}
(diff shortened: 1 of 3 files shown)


deployment/network-operator/templates/operator.yaml
```diff
@@ -74,6 +74,14 @@ spec:
             - name: HTTPS_PROXY
               value: {{ .Values.proxy.httpsProxy }}
             {{- end }}
+            {{- if .Values.proxy.httpProxy | empty | not }}
```
ykulazhenkov, line 77: Q: why we use `{{- if .Values.proxy.httpProxy | empty | not }}` instead of just `{{- if .Values.proxy.httpProxy }}`?

deployment/network-operator/templates/operator.yaml
```diff
@@ -74,6 +74,14 @@ spec:
...
+            - name: http_proxy
+              value: {{ .Values.proxy.httpProxy }}
+            {{- end }}
+            {{- if .Values.proxy.httpsProxy | empty | not }}
+            - name: https_proxy
+              value: {{ .Values.proxy.httpsProxy }}
+            {{- end }}
             {{- if .Values.proxy.noProxy | empty | not }}
             - name: NO_PROXY
```
ykulazhenkov, line 86: Do we want to add no_proxy in lowercase as well? This is mostly for consistency, I don't know if any tool that strictly require lowercase version exist.

PR #375, opened 2022-07-03
Network Operator deployment support in vanilla K8S air gap environment

e0ne:


api/v1alpha1/nicclusterpolicy_types.go
```diff
@@ -55,6 +55,18 @@ type PodProbeSpec struct {
 	PeriodSeconds       int `json:"periodSeconds"`
 }
 
+// DriverCertConfigSpec defines custom certificates configuration for driver container
+type DriverCertConfigSpec struct {
+	// +kubebuilder:validation:Optional
+	Name string `json:"name,omitempty"`
```
ykulazhenkov, line 61: nit: if fields has "omitempty" json tag it will be optional by default

pkg/state/state_ofed.go
```diff
@@ -61,20 +80,89 @@ type stateOFED struct {
...
+
+// createConfigMapVolumeMounts creates a VolumeMount for each key
+// in the ConfigMap. Use subPath to ensure original contents
+// at destinationDir are not overwritten.
+func (s *stateOFED) createConfigMapVolumeMounts(namespace, configMapName, destinationDir string) (
+	[]v1.VolumeMount, []v1.KeyToPath, error) {
+	// get the ConfigMap
+	cm := &v1.ConfigMap{}
+
+	opts := client.ObjectKey{Namespace: namespace, Name: configMapName}
```
ykulazhenkov, line 126: nit: imo this variable doesn't contains options. opts -> objKey?

ykulazhenkov (commented):
I think overall logic is fine, I added few nit comments.

Q: do we need to support use-case when user updates content of  configMap with repo or CA configuration?

pkg/state/state_ofed.go
```diff
@@ -67,9 +87,68 @@ type ofedRuntimeSpec struct {
...
+	if err != nil {
+		return nil, nil, fmt.Errorf("ERROR: could not get ConfigMap %s from client: %v", configMapName, err)
+	}
+
+	// create one volume mount per file in the ConfigMap and use subPath
+	var filenames = make([]string, 0, len(cm.Data))
+	for filename := range cm.Data {
+		filenames = append(filenames, filename)
+	}
+	// sort so volume mounts are added to spec in deterministic order
```
adrianchiris, line 137: can you extend this comment to explain why this is needed ?

api/v1alpha1/nicclusterpolicy_types.go
adrianchiris: do you see an issue in being explicit here with the kubebuilder annotations ?

pkg/state/state_ofed.go
```diff
@@ -248,9 +262,9 @@ func (s *stateOFED) getManifestObjects(
 	}
 
 	additionalVolMounts := additionalVolumeMounts{}
+	osname := attrs[0].Attributes[nodeinfo.AttrTypeOSName]
 	// set any custom ssl key/certificate configuration provided
```
adrianchiris, line 266: any chance to move logic related to additional volume mounts to a separate method ?

like `getAdditionalVolumeMouts` or `handleAdditionalVolumeMounts` ?

deployment/network-operator/README.md
```diff
@@ -339,6 +339,8 @@ imagePullSecrets:
 | `ofedDriver.version` | string | `5.5-1.0.3.2` | Mellanox OFED driver version                                                                                                                                              |
 | `ofedDriver.imagePullSecrets` | list | `[]` | An optional list of references to secrets to use for pulling any of the Mellanox OFED driver image                                                                        |
 | `ofedDriver.env` | list | `[]` | An optional list of [environment variables](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.22/#envvar-v1-core) passed to the Mellanox OFED driver image |
+| `ofedDriver.repoConfig.configMapName` | string | `` | Private mirror repository configuration configMap name |
```
adrianchiris, line 342: i feel like these two should be named the same : either `name` or `configMapName` WDYT ?

need to align thoroughout the PR (helm, CRD, code)

pkg/state/state_ofed.go
```diff
@@ -295,7 +295,7 @@ func (s *stateOFED) getManifestObjects(
 		if err != nil {
 			return nil, fmt.Errorf("ERROR: failed to create ConfigMap VolumeMounts for custom repo config: %v", err)
 		}
-		volume := s.createConfigMapVolume(cr.Spec.OFEDDriver.CertConfig.Name, itemsToInclude)
+		volume := s.createConfigMapVolume(cr.Spec.OFEDDriver.RepoConfig.ConfigMapName, itemsToInclude)
```
adrianchiris, line 298: this needs to be in previous commit

adrianchiris (commented):
@e0ne should we have under `docs/` a document about disconnected env covering these configurations ? or do you think its better to just have it in the official docs ? as its not regarded as a "common" flow

api/v1alpha1/nicclusterpolicy_types.go
ykulazhenkov: No technical reasons. For the network-operator we use various technics to mark a field as optional. As far as I can see, this style is not used anywhere in the project yet. I prefer to keep things consistent, but I don't have a preference for a specific style.
This is a "nit" comment and can be completely ignored.

deployment/network-operator/README.md
ykulazhenkov: I think this is to align with NVIDIA gpu-operator.

adrianchiris: these two parameters are very similar, and should be named the same for consistency.

we do not have a strict requirement to have the same api as gpu operator. where we can improve / provide a more consistent api , i think we should.

adrianchiris:
> do we need to support use-case when user updates content of configMap with repo or CA configuration

ATM this is not a defined use-case.
however, keeping in mind that these are used only during driver compilation (pre-step to install kernel-headers)  then even if these change and driver is already compiled (or rather after kernel-header install phase) then it will have no effect.

in case driver container/node restarted, it would have new credentials

deployment/network-operator/README.md
rollandf: @e0ne please also change here in the README

pkg/state/state_ofed.go
```diff
@@ -87,9 +110,76 @@ type ofedRuntimeSpec struct {
...
+				SubPath:   filename})
+		itemsToInclude = append(itemsToInclude, v1.KeyToPath{
+			Key:  filename,
+			Path: filename,
+		})
+	}
+	return volumeMounts, itemsToInclude, nil
+}
+
+func (s *stateOFED) createConfigMapVolume(configMapName string, itemsToInclude []v1.KeyToPath) v1.Volume {
```
adrianchiris, line 173: question: doesnt it make more sense to move the logic of fetching config map and creating "itemsToInclude" here ? then using it in createConfigMapVolumeMounts

i.e 
you first create the volume and then create the mounts for the volume.

then `createConfigMapVolumeMounts` would be just `createVolumeMounts(volumeName string, destDir string, items []v1.KeyToPath)`

pkg/state/state_ofed.go
```diff
@@ -188,6 +294,34 @@ func (s *stateOFED) getManifestObjects(
 		}
 	}
 
+	additionalVolMounts := additionalVolumeMounts{}
+	osname := attrs[0].Attributes[nodeinfo.AttrTypeOSName]
+	// set any custom ssl key/certificate configuration provided
+	if cr.Spec.OFEDDriver.CertConfig != nil && cr.Spec.OFEDDriver.CertConfig.Name != "" {
+		destinationDir, err := getCertConfigPath(osname)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get destination directory for custom repo config: %v", err)
```
adrianchiris, line 303: nit: this is not custom repo config its ssl cert config

pkg/state/state_ofed.go
```diff
@@ -188,6 +294,34 @@ func (s *stateOFED) getManifestObjects(
...
+		}
+
+		err = s.mountAdditionalVolumesFromConfigMap(&additionalVolMounts, cr.Spec.OFEDDriver.CertConfig.Name, destinationDir)
+		if err != nil {
+			return nil, fmt.Errorf("failed to custom certificates: %v", err)
```
adrianchiris, line 308: > failed to custom certificates 

this is not a sentence.

maybe "failed to generate configuration for custom certificates: %v"

pkg/state/state_ofed.go
```diff
@@ -188,6 +294,34 @@ func (s *stateOFED) getManifestObjects(
...
+	// set any custom repo configuration provided
+	if cr.Spec.OFEDDriver.RepoConfig != nil && cr.Spec.OFEDDriver.RepoConfig.Name != "" {
+		destinationDir, err := getRepoConfigPath(osname)
+		if err != nil {
+			return nil, fmt.Errorf("ERROR: failed to get destination directory for custom repo config: %v", err)
+		}
+
+		err = s.mountAdditionalVolumesFromConfigMap(&additionalVolMounts, cr.Spec.OFEDDriver.RepoConfig.Name, destinationDir)
+		if err != nil {
+			return nil, fmt.Errorf("failed to custom repositories configuration: %v", err)
```
adrianchiris, line 321: same comment as above

adrianchiris (changes requested):

api/v1alpha1/nicclusterpolicy_types.go
```diff
@@ -55,6 +55,18 @@ type PodProbeSpec struct {
 	PeriodSeconds       int `json:"periodSeconds"`
 }
 
+// DriverCertConfigSpec defines custom certificates configuration for driver container
+type DriverCertConfigSpec struct {
```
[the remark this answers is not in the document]
adrianchiris: or name it `ConfigMapNameReference` like in openshift :)

https://github.com/openshift/api/blob/8324d657dee1d594a8a7768e5569fea6a8f887a9/config/v1/types.go#L18

api/v1alpha1/nicclusterpolicy_types.go
```diff
@@ -69,6 +81,10 @@ type OFEDDriverSpec struct {
 	Env []v1.EnvVar `json:"env,omitempty"`
 	// Ofed auto-upgrade settings
 	OfedUpgradePolicy *OfedUpgradePolicySpec `json:"upgradePolicy,omitempty"`
+	// Optional: Custom certificates configuration for driver container
+	CertConfig *DriverCertConfigSpec `json:"certConfig,omitempty"`
```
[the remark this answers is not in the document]
e0ne: done

config/crd/bases/mellanox.com_nicclusterpolicies.yaml
```diff
@@ -405,6 +412,12 @@ spec:
                     - initialDelaySeconds
                     - periodSeconds
                     type: object
+                  repoConfig:
+                    description: 'Optional: Custom repo configuration for OFED container'
```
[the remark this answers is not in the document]
e0ne: done

deployment/network-operator/README.md
e0ne: done

deployment/network-operator/crds/mellanox.com_nicclusterpolicies.yaml
```diff
@@ -405,6 +412,12 @@ spec:
                     - initialDelaySeconds
                     - periodSeconds
                     type: object
+                  repoConfig:
+                    description: 'Optional: Custom repo configuration for OFED container'
```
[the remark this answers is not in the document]
e0ne: done

pkg/state/state_ofed.go
e0ne: it's already wrapped into mountAdditionalVolumesFromConfigMap function witch  should be used by the controller

pkg/state/state_ofed.go
e0ne: fixed

api/v1alpha1/nicclusterpolicy_types.go
```diff
@@ -55,6 +55,18 @@ type PodProbeSpec struct {
...
 }
 
+// ConfigMapNameReference references a config map in a specific namespace.
+// The namespace must be specified at the point of use.
+type ConfigMapNameReference struct {
+	Name string `json:"name,omitempty"`
+}
+
+// DriverRepoConfigSpec defines custom repo configuration for OFED container
+type DriverRepoConfigSpec struct {
```
adrianchiris, line 65: not being used anymore, can you remove ?

e0ne: done

pkg/state/state_ofed.go
```diff
@@ -152,6 +239,22 @@ func (s *stateOFED) GetWatchSources() map[string]*source.Kind {
 	return wr
 }
 
+func (s *stateOFED) mountAdditionalVolumesFromConfigMap(
```
adrianchiris, line 242: nit: this method doesnt actually mount anything but rather generate volumes and volume mounts. also it mainly revolves around additionalVolumeMounts type so perhaps we should have a method in that struct.

suggestion:

define:
`(am *additionalMounts) FromConfigMap(cm *v1.ConfigMap, destDir) error`

which will populate additionalMounts from configmap.

then you just need to rename this method to `handleAdditionalMounts` which will get configmap and call `am.FromConfigMap(....)`


wdyt ?

adrianchiris: also please add docstring to added methods above :) if we go this route.

adrianchiris:
@e0ne PTAL

pkg/state/state_ofed.go
```diff
@@ -78,21 +96,94 @@ type stateOFED struct {
...
+		return path, nil
+	}
+	return "", fmt.Errorf("distribution not supported")
+}
+
+// FromConfigMap generates Volumes and VolumeMounts data for the specified ConfigMap object
+func (a *additionalVolumeMounts) FromConfigMap(configMap *v1.ConfigMap, destDir string) error {
+	volumeMounts, itemsToInclude, err := a.createConfigMapVolumeMounts(configMap, destDir)
+	if err != nil {
+		return fmt.Errorf("ERROR: failed to create ConfigMap VolumeMounts for custom repo config: %v", err)
```
adrianchiris, line 139: i think its clear its an error so no need for ERROR: prefix.

also error should not reference custom repo config but rather something generic like:
"failed to create Volume mounts from configmap. %v, err"

pkg/state/state_ofed.go
```diff
@@ -145,13 +236,33 @@ func (s *stateOFED) Sync(customResource interface{}, infoCatalog InfoCatalog) (S
...
+	namespace := config.FromEnv().State.NetworkOperatorResourceNamespace
+	objKey := client.ObjectKey{Namespace: namespace, Name: configMapName}
+	err := s.client.Get(context.TODO(), objKey, configMap)
+	if err != nil {
+		return fmt.Errorf("ERROR: could not get ConfigMap %s from client: %v", configMapName, err)
+	}
+
+	err = volMounts.FromConfigMap(configMap, destDir)
+	if err != nil {
+		return fmt.Errorf("ERROR: could not create volume mounts for ConfigMap: %s", configMapName)
```
adrianchiris, line 260: nit: Drop ERROR: prefix the object returned here is an error

adrianchiris (approved):

adrianchiris:
LGTM, thanks @e0ne !

just need CI to pass and this can be merged

PR #376, opened 2022-07-04
Fixes for OFED auto-upgrade flow

almaslennikov:


deployment/network-operator/templates/role.yaml
```diff
@@ -256,3 +256,15 @@ rules:
       - patch
       - update
       - watch
+  - apiGroups:
```
e0ne, line 259: these files https://github.com/Mellanox/network-operator/blob/master/config/rbac/role.yaml and https://github.com/Mellanox/network-operator/blob/master/deploy/role.yaml should be updated too

e0ne (changes requested):

deployment/network-operator/templates/role.yaml
almaslennikov: Thank you! Updated them too

pkg/upgrade/drain_manager_test.go
```diff
@@ -53,6 +53,37 @@ var _ = Describe("DrainManager tests", func() {
...
+			TimeoutSecond:  1,
+			DeleteEmptyDir: true,
+		}
+		nodeArray := []*corev1.Node{node1, node2, node3}
+		err := drainManager.ScheduleNodesDrain(ctx, &upgrade.DrainConfiguration{Nodes: nodeArray, Spec: drainSpec})
+		Expect(err).To(Succeed())
+
+		time.Sleep(time.Second)
+
+		err = k8sClient.Get(ctx, types.NamespacedName{Name: node1.Name}, node1)
```
adrianchiris, line 77: can you use a new var for the observed node ? e.g `observedNode` instead of reusing node1 ?

adrianchiris (changes requested):
can you re format commit messages so it has header and body ?

otherwise LGTM

pkg/upgrade/drain_manager_test.go
almaslennikov: Thanks, addressed

ykulazhenkov (approved):

adrianchiris (approved):

e0ne (approved):

adrianchiris (approved):

PR #381, opened 2022-07-06
Skip network-operator pod when draining nodes during OFED upgrade

almaslennikov:
We want to skip network-operator itself during the drain because the upgrade process might hang if the operator is evicted and can't be rescheduled to any other node, e.g. in a single-node cluster.
It's safe to do because the goal of the node draining during the upgrade is to evict pods that might use MOFED and network-operator doesn't use in its own pod.

### DIFF
--- a/deployment/network-operator/templates/operator.yaml
+++ b/deployment/network-operator/templates/operator.yaml
@@ -28,6 +28,7 @@ spec:
   template:
     metadata:
       labels:
+        nvidia.com/ofed-upgrade.skip-drain: "true"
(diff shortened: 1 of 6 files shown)


deployment/network-operator/templates/operator.yaml
```diff
@@ -28,6 +28,7 @@ spec:
   template:
     metadata:
       labels:
+        nvidia.com/ofed-upgrade.skip-drain: true
```
adrianchiris, line 31: deployment will now not match on this pod template.

adrianchiris (approved):

rollandf (approved):

PR #387, opened 2022-07-12
Requeue if pod not available in AutoUpgrade

rollandf:
In the auto upgrade flow, the upgrade state of the nodes is built
based on the Ofed pods.

However, there can be a stage where pods are not available as
the Upgrade flow is deleting them and new ones were not created
yet.

In that case, the corresponding node will not be considered in the
upgrade logic at this specific time when the pod is not re-created.

This can affect the "MaxParallelUpgrades" feature.

With this commit, in case a Ofed DaemonSet has one or more unavailable
pods, the upgrade logic will not run and the notification will be requeue
to be handle later.

### DIFF
--- a/controllers/upgrade_controller.go
+++ b/controllers/upgrade_controller.go
@@ -18,6 +18,7 @@ package controllers
 
 import (
 	"context"
+	"fmt"
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


rollandf:
From DaemonSet [documentation](https://pkg.go.dev/k8s.io/api/apps/v1#DaemonSetStatus):
```
// The number of nodes that should be running the
// daemon pod and have none of the daemon pod running and available
// (ready for at least spec.minReadySeconds)
// +optional
NumberUnavailable int32 `json:"numberUnavailable,omitempty" protobuf:"varint,8,opt,name=numberUnavailable"`
```

e0ne (approved):
LGTM

controllers/upgrade_controller.go
```diff
@@ -152,6 +153,13 @@ func (r *UpgradeReconciler) BuildState(ctx context.Context) (*upgrade.ClusterUpg
 
 	r.Log.V(consts.LogLevelDebug).Info("Got driver daemon sets", "length", len(daemonSets))
 
+	for _, ds := range daemonSets {
```
adrianchiris, line 156: as discussed offline, the DS status may not yet be updated by k8s daemonset controller. so we may be called after pod delete but before ds got a chance to update its status, leading to the same issue.

i think we should rely on ds status DesiredNumberScheduled (which just maps to the number of nodes that should run a ds pod) as this is not expected to change (scale up / down) in the middle of an active update flow.

then we can list the pods and count the ones associated with daemonset (have owner reference) and dont have deletion timestap set.

if the two are the same it means that pods that were deleted got rescheduled by the daemon. so we sort of do a similar logic to k8s daemonset controller.

https://github.com/kubernetes/kubernetes/blob/a1c8e9386af844757333733714fa1757489735b3/pkg/controller/daemon/daemon_controller.go#L755

controllers/upgrade_controller.go
```diff
@@ -162,22 +163,19 @@ func (r *UpgradeReconciler) BuildState(ctx context.Context) (*upgrade.ClusterUpg
...
-		if !ok {
-			r.Log.V(consts.LogLevelWarning).Info("OFED Driver Pod is not owned by an OFED Driver DaemonSet",
-				"pod", pod, "actual owner", pod.OwnerReferences[0])
-			continue
+	filteredPodList := []corev1.Pod{}
+	for _, ds := range daemonSets {
+		dsPods := r.getPodsOwnedbyDs(ds, podList.Items)
+		if int(ds.Status.DesiredNumberScheduled) < len(dsPods) {
+			r.Log.V(consts.LogLevelInfo).Info("Driver daemon set has Unscheduled pods", "name", ds.Name)
+			return nil, fmt.Errorf("DS should not have Unscheduled pods")
```
adrianchiris, line 171: note: this is not really an error but a valid system state we might hit. Ideally we would need to just return Nil and request requeue.

adrianchiris (approved):

PR #393, opened 2022-07-24
Fix condition on MOFED upgrade

rollandf:
In case that the DaemonSet DesiredNumberScheduled is
different than the number of pods, requeue.


adrianchiris (approved):
LGTM ! lets just wait for CI to pass.

PR #397, opened 2022-08-01
Set a default timeout for node drain in OFED auto-upgrade config

almaslennikov:
When attempting to drain a node, a non-fatal error might occur and the drain library will retry the action which caused this error in a few seconds (for example, evicting the pod violates the deployment's disruption budget).
If the error's cause is not fixed, there will be infinite retries unless the drain is terminated by timeout.
Since drain operation is disruptive and makes the node unschedulable, introducing the default timeout to avoid taking the node out of action indefinitely.

Signed-off-by: Alexander Maslennikov<amaslennikov@nvidia.com>

### DIFF
--- a/deployment/network-operator/values.yaml
+++ b/deployment/network-operator/values.yaml
@@ -179,7 +179,8 @@ ofedDriver:
       enable: true
       force: false
       podSelector: ""
-      timeoutSeconds: 0
+      # It's recommended to set a timeout to avoid infinite drain in case non-fatal error keeps happening on retries
+      timeoutSeconds: 300
       deleteEmptyDir: false
 
 nvPeerDriver:
--- a/docs/automatic-ofed-upgrade.md
+++ b/docs/automatic-ofed-upgrade.md
@@ -35,7 +35,8 @@ spec:
         # specify a label selector to filter pods on the node that need to be drained
         podSelector: ""
         # specify the length of time in seconds to wait before giving up drain, zero means infinite
-        timeoutSeconds: 0
+        # if not specified, the default is 300 seconds
+        timeoutSeconds: 300
         # specify if should continue even if there are pods using emptyDir
         deleteEmptyDir: false
 ```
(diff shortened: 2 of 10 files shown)


deployment/network-operator/values.yaml
```diff
@@ -179,7 +179,8 @@ ofedDriver:
       enable: true
       force: false
       podSelector: ""
-      timeoutSeconds: 0
+      # It's recommended to set a timeout to avoid infinite drain in case non-fatal error keeps happening on retries
+      timeoutSeconds: 300
```
ykulazhenkov, line 183: Do we also want to set a default timeout in the CRD itself?

almaslennikov: Good idea, thanks

ykulazhenkov: Let's also check with @adrianchiris and @e0ne. WDYT? Do we need to set such defaults in both: charts and CRD itself?

ykulazhenkov: discussed offline: need to apply to both

ykulazhenkov (approved):

deployment/network-operator/values.yaml
adrianchiris: +1 on setting both

docs/automatic-ofed-upgrade.md
```diff
@@ -35,7 +35,7 @@ spec:
         # specify a label selector to filter pods on the node that need to be drained
         podSelector: ""
         # specify the length of time in seconds to wait before giving up drain, zero means infinite
```
adrianchiris, line 37: nit you should mention in this doc that the default timeout if not specified is 5 minutes

deployment/network-operator/values.yaml
```diff
@@ -179,7 +179,8 @@ ofedDriver:
       enable: true
```
adrianchiris, line 179: should we add upgrade policy chart parameters to `deployment/README` ?

e0ne (changes requested):
Please, update examples too

deployment/network-operator/values.yaml
e0ne: Let's do it in a separate PR

e0ne (approved):

PR #399, opened 2022-08-05
Release Network Operator v1.3.0

e0ne:
depends-on: kubernetes#80

### DIFF
--- a/deployment/network-operator/Chart.yaml
+++ b/deployment/network-operator/Chart.yaml
@@ -1,8 +1,8 @@
 apiVersion: v2
 name: network-operator
-version: 1.3.0-beta.4
+version: 1.3.0
 kubeVersion: '>= 1.17.0'
-appVersion: v1.3.0-beta.4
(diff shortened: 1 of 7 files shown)


example/rdma-test-pod1.yml
```diff
@@ -27,8 +27,8 @@ spec:
         rdma/rdma_shared_device_a: 1
       requests:
         rdma/rdma_shared_device_a: 1
-    command:
-    - sh
+    command:    - sh
+
```
adrianchiris, line 31: nit: no need for this change

adrianchiris (approved):
should we also update : https://github.com/Mellanox/network-operator/blob/master/config/samples/mellanox.com_v1alpha1_nicclusterpolicy.yaml ?

e0ne:
> should we also update : https://github.com/Mellanox/network-operator/blob/master/config/samples/mellanox.com_v1alpha1_nicclusterpolicy.yaml ?

It will be updated in #401

PR #406, opened 2022-09-21
Fix support for k8s 1.25

e0ne:


pkg/state/state_pod_security_policy.go
```diff
@@ -102,6 +104,13 @@ func (s *statePodSecurityPolicy) Sync(customResource interface{}, infoCatalog In
 // Get a map of source kinds that should be watched for the state keyed by the source kind name
 func (s *statePodSecurityPolicy) GetWatchSources() map[string]*source.Kind {
 	wr := make(map[string]*source.Kind)
+	psp := &policyv1beta1.PodSecurityPolicyList{}
+	err := s.client.List(context.TODO(), psp)
```
ykulazhenkov, line 108: Should we check for error type for NotMatch?

"k8s.io/apimachinery/pkg/api/meta"

meta.IsNoMatchError(err)

pkg/state/state_pod_security_policy.go
```diff
@@ -102,6 +104,13 @@ func (s *statePodSecurityPolicy) Sync(customResource interface{}, infoCatalog In
...
+	if err != nil {
+		// We assume it's k8s v1.25 or newer so PodSecurityPolicy is not supported and no need to reconcile them
+		log.V(consts.LogLevelInfo).Info("pod security policy is not available")
```
almaslennikov, line 111: Do we need to add an error message if helm chart is deployed with PSP enabled on k8s 1.25+?

e0ne: IMO, it should be done on CRD validation stage in an admission controller which isn't implemented yet

almaslennikov: Fair point.
I guess to explicitly deprecate PSP in NicClusterPolicy we also need an admission controller?

pkg/state/state_pod_security_policy.go
e0ne: Let's do it! I checked "k8s.io/apimachinery/pkg/api/errors" and it doesn't contain required logic. Thanks for the pointing me on it

almaslennikov (approved):

PR #413, opened 2022-10-07
Enable security context for ipoib-cni in OCP

almaslennikov:
Add service account, role and securityContextConstraints


manifests/stage-ipoib-cni/0050-ipoib-cni-ds.yml
```diff
@@ -47,6 +47,7 @@ spec:
         {{- else }}
           {{- .NodeAffinity | yaml | nindent 10 }}
         {{- end }}
+      serviceAccountName: ipoib-cni
```
e0ne, line 50: It should be:
```
{{if eq .RuntimeSpec.OSName "rhcos"}}
      serviceAccountName: ipoib-cni
{{end}}
```

e0ne (changes requested):

manifests/stage-ipoib-cni/0040_scc.openshift.yaml
```diff
@@ -0,0 +1,49 @@
+{{if eq .RuntimeSpec.OSName "rhcos"}}
+# Please edit the object below. Lines beginning with a '#' will be ignored,
```
ykulazhenkov, line 2: nit: For me it is a bit easier to read specs if yaml keys are ordered with k8s style:
apiVersion:
kind:
metadata:
..other fields

WDYT?

almaslennikov: This appears to be a common format for OCP Security Context Constraints: https://docs.openshift.com/container-platform/4.6/authentication/managing-security-context-constraints.html
We use the same format for all other states.
We can change but for consistency it needs to be done for all similar files in the project

ykulazhenkov: Ok, let's keep it as is

ykulazhenkov (approved):

e0ne (approved):
LGTM. Will merge it after testing

e0ne (approved):

PR #417, opened 2022-10-19
Disable ofed driver deployment by default in OCP bundle

almaslennikov:
Remove ofed driver section from the default NicClusterPolicy for OCP bundle.
OCP bundle manifests are built from the config/ folder using kubebuilder. So, the changes are also reflected there.


shivamerla:
@almaslennikov any reason why MOFED deployment is removed for OCP?

almaslennikov:
@shivamerla The default behavior for network-operator is to not deploy the OFED driver unless the user intends to. This PR brings OCP bundle in accordance with helm chart

PR #419, opened 2022-10-22
Watch for node label change event in NicClusterPolicy reconcile loop

e0ne:
We need to watch for node create event for correct cluster scale up by adding 'network.nvidia.com/operator.mofed.wait' label.

### DIFF
--- a/controllers/nicclusterpolicy_controller.go
+++ b/controllers/nicclusterpolicy_controller.go
@@ -28,8 +28,11 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/util/workqueue"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
@@ -249,20 +252,33 @@ func (r *NicClusterPolicyReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	}
 	r.stateManager = stateManager
 
-	builder := ctrl.NewControllerManagedBy(mgr).
+	ctl := ctrl.NewControllerManagedBy(mgr).
 		For(&mellanoxv1alpha1.NicClusterPolicy{}).
 		// Watch for changes to primary resource NicClusterPolicy
 		Watches(&source.Kind{Type: &mellanoxv1alpha1.NicClusterPolicy{}}, &handler.EnqueueRequestForObject{})
 
-	// Watch for changes to secondary resource DaemonSet and requeue the owner NicClusterPolicy
+	// we always add object with a same(static) key to the queue to reduce
+	// reconciliation count
+	updateEnqueue := handler.Funcs{
+		UpdateFunc: func(e event.UpdateEvent, q workqueue.RateLimitingInterface) {
+			q.Add(reconcile.Request{NamespacedName: types.NamespacedName{
+				Name: consts.NicClusterPolicyResourceName,
+			}})
+		},
+	}
+
+	// Watch for "feature.node.kubernetes.io/pci-15b3.present" label applying
+	nodePredicates := builder.WithPredicates(MlnxLabelChangedPredicate{})
+	ctl = ctl.Watches(&source.Kind{Type: &corev1.Node{}}, updateEnqueue, nodePredicates)
+
 	ws := stateManager.GetWatchSources()
 	r.Log.V(consts.LogLevelInfo).Info("Watch Sources", "Kind:", ws)
 	for i := range ws {
-		builder = builder.Watches(ws[i], &handler.EnqueueRequestForOwner{
+		ctl = ctl.Watches(ws[i], &handler.EnqueueRequestForOwner{
 			IsController: true,
 			OwnerType:    &mellanoxv1alpha1.NicClusterPolicy{},
 		})
 	}
 
-	return builder.Complete(r)
+	return ctl.Complete(r)
 }
--- a/controllers/predicate.go
+++ b/controllers/predicate.go
@@ -1,0 +1,38 @@
+/*
+Copyright 2022 NVIDIA CORPORATION & AFFILIATES
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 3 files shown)


controllers/nicclusterpolicy_controller.go
```diff
@@ -249,20 +272,35 @@ func (r *NicClusterPolicyReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	// reconciliation count
+	qHandler := func(q workqueue.RateLimitingInterface) {
+		q.Add(reconcile.Request{NamespacedName: types.NamespacedName{
+			Namespace: "nic-cluster-policy-reconcile-namespace",
+			Name:      "nic-cluster-policy-reconcile-name",
+		}})
+	}
+
+	createEnqueue := handler.Funcs{
+		CreateFunc: func(e event.CreateEvent, q workqueue.RateLimitingInterface) {
```
e0ne, line 290: This won't work. We need to add label after NFD applied feature.node.kubernetes.io/pci-15b3.present label

controllers/nicclusterpolicy_controller.go
```diff
@@ -249,20 +272,35 @@ func (r *NicClusterPolicyReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+
 	ws := stateManager.GetWatchSources()
 	r.Log.V(consts.LogLevelInfo).Info("Watch Sources", "Kind:", ws)
 	for i := range ws {
-		builder = builder.Watches(ws[i], &handler.EnqueueRequestForOwner{
+		controller = controller.Watches(ws[i], &handler.EnqueueRequestForOwner{
 			IsController: true,
 			OwnerType:    &mellanoxv1alpha1.NicClusterPolicy{},
-		})
+		}).Watches(&source.Kind{Type: &corev1.Node{}}, createEnqueue, nodePredicates)
```
adrianchiris, line 302: i think you would want to add this (watching on nodes) only once ?

controllers/nicclusterpolicy_controller.go
adrianchiris: so it means you need to watch on just updates or create & update IMO

controllers/nicclusterpolicy_controller.go
```diff
@@ -249,20 +253,39 @@ func (r *NicClusterPolicyReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	}
 	r.stateManager = stateManager
 
-	builder := ctrl.NewControllerManagedBy(mgr).
+	ctl := ctrl.NewControllerManagedBy(mgr).
 		For(&mellanoxv1alpha1.NicClusterPolicy{}).
+		WithOptions(controller.Options{MaxConcurrentReconciles: 1}).
```
adrianchiris, line 258: is it related to the issue at hand ?
what is the default if not provided ?

in any case, im ok with limiting reconciles to 1

controllers/nicclusterpolicy_controller.go
```diff
@@ -249,20 +253,39 @@ func (r *NicClusterPolicyReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
 		// Watch for changes to primary resource NicClusterPolicy
 		Watches(&source.Kind{Type: &mellanoxv1alpha1.NicClusterPolicy{}}, &handler.EnqueueRequestForObject{})
 
-	// Watch for changes to secondary resource DaemonSet and requeue the owner NicClusterPolicy
+	// we always add object with a same(static) key to the queue to reduce
+	// reconciliation count
+	qHandler := func(q workqueue.RateLimitingInterface) {
+		q.Add(reconcile.Request{NamespacedName: types.NamespacedName{
```
adrianchiris, line 265: wont reconcile loop fail in this case ? as we are trying to get a non existent resource ?

nic cluster policy is cluster scope (no namespace)

adrianchiris: also you are calling qhandler exactly once, so you can just put the logic in UpdateFunc at L#273 below directly

controllers/nicclusterpolicy_controller.go
```diff
@@ -249,20 +253,39 @@ func (r *NicClusterPolicyReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+			Namespace: consts.DefaultNamespace,
+			Name:      consts.NicClusterPolicyResourceName,
+		}})
+	}
+
+	updateEnqueue := handler.Funcs{
+		UpdateFunc: func(e event.UpdateEvent, q workqueue.RateLimitingInterface) {
```
adrianchiris, line 272: are you planning to also handle delete ? (post 1.4 ?) should we have an issue on it ?

controllers/nicclusterpolicy_controller.go
ykulazhenkov: default is one, I think it is safe to drop this option

controllers/nicclusterpolicy_controller.go
e0ne: It will be handled in a separate PR

controllers/nicclusterpolicy_controller.go
e0ne: > also you are calling qhandler exactly once, so you can just put the logic in UpdateFunc at L#273 below directly

IMO, it's more readable but I can change if needed

controllers/nicclusterpolicy_controller.go
adrianchiris: > default is one, I think it is safe to drop this option

thats what i remembered :)

controllers/nicclusterpolicy_controller.go
e0ne: it will be handled in a separate PR

controllers/nicclusterpolicy_controller.go
```diff
@@ -249,20 +252,34 @@ func (r *NicClusterPolicyReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
 		// Watch for changes to primary resource NicClusterPolicy
 		Watches(&source.Kind{Type: &mellanoxv1alpha1.NicClusterPolicy{}}, &handler.EnqueueRequestForObject{})
 
-	// Watch for changes to secondary resource DaemonSet and requeue the owner NicClusterPolicy
+	// we always add object with a same(static) key to the queue to reduce
+	// reconciliation count
+	updateEnqueue := handler.Funcs{
+		UpdateFunc: func(e event.UpdateEvent, q workqueue.RateLimitingInterface) {
+			q.Add(reconcile.Request{NamespacedName: types.NamespacedName{
+				Namespace: consts.DefaultNamespace,
```
adrianchiris, line 265: this is confusing to me. 
as its not namespace AND we rely on the name in the reconcile loop. can we have Namespace as just an empty string ?

thats what being added to the queue on updates to nicclusterpolicy object right ? (empty namespace)
why do you want to do different here ?

ykulazhenkov (approved):

controllers/nicclusterpolicy_controller.go
```diff
@@ -249,20 +252,34 @@ func (r *NicClusterPolicyReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
 		// Watch for changes to primary resource NicClusterPolicy
 		Watches(&source.Kind{Type: &mellanoxv1alpha1.NicClusterPolicy{}}, &handler.EnqueueRequestForObject{})
 
-	// Watch for changes to secondary resource DaemonSet and requeue the owner NicClusterPolicy
+	// we always add object with a same(static) key to the queue to reduce
+	// reconciliation count
+	updateEnqueue := handler.Funcs{
+		UpdateFunc: func(e event.UpdateEvent, q workqueue.RateLimitingInterface) {
+			q.Add(reconcile.Request{NamespacedName: types.NamespacedName{
+				Namespace: "",
```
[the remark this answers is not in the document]
e0ne: thanks!

PR #430, opened 2022-11-15
update psp permissions in clusterrole

tariq1890:
The kube client used in the Network Operators seems to perfom Watch operations on the resources. If the `watch` permission isn't provided, the pod logs get inundated with the following errors:

```
reflector.go:138] pkg/mod/k8s.io/client-go@v0.20.2/tools/cache/reflector.go:167: Failed to watch *v1beta1.PodSecurityPolicy: unknown (get podsecuritypolicies.policy)
```

### DIFF
--- a/deployment/network-operator/templates/role.yaml
+++ b/deployment/network-operator/templates/role.yaml
@@ -227,6 +227,7 @@ rules:
       - get
       - update
       - list
+      - watch
   - apiGroups:
       - policy
     resources:


tariq1890:
@e0ne Can you take a look at this PR? Without this fix, the logs of the network-operator container are extremely hard to read

adrianchiris:
@e0ne we should really sync these templates from kubebuilder autogenerated code as i see we have a watch annot added to controller[1]

LGTM

[1] https://github.com/Mellanox/network-operator/blob/762d43f8455b10c9e68123e4dbfeeef062281e73/controllers/nicclusterpolicy_controller.go#L61

adrianchiris (approved):

PR #440, opened 2022-12-19
add support for sriov-cni deployments

tariq1890:
[SR-IOV CNI](https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin/blob/master/docs/vf-setup.md) is an important component for setting up RDMA in Kubernetes node. This should de a deployment artifact managed by Network operator


adrianchiris:
Hi,

SR-IOV related components are deployed by sriov-network-operator[1] which network operator deploys via helm chart[2].

[1] https://github.com/k8snetworkplumbingwg/sriov-network-operator
[2]https://github.com/Mellanox/network-operator/blob/d3400d73444db4d4fcac0e98cde6670817bd37e1/deployment/network-operator/values.yaml#L26

tariq1890:
Hi @adrianchiris the issue with sriov network operator is that it configures the VFs. The VF configuration takes too long, so we have chosen to just configure the VFs as part of cloud init.

For this reason, we don't use the sriov network operator. But we still need the sriov cni

tariq1890:
Closing this as we should be using `sriov-network-operator` instead

PR #443, opened 2023-01-03
Add Mofed termination grace period param

rollandf:
Adding to NicClusterPolicy CRD the possibility to specify the termination grace period for the Mofed container. The default is 300 seconds.

It is also configurable in Helm values file.

### DIFF
--- a/api/v1alpha1/nicclusterpolicy_types.go
+++ b/api/v1alpha1/nicclusterpolicy_types.go
@@ -79,6 +79,12 @@ type OFEDDriverSpec struct {
 	CertConfig *ConfigMapNameReference `json:"certConfig,omitempty"`
 	// Optional: Custom package repository configuration for OFED container
 	RepoConfig *ConfigMapNameReference `json:"repoConfig,omitempty"`
+	// TerminationGracePeriodSeconds specifies the length of time in seconds
+	// to wait before killing the OFED pod on termination
+	// +optional
+	// +kubebuilder:default:=300
+	// +kubebuilder:validation:Minimum:=0
+	TerminationGracePeriodSeconds int64 `json:"terminationGracePeriodSeconds,omitempty"`
 }
 
 // NVPeerDriverSpec describes configuration options for NV Peer Memory driver
--- a/config/crd/bases/mellanox.com_macvlannetworks.yaml
+++ b/config/crd/bases/mellanox.com_macvlannetworks.yaml
@@ -1,10 +1,9 @@
-
 ---
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.4.1
+    controller-gen.kubebuilder.io/version: v0.10.0
   creationTimestamp: null
   name: macvlannetworks.mellanox.com
 spec:
@@ -92,9 +91,3 @@ spec:
     storage: true
     subresources:
       status: {}
-status:
-  acceptedNames:
-    kind: ""
-    plural: ""
-  conditions: []
-  storedVersions: []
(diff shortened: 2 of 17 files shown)


config/samples/mellanox.com_v1alpha1_nicclusterpolicy.yaml
```diff
@@ -20,6 +20,7 @@ spec:
     image: mofed
     repository: mellanox
     version: 5.5-1.0.3.2
+    terminationGracePeriodSeconds: 300
```
e0ne, line 23: Please, add the same for alm-examples in OCP bundle

rollandf: Done

e0ne (approved):
LTGM once CI passed

api/v1alpha1/nicclusterpolicy_types.go
```diff
@@ -79,6 +79,12 @@ type OFEDDriverSpec struct {
 	CertConfig *ConfigMapNameReference `json:"certConfig,omitempty"`
 	// Optional: Custom package repository configuration for OFED container
 	RepoConfig *ConfigMapNameReference `json:"repoConfig,omitempty"`
+	// TerminationGracePeriodSeconds specifies the length of time in seconds
+	// to wait before killing the OFED pod on termination
+	// +optional
+	// +kubebuilder:default:=300
+	// +kubebuilder:validation:Minimum:=0
+	TerminationGracePeriodSeconds int `json:"terminationGracePeriodSeconds,omitempty"`
```
adrianchiris, line 87: https://github.com/kubernetes/api/blob/92fe7a3ca2f4bd72b8f6faeab90fc75f4d1aab27/core/v1/types.go#L3173

lets make it the same as in k8s (i.e int64) WDYT ?

also do you see a benefit of having it a pointer vs non-pointer ?
for me, if we have default i think it fine to go with non-pointer type.

config/crd/bases/mellanox.com_macvlannetworks.yaml
```diff
@@ -1,10 +1,9 @@
-
 ---
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.4.1
+    controller-gen.kubebuilder.io/version: v0.10.0
```
adrianchiris, line 6: we really need to have some makefile targets to get a pre-defined version of controller-gen/operator-sdk so all generate with same version

api/v1alpha1/nicclusterpolicy_types.go
e0ne: > lets make it the same as in k8s (i.e int64) WDYT ?

Good catch, I agree

config/crd/bases/mellanox.com_macvlannetworks.yaml
e0ne: Here is a bug filed: https://github.com/Mellanox/network-operator/issues/444

api/v1alpha1/nicclusterpolicy_types.go
rollandf: updated to int64

config/crd/bases/mellanox.com_macvlannetworks.yaml
rollandf: The version is defined here: https://github.com/Mellanox/network-operator/blob/master/Makefile#L248

The problem is that if you have an older version, it will not be updated unless you delete the older version from `./bin` directory.

adrianchiris: ack so it just making sure CRD are updated using up2date version. in that case i dont think we need the bug.

adrianchiris (approved):

PR #448, opened 2023-01-17
Upgrade k8s lib to v0.25.5

rollandf:
Fix CVE-2020-26160


adrianchiris (approved):

e0ne (approved):
CI issue isn't related to this change but I still would like to have it passed before merge

PR #449, opened 2023-01-23
Add control-plane toleration for ib-kubernetes

almaslennikov:
We want to schedule ib-kubernetes pod onto the control plane, so in order to pass the control-plane taint, we need to add a corresponding toleration to the deployment spec

### DIFF
--- a/manifests/stage-ib-kubernetes/0070-deployment.yaml
+++ b/manifests/stage-ib-kubernetes/0070-deployment.yaml
@@ -71,6 +71,13 @@ spec:
                     values:
                       - ib-kubernetes
               topologyKey: kubernetes.io/hostname
+      tolerations:
+        - key: "node-role.kubernetes.io/master"
+          operator: "Exists"
+          effect: "NoSchedule"
+        - key: "node-role.kubernetes.io/control-plane"
+          operator: "Exists"
+          effect: "NoSchedule"
       containers:
         - name: ib-kubernetes
           image: {{ .CrSpec.Repository }}/{{ .CrSpec.Image }}:{{ .CrSpec.Version }}


manifests/stage-ib-kubernetes/0070-deployment.yaml
```diff
@@ -71,6 +71,15 @@ spec:
                     values:
                       - ib-kubernetes
               topologyKey: kubernetes.io/hostname
+      tolerations:
+        - key: "node-role.kubernetes.io/master"
+          operator: "Equal"
```
adrianchiris, line 76: shouldnt it be operator "Exist" ?

did you test it ? does this form work ?

manifests/stage-ib-kubernetes/0070-deployment.yaml
```diff
@@ -71,6 +71,15 @@ spec:
...
+          value: "true"
+          effect: "NoSchedule"
+        - key: "node-role.kubernetes.io/control-plane"
```
adrianchiris, line 79: same comment as above

manifests/stage-ib-kubernetes/0070-deployment.yaml
adrianchiris: https://github.com/Mellanox/network-operator/blob/baa3be5fb7719f428cb42fb920fe990ab27b729d/manifests/stage-ofed-driver/0050_ofed-driver-ds.yaml#L36

almaslennikov: You are right, thank you. Applied the change

e0ne (approved):

adrianchiris (approved):

PR #457, opened 2023-02-13
Change ib-kubernetes update strategy to Recreate

almaslennikov:
When RollingUpdate strategy is used, the old pod is only deleted when the new pod gets to the running state, to reduce the downtime. https://kubernetes.io/docs/tutorials/kubernetes-basics/update/update-intro/
However, for ib-kubernetes we specified the anti affinity, which prevent two ib-kubernetes pods running on the same node. In this case, when the deployment is updated, the new pod is forever pending because the old one is not deleted.

To fix this, changing the update strategy to Recreate.

### DIFF
--- a/manifests/stage-ib-kubernetes/0070-deployment.yaml
+++ b/manifests/stage-ib-kubernetes/0070-deployment.yaml
@@ -22,7 +22,7 @@ metadata:
 spec:
   progressDeadlineSeconds: 600
   strategy:
-    type: RollingUpdate
+    type: Recreate
   replicas: 1
   selector:
     matchLabels:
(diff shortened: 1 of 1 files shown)


e0ne (approved):
LGTM

almaslennikov:
@adrianchiris We don't want two instances of ib-kubernetes running at the same time. It was designed as a singleton and two instances working in parallel might mess the workflow up

adrianchiris:
> @adrianchiris We don't want two instances of ib-kubernetes running at the same time. It was designed as a singleton and two instances working in parallel might mess the workflow up

ack, so to prevent possible issue with two instances of ib kubernetes running at the same time (during update) the proposal is to change update strategy. (unless leader election is implemented)

i dont think we need the pod anti affinity in general. should we remove it ?

almaslennikov:
@adrianchiris Pod Anti Affinity could serve as additional protection from possible other deployments of ib-kubernetes. If we treat it as an invalid scenario, let's remove anti affinity altogether - Recreate strategy would ensure that two instances won't be run as part of the same deployment.

adrianchiris:
> Pod Anti Affinity could serve as additional protection from possible other deployments of ib-kubernetes.

it would just protect from two deployments running on the same node. which i dont think is the issue.

lets remove pod anti affinity for ib-kubernetes here and in ib-kuberentes example depoyment

adrianchiris (approved):

e0ne (approved):

PR #459, opened 2023-02-16
Use Patch to update node label

adrianchiris:
When MOFED driver is not configured, use Patch
instead of Update as done when updating node labels when MOFED driver is specified.

In addition, enhance errors in case of failure.


e0ne (approved):
LGTM. CI failures aren't related to the change

PR #463, opened 2023-02-22
Release network-operator-23.1.0

e0ne:


release/index.yaml
```diff
@@ -55,7 +81,7 @@ entries:
     version: 1.3.0
   - apiVersion: v2
     appVersion: v1.2.0
-    created: "2022-11-17T23:13:26.262183247+02:00"
+    created: "2023-02-22T16:55:28.534667265+02:00"
```
adrianchiris, line 84: technically we should generate and only merge new entires to avoid such updates to existing entires. but thats not a new issue...

adrianchiris (approved):

PR #465, opened 2023-03-01
Support delete of resources

rollandf:
In case the user deletes a subsection of the
NicClusterPolicy, created resources needs to be deleted.

The state of the stage will be `notReady` until all resources are deleted. Then it will move to `ignore`.

### DIFF
--- a/controllers/nicclusterpolicy_controller.go
+++ b/controllers/nicclusterpolicy_controller.go
@@ -53,6 +53,7 @@ type NicClusterPolicyReconciler struct {
 	stateManager state.Manager
 }
 
+// In case of adding support for additional types, also update in getSupportedGVKs func in pkg/state/state_skel.go
 //nolint
 // +kubebuilder:rbac:groups=mellanox.com,resources=*,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=security.openshift.io,resources=securitycontextconstraints,verbs=get;list;watch;create;update;patch;delete
--- a/controllers/nicclusterpolicy_controller_test.go
+++ b/controllers/nicclusterpolicy_controller_test.go
@@ -1,0 +1,141 @@
+/*
+2023 NVIDIA CORPORATION & AFFILIATES
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controllers //nolint:dupl
+
+import (
+	goctx "context"
+
+	"k8s.io/apimachinery/pkg/types"
+
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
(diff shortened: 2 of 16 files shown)


pkg/state/state_skel.go
```diff
@@ -59,6 +61,71 @@ func (s *stateSkel) Description() string {
 	return s.description
 }
 
+func getSupportedGVKs() []schema.GroupVersionKind {
```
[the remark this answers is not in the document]
rollandf: Sure, good idea.

ykulazhenkov (approved):

controllers/nicclusterpolicy_controller.go
```diff
@@ -53,6 +53,7 @@ type NicClusterPolicyReconciler struct {
 	stateManager state.Manager
 }
 
+// In case of adding support for additional types, also update in getSupportedGVKs func in pkg/state/state_skel.go
```
adrianchiris, line 56: An Idea that popped into mind:

when creating/updating objects, check that an object is in supported GVKs, if it is not then emit a warning in logs that it will not be deleted WDYT ?

we will then have also a visual queue during runtime, WDYT ?

pkg/state/state_skel.go
```diff
@@ -106,6 +173,13 @@ func (s *stateSkel) createOrUpdateObjs(
 			return errors.Wrap(err, "failed to set controller reference for object")
 		}
 
+		labels := desiredObj.GetLabels()
```
adrianchiris, line 176: can you add comment here ?
// add state specific labels 

or move it to separate function : `addStateSpecificLabels(obj)`  ?

pkg/state/state_sriov_dp.go
```diff
@@ -81,7 +81,15 @@ func (s *stateSriovDp) Sync(customResource interface{}, infoCatalog InfoCatalog)
 	if cr.Spec.SriovDevicePlugin == nil {
 		// Either this state was not required to run or an update occurred and we need to remove
 		// the resources that where created.
-		log.V(consts.LogLevelInfo).Info("Device plugin spec in CR is nil, no action required")
+		log.V(consts.LogLevelInfo).Info("SRIOV Device plugin spec in CR is nil, deleting existing objects if needed")
```
adrianchiris, line 84: This whole logic repeats itself across states.

WDYT about putting it in stateSkel ?

then in logs, just use `state.name` ?

controllers/nicclusterpolicy_controller.go
rollandf: Done

pkg/state/state_skel.go
rollandf: done

pkg/state/state_skel.go
```diff
@@ -134,6 +208,37 @@ func (s *stateSkel) createOrUpdateObjs(
 	return nil
 }
 
+func (s *stateSkel) deleteObjects() (bool, error) {
```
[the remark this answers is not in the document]
rollandf: Changed

pkg/state/state_sriov_dp.go
rollandf: Done

controllers/nicclusterpolicy_controller_test.go
```diff
@@ -0,0 +1,141 @@
...
+
+	mellanoxv1alpha1 "github.com/Mellanox/network-operator/api/v1alpha1"
+	"github.com/Mellanox/network-operator/pkg/consts"
+)
+
+//nolint:dupl
+var _ = Describe("NicClusterPolicyReconciler Controller", func() {
+
+	Context("When NicClusterPolicy CR is created", func() {
+		It("should create and delete whereabout", func() {
```
adrianchiris, line 40: nit: can you elaborate a bit more on the It statement 

as this test:
1. creates nicclusterpolcy CR with ipam plugin
2. ensure objects created
3. update nicclusterpolicy CR with nil val for secondary netwokr
4. ensure related  obj deleted

smthn like: should create whereabouts and delete it after un-setting CR value

adrianchiris (approved):
LGTM, 

please see minor comment. can be merged IMO after checking CI issue.

Thanks for addressing my comments !

controllers/nicclusterpolicy_controller_test.go
rollandf: Done

PR #468, opened 2023-03-07
Make OCP 4.12 and 4.13 supported

e0ne:


bundle/metadata/annotations.yaml
```diff
@@ -10,7 +10,7 @@ annotations:
   operators.operatorframework.io.metrics.mediatype.v1: metrics+v1
   operators.operatorframework.io.metrics.project_layout: go.kubebuilder.io/v3
 
-  com.redhat.openshift.versions: v4.10-4.11
+  com.redhat.openshift.versions: v4.10-4.13
```
adrianchiris, line 13: we support 4 versions ? 
do we want to make the tail shorter ?  (i.e start from 4.11)

e0ne: We're aligning with GPU operator [1] https://github.com/NVIDIA/gpu-operator/blob/b588e4e1cbc7fa4e875d30b5c1a3ba046421aff0/bundle/metadata/annotations.yaml#L17 and MOFED images availability. The current assumption is: testing on the latests with assumption backward compatibility works

adrianchiris (approved):

PR #478, opened 2023-04-04
Fix defaults in clusterserviceversion

adrianchiris:


rollandf:
I thought that this file is generated by `make bundle` and that the examples are extracted from https://github.com/Mellanox/network-operator/blob/master/config/samples

adrianchiris:
Fixed and updated commit msg.

there are still some values which needed to be overwritten in clusterserviceversion as they are not derived from config/samples

adrianchiris:
WIP - as need to test on openshift cluster

rollandf (approved):

adrianchiris:
Tested change on Openshift 4.12 it creates nicclusterpolicy by default without sriov network device plugin.

e0ne (approved):

PR #483, opened 2023-04-12
Remove whereabouts IP reconcile CronJob

ykulazhenkov:
IP reconciliation logic is now built in whereabouts,
 and we don't need to deploy CronJob separately

- remove manifest to prevent deployment for new installations
- add migration handler, which will remove CronJob deployed by the previous operator version


rollandf (approved):

main.go
```diff
@@ -120,6 +126,19 @@ func main() {
...
 
+	directClient, err := client.New(clientConf,
+		client.Options{Scheme: mgr.GetScheme(), Mapper: mgr.GetRESTMapper()})
+	if err != nil {
+		setupLog.Error(err, "failed to create direct client")
+		os.Exit(1)
+	}
+
+	// run migration logic before controllers start
+	if err := migrate.Migrate(stopCtx, setupLog.WithName("migrate"), directClient); err != nil {
```
adrianchiris, line 137: Question: do we want some flag to disable migration ? for an un-foreseen case where we would not want to run it ?

adrianchiris:
overall LGTM, one small Q and we can merge this one

main.go
ykulazhenkov: ok, I added "disable-migration" flag to optional turn off migration logic

ykulazhenkov:
PR rebased on master to resolve merge conflict. Added "disable-migrate" flag to optionally disable migration logic

rollandf (approved):

main.go
ykulazhenkov: @e0ne Are you talking about the flag or migration logic in general? 
If about the flag, then think about it in a broader context. For me, this flag is a way to completely disable migration logic. Later, the migration handler can contain much more logic than it has now, and it is better to have a way to disable it.

e0ne: I prefer to keep as few config options as possible. This PR adds some flag with we don't need at the moment and also it requires to re-build image or create a custom helm chart to enable it

ykulazhenkov: Ok, let's remove the flag. The current logic is relatively safe, and there should be no reason to disable it now.

e0ne (approved):

main.go
adrianchiris: My fear that later on as migration logics are added, we will forget to add this flag.
then we might encounter cases where we wish we had it.

ykulazhenkov: After discussion we decided to add env variable which will turn off all migrations

ykulazhenkov:
@e0ne @rollandf  @adrianchiris  I added env variable to disable migrations, PTAL

rollandf (approved):

e0ne (approved):

adrianchiris (approved):
LGTM !

PR #487, opened 2023-04-19
Set sriovNetworkOperator node selector in a helm chart in a safer manner

ykulazhenkov:
If helm value sriovNetworkOperator.configDaemonNodeSelectorExtra map is manually
 removed from the default values.yaml, then
 wrong type for value error may occur while deploying helm chart.


ykulazhenkov:
This error may happen only if the user replaced default helm values (values.yaml) with some old values.yaml which has no this field. This is miss usage from the user, but from our side it is better to set these values in a more safer manner

adrianchiris (approved):

adrianchiris:
@e0ne this one is pretty straight forward can you take a look ?

e0ne (approved):

Issue #482, opened 2023-04-11
Network operator continues to execute sync loop even if NicClusterPolicy.state is ready


ykulazhenkov:
**What happened**:
The network operator runs Reconcile(sync) loop repeatedly after the NicClusterPolicy.state is ready, even if there are no changes in configuration.

The operator updates specs of DaemonSets in each iteration. The content of the DaemonSet object is the same, but the resource version changes.

**What you expected to happen**:

The operator should not run Reconciliation loop if there are no new changes in configuration.

**How to reproduce it (as minimally and precisely as possible)**:

Deploy latest network operator on k8s > 1.25

**Anything else we need to know?**:

```
# NicClusterPolicy is ready

kubectl get nicclusterpolicies.mellanox.com nic-cluster-policy  -o jsonpath="{.status}" | jq

{
  "appliedStates": [
    {
      "name": "state-pod-security-policy",
      "state": "ignore"
    },
    {
      "name": "state-multus-cni",
      "state": "ready"
    },
    {
      "name": "state-container-networking-plugins",
      "state": "ready"
    },
    {
      "name": "state-ipoib-cni",
      "state": "ignore"
    },
    {
      "name": "state-whereabouts-cni",
      "state": "ready"
    },
    {
      "name": "state-OFED",
      "state": "ignore"
    },
    {
      "name": "state-SRIOV-device-plugin",
      "state": "ignore"
    },
    {
      "name": "state-RDMA-device-plugin",
      "state": "ignore"
    },
    {
      "name": "state-NV-Peer",
      "state": "ignore"
    },
    {
      "name": "state-ib-kubernetes",
      "state": "ignore"
    }
  ],
  "state": "ready"
}

# never ending sync

go run main.go 2>&1 | grep "Syncing system state\|Sync Done for custom resource"
1.6812255611425316e+09  INFO    state   Syncing system state
1.6812255708206768e+09  INFO    state   Sync Done for custom resource
1.6812255708334832e+09  INFO    state   Syncing system state
1.6812255806208274e+09  INFO    state   Sync Done for custom resource
1.6812255806328645e+09  INFO    state   Syncing system state
1.6812255904208226e+09  INFO    state   Sync Done for custom resource
1.6812255904326105e+09  INFO    state   Syncing system state
1.6812256002206306e+09  INFO    state   Sync Done for custom resource
1.6812256002327995e+09  INFO    state   Syncing system state


```

**Environment**:
[...]

PR #488, opened 2023-04-19
Revert "server-side apply" patch and apply new fix for never-ending sync loop issue

ykulazhenkov:
This PR reverts https://github.com/Mellanox/network-operator/pull/481

server-side apply may be problematic when we try to upgrade network-operator version

The new version of the operator may not clean fields set by the previous release.

Proposed solution: keep existing Create/Update logic for resource and fix for https://github.com/Mellanox/network-operator/issues/482 in a different way - try to ignore updates of the object if the content (excluding resourceVersion of the object and managedFileds) is the same.

### DIFF
--- a/controllers/predicate.go
+++ b/controllers/predicate.go
@@ -17,6 +17,11 @@ limitations under the License.
 package controllers
 
 import (
+	"reflect"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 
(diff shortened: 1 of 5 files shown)


adrianchiris (approved):

rollandf (approved):

adrianchiris (commented):
@ykulazhenkov  are we OK with merging this one ?

controllers/predicate.go
```diff
@@ -36,3 +41,35 @@ func (p MlnxLabelChangedPredicate) hasMlnxLabel(labels map[string]string) bool {
...
+		return false
+	}
+	oldObj := e.ObjectOld.DeepCopyObject().(client.Object)
+	newObj := e.ObjectNew.DeepCopyObject().(client.Object)
+	// ignore resource version
+	oldObj.SetResourceVersion("")
+	newObj.SetResourceVersion("")
+	// ignore managed fields
+	oldObj.SetManagedFields([]metav1.ManagedFieldsEntry{})
+	newObj.SetManagedFields([]metav1.ManagedFieldsEntry{})
```
adrianchiris, line 65: do we need to ignore status field if it exist ?

Issue #496, opened 2023-05-01
Restore upgrade logic to exclude operator from draining in MOFED upgrade


rollandf:
In commit https://github.com/Mellanox/network-operator/commit/41c79034968be686ed71f180b7f59403f218ab54 a logic to exclude operator from draining in MOFED upgrade using label `ofed-upgrade.skip-drain` was added.

When integrating common upgrade lib in https://github.com/Mellanox/network-operator/pull/446 this logic was removed.
It should be restored in upgrade_controller. (https://github.com/Mellanox/network-operator/blob/master/controllers/upgrade_controller.go#L112)

rollandf:
Fixed in #497

PR #497, opened 2023-05-02
Align the skip-drain label for the operator with the common upgrade lib

almaslennikov:
Fix for https://github.com/Mellanox/network-operator/issues/496

When the upgrade functionality was extracted into a separate upgrade library, the pod label for skipping drain was changed (https://github.com/NVIDIA/k8s-operator-libs/blob/89451d4d7524278f985a7e813d654084c9e5d146/pkg/upgrade/consts.go#L23). The operator's spec was not aligned to use this new label.

This label is required for the operator, when deployed on a single-node cluster, where operator runs on the same node as the ofed driver container. Without this label, the operator pod would drain itself, thus breaking the complete network-operator deployment.

With this label, the pod is omitted during the drain (https://github.com/NVIDIA/k8s-operator-libs/blob/89451d4d7524278f985a7e813d654084c9e5d146/pkg/upgrade/upgrade_state.go#L603)

### DIFF
--- a/bundle/manifests/nvidia-network-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/nvidia-network-operator.clusterserviceversion.yaml
@@ -455,7 +455,7 @@ spec:
                 kubectl.kubernetes.io/default-container: manager
               labels:
                 control-plane: controller-manager
-                nvidia.com/ofed-upgrade.skip-drain: "true"
+                nvidia.com/ofed-driver-upgrade-drain.skip: "true"
             spec:
               containers:
               - args:
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -21,7 +21,7 @@ spec:
     metadata:
       labels:
         control-plane: controller-manager
-        nvidia.com/ofed-upgrade.skip-drain: "true"
+        nvidia.com/ofed-driver-upgrade-drain.skip: "true"
       annotations:
         kubectl.kubernetes.io/default-container: manager
     spec:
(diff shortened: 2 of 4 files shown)


rollandf (approved):

ykulazhenkov (approved):

adrianchiris:
now that we have moved the logic to a common place, technically each user (netop/gpuop) can provide their on drain labels 
i dont think its needed to be handled by the common lib

imo a better solution is:
1. remove the logic from operator-lib (it just handles drain spec as is)
2. add logic to add their own custom labels it to each operator. so they have full control on whats being drained

Thoughts ?

almaslennikov:
> now that we have moved the logic to a common place, technically each user (netop/gpuop) can provide their on drain labels i dont think its needed to be handled by the common lib
> 
> imo a better solution is:
> 
> 1. remove the logic from operator-lib (it just handles drain spec as is)
> 2. add logic to add their own custom labels it to each operator. so they have full control on whats being drained
> 
> Thoughts ?
> 
> that said, its not a blocker for merging this one for now.



> now that we have moved the logic to a common place, technically each user (netop/gpuop) can provide their on drain labels i dont think its needed to be handled by the common lib
> 
> imo a better solution is:
> 
> 1. remove the logic from operator-lib (it just handles drain spec as is)
> 2. add logic to add their own custom labels it to each operator. so they have full control on whats being drained
> 
> Thoughts ?
> 
> that said, its not a blocker for merging this one for now.

That makes sense, created a new issue for that: https://github.com/Mellanox/network-operator/issues/500

adrianchiris (approved):

PR #503, opened 2023-05-03
Refactor: Remove State Groups

adrianchiris:
Its not really needed to manage states in groups
since after reboot all components will start at
the same time.

State manager is simplified and now manages list of states


adrianchiris:
Need to test this one on a setup

rollandf (approved):

e0ne (approved):

Issue #508, opened 2023-05-08
Set mofed.wait label to false when NicClusterPolicy is deleted.


rollandf:
In the case that NicClusterPolicy is deleted, we need to make sure that `mofed.wait` label on nodes is `false`.

This is needed in case other components that relies on this label (like SRIOV Network Operator for example) won't be stuck waiting for it to change.

rollandf:
Fixed by #509

PR #509, opened 2023-05-08
Update mofed.wait label on CR deletion

rollandf:
In the case that NicClusterPolicy is deleted,
we need to make sure that mofed.wait label on nodes is false.

This is needed in case other components that relies on this label (like SRIOV Network Operator for example) won't be stuck waiting for it to change.

Solves issue #508


e0ne (approved):
@rollandf  thanks for adding unit-tests!

almaslennikov (approved):

controllers/nicclusterpolicy_controller.go
```diff
@@ -176,23 +181,29 @@ func (r *NicClusterPolicyReconciler) updateNodeLabels(cr *mellanoxv1alpha1.NicCl
...
+func (r *NicClusterPolicyReconciler) clearMofedWaitLabel() error {
+	// We deploy OFED and Device plugins only on a nodes with Mellanox NICs
+	nodes := &corev1.NodeList{}
+
+	err := r.Client.List(context.TODO(), nodes, client.MatchingLabels{nodeinfo.NodeLabelMlnxNIC: "true"})
+	if err != nil {
+		return errors.Wrap(err, "failed to list nodes")
 	}
 
+	for i := range nodes.Items {
```
adrianchiris, line 199: can we only set it to false if label exist ?

that is, if a deployment doesnt use MOFED container, we will not add the label.

makes sense ?

rollandf: It will change the previous behavior that applied the label with `false` to all nodes with Mellanox NICs if Mofed is not set.
I am not sure what the impact will be.

controllers/nicclusterpolicy_controller.go
```diff
@@ -176,23 +181,29 @@ func (r *NicClusterPolicyReconciler) updateNodeLabels(cr *mellanoxv1alpha1.NicCl
...
-			err := r.Client.Patch(context.TODO(), &nodes.Items[i], client.RawPatch(types.StrategicMergePatchType, patch))
-			if err != nil {
-				return errors.Wrapf(err, "unable to patch %s node label for node %s",
-					nodeinfo.NodeLabelWaitOFED, nodes.Items[i].Name)
-			}
-		}
+	return nil
+}
+
+func (r *NicClusterPolicyReconciler) clearMofedWaitLabel() error {
```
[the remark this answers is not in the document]
rollandf: Added docstring

controllers/nicclusterpolicy_controller.go
adrianchiris: ack, lets keep as is

adrianchiris (approved):

PR #515, opened 2023-05-10
Update examples

adrianchiris:
- Remove nv-peer-mem from examples as it is not recommended to deploy this driver using network-operator
- fix typo in sriov device plugin config


ykulazhenkov (approved):

adrianchiris:
need to update also under `hack/` folder

e0ne (approved):

PR #519, opened 2023-05-12
Fix operator logo in OCP bundle

e0ne:


rollandf (approved):

rollandf:
Seems OK, checked with https://base64.guru/converter/decode/image/png

image (1)

Issue #466, opened 2023-03-06
CI improvements


e0ne:
**What would you like to be added**:
- CI check that all CRDs are up to date
- CI check that all Dependencies are updated

**Why is this needed**:
We'd got this [1] script in the past wich was accidentally removed in this commit [2]. We need bring back this check and add a new one for dependencies to not forget update files in each PR

[1] https://github.com/Mellanox/network-operator/blob/v1.0.0/scripts/check-manifests.sh
[2] https://github.com/Mellanox/network-operator/commit/a9c8812bc65d6a2cf93ad345d183e6fc9b3e17ea

adrianchiris:
in additions to checking manifests have not changed (i.e `make manifest` results in no diffs in repo)
we should also check artifacts from `make release-build` target

PR #525, opened 2023-05-15
Update templates

adrianchiris:


adrianchiris:
@e0ne i think we should add a check in CI to ensure when running `make release-build` there are no diffs

WDYT ?

e0ne:
> @e0ne i think we should add a check in CI to ensure when running `make release-build` there are no diffs
> 
> WDYT ?

We should have this check and the same for `make manifests`

e0ne (approved):

rollandf (approved):

adrianchiris:
updated #466

PR #526, opened 2023-05-16
Add support for CRD upgrade with Helm pre-upgrade hook

ykulazhenkov:
This MR contains two changes:
1. Add CRDs for network-operator and sriov-network-operator to network-operator image
2. Add helm pre-upgrade hook to automatically apply CRDs

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,6 +15,8 @@
 # Build the manager binary
 FROM golang:1.20 as builder
 
+ARG TARGETARCH=amd64
+
 WORKDIR /workspace
 # Copy the Go Modules manifests
 COPY go.mod go.mod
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -29,6 +31,18 @@ COPY api/ api/
 COPY controllers/ controllers/
 COPY pkg/ pkg/
 
+# Add kubectl tool
+RUN curl -LO "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/${TARGETARCH}/kubectl"
+RUN chmod +x ./kubectl
+
+# Add helm chart DIR to get CRDs from it
+COPY deployment/network-operator chart
+# Update helm dependencies and copy CRDs from them
```
adrianchiris, line 40: nit: there is no update of helm dependencies here

Dockerfile
```diff
@@ -29,6 +31,18 @@ COPY api/ api/
...
+RUN mkdir crds && \
+    cd ./chart && \
+    cp -r ./crds /workspace/crds/network-operator/ && \
+    cp -r charts/sriov-network-operator/crds /workspace/crds/sriov-network-operator/
```
adrianchiris, line 44: do we want NFD CRDs ? or we going to have issues with NFD deployed from GPU Op ?

deployment/network-operator/README.md
```diff
@@ -183,22 +183,6 @@ helm search repo mellanox/network-operator -l
 
 > __NOTE__: add `--devel` option if you want to list beta releases as well
 
-### Download CRDs for the specific release
```
adrianchiris, line 186: Im thinking we should still keep a section about upgrading CRDs.

we should have two subsections

1. manual upgrade (what we have here)
2. automatic upgrade using helm chart by setting .Values.upgradeCRDs to true

ykulazhenkov: I changed this section to keep info about manual process

Dockerfile
ykulazhenkov: Correct. I updated the docstring. Update here will be required after we will upgrade the version of the NFD chart

adrianchiris (approved):

rollandf (approved):

ykulazhenkov:
Latest released kubectl has security known security issue which is detected by image_scan job. Need to wait for new kubectl release (the issue is already fixed in master)

PR #534, opened 2023-05-23
Log error in state manager

adrianchiris:
we should log error if it happend regardless
of SyncState.

### DIFF
--- a/hack/templates/values/values.template
+++ b/hack/templates/values/values.template
@@ -22,6 +22,8 @@ nfd:
 psp:
   enabled: false
 
+upgradeCRDs: true
+
 sriovNetworkOperator:
   enabled: false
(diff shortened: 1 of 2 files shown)


deployment/network-operator/values.yaml
```diff
@@ -22,8 +22,6 @@ nfd:
 psp:
   enabled: false
 
-upgradeCRDs: true
```
e0ne, line 25: do we want to roll-back this change? what should be our defaults?

adrianchiris: updated the template instead

adrianchiris:
@e0ne  can we merge this one ?

rollandf (approved):

PR #539, opened 2023-05-23
Ocp v23.4.0

e0ne:


bundle/manifests/nvidia-network-operator.clusterserviceversion.yaml
```diff
@@ -568,11 +608,4 @@ spec:
   provider:
     name: NVIDIA
     url: https://github.com/Mellanox/network-operator/
-  relatedImages:
```
rollandf, line 571: I wonder why these were deleted.

PR #541, opened 2023-05-24
Run CRD upgrade hook on helm chart install

ykulazhenkov:
This required to handle use-case when network-operator upgrade is performed as:
step 1. uninstall an old version
step 2. install a new version

### DIFF
--- a/deployment/network-operator/templates/upgrade-crd.yaml
+++ b/deployment/network-operator/templates/upgrade-crd.yaml
@@ -1,11 +1,54 @@
 {{- if .Values.upgradeCRDs }}
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ include "network-operator.fullname" . }}-hooks-sa
+  annotations:
+    helm.sh/hook: pre-install,pre-upgrade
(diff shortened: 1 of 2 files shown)


deployment/network-operator/templates/upgrade-crd.yaml
```diff
@@ -1,11 +1,54 @@
 {{- if .Values.upgradeCRDs }}
+---
+apiVersion: v1
```
adrianchiris, line 3: we were missing this from original PR ?

ykulazhenkov: No, we were using main service-account of network-operator. The problem is that this account is not exist during pre-install hook

adrianchiris: ack, 

should we suffix them with <full-name>__-hooks__-sa/role/binding ?

ykulazhenkov: This doesn't really matter, because this objects will be immediately removed after completion of the hook. But I renamed them for clarity

adrianchiris (approved):
We would need to cherry-pick to 23.5..x branch

PR #564, opened 2023-06-26
Watch ConfigMap changes for related states

adrianchiris:
Currently config map changes will not trigger reconcile. Add ConfigMap as watched resource for nicclusterpolicy controller using nv-ipam state.


adrianchiris:
@ykulazhenkov @rollandf PTAL, i found that a few other states were missing watch for configmap.

p.s
technically we should watch every resource that we create (and own) but for now lets just align configmap

ykulazhenkov:
LGTM, thx

rollandf (approved):
LGTM

PR #566, opened 2023-07-09
Support new MOFED image format for RHEL

e0ne:


adrianchiris:
please add more info in each commit message.

1. why we drop support for MOFED image format older than 5.7  (because its old and we dont really need it anymore)
2. specify what is the new image format.

pkg/state/state_ofed.go
```diff
@@ -456,12 +460,22 @@ func (s *stateOFED) getMofedDriverImageName(cr *mellanoxv1alpha1.NicClusterPolic
...
 	reqLogger.V(consts.LogLevelDebug).Info("Generating ofed driver image name for version: %v", "version", curDriverVer)
 
+	oldRhelFormatDriverVer, err := semver.NewVersion(oldMofedRhelVersion)
+	if err != nil {
+		reqLogger.V(consts.LogLevelDebug).Info("failed to parse ofed driver version as semver")
+	}
+	osName := nodeAttr[nodeinfo.AttrTypeOSName]
+	osVer := nodeAttr[nodeinfo.AttrTypeOSVer]
+	if curDriverVer != nil && curDriverVer.GreaterThan(oldRhelFormatDriverVer) && osName == rhelOSName {
+		osVer = nodeAttr[nodeinfo.AttrTypeOSVerMajor] + ".x"
```
adrianchiris, line 470: I asked Vlad to tag (re-tag) each version explicitly why do we need this.

in addition IF we will need a special version for a specific RHEL release we will loose the ability to do it.

adrianchiris: @e0ne i think we should try to keep a single format in network operator.

MOFED driver container can be re-tagged as needed for multiple OS versions

adrianchiris (approved):

PR #567, opened 2023-07-09
Enable NodeFeatureApi in NFD

wmousa:
…namespace

This patch does the following:
 - Set enableNodeFeatureApi to true by default
 - Run the NFD with default annotation namespace
 - Enhance READE NFD documentation

### DIFF
--- a/deployment/network-operator/values.yaml
+++ b/deployment/network-operator/values.yaml
@@ -32,7 +32,7 @@ sriovNetworkOperator:
 
 # Node Feature discovery chart related values
 node-feature-discovery:
-  enableNodeFeatureApi: false
+  enableNodeFeatureApi: true
   worker:
     serviceAccount:
       name: node-feature-discovery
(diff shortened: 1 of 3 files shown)


deployment/network-operator/values.yaml
```diff
@@ -32,7 +32,7 @@ sriovNetworkOperator:
 
 # Node Feature discovery chart related values
 node-feature-discovery:
-  enableNodeFeatureApi: false
+  enableNodeFeatureApi: true
```
rollandf, line 35: These are generated. Change in https://github.com/Mellanox/network-operator/blob/master/hack/templates/values/values.template

then run `make release-build` and commit the generated files

e0ne (changes requested):
Please. re-format this commit message to:

```
Enable NodeFeatureApi in NFD

This patch does the following:
 - Set enableNodeFeatureApi to true by default
 - Run the NFD with default annotation namespace
 - Enhance README NFD documentation

Signed-off-by: Waleed Mousa <waleedm@nvidia.com>

```

e0ne (approved):
LGTM. It could be merged once CI passes

PR #571, opened 2023-07-12
Add NodeFeatureRules to label Nvidia NICs

wmousa:
This patch does the following:
 - Add "Nvidia NICs PCI" NFR to lable the node with `pci-15b3.present": "true"` if any of Nvidia NICs is exist
 - Add "Nvidia NICs SRIOV"  NFR to lable the node with `pci-15b3.sriov.capable": "true"` if any of Nvidia NICs has SRIOV enabled
 - Add nfd.nodefeaturerules flag and set it true by default to allow creating NFRs above
 - Remove the networking PCI class from the source configuration


deployment/network-operator/templates/nodefeaturerules.yaml
```diff
@@ -0,0 +1,25 @@
...
+      labels:
+        "pci-15b3.present": "true"
+      matchFeatures:
+        - feature: pci.device
+          matchExpressions:
+            vendor: {op: In, value: ["15b3"]}
+            class: {op: In, value: ["0200", "0207"]}
+    - name: "Nvidia NICs SRIOV"
+      labels:
+        "pci-15b3.sriov.capable": "true"
```
adrianchiris, line 18: i dont think we are using it afaik. (also did a quick search in the code)

@e0ne you know ?

deployment/network-operator/values.yaml
```diff
@@ -18,6 +18,7 @@
 
 nfd:
   enabled: true
+  nodefeaturerules: true
```
adrianchiris, line 21: nit:

deployNodeFeatureRules ?

adrianchiris: need to update documentation under deployment

wmousa: Done

hack/templates/values/values.template
```diff
@@ -52,9 +53,6 @@ node-feature-discovery:
       sources:
         pci:
           deviceClassWhitelist:
-          - "02"
-          - "0200"
-          - "0207"
```
ArangoGutierrez, line 57: I would leave this, in case the users decide to set deployNodeFeatureRules to false. NFD can handle the duplicated config giving NodeFeatureRule priority

deployment/network-operator/templates/nodefeaturerules.yaml
e0ne: We don't use it bevause by default NVIDIA nics could have SRIOV disabled in the FW and this selector won't work. but in general It's good to have it to be less vendor-speciffic if possible

README.md
```diff
@@ -57,16 +57,15 @@ NFD is used to label nodes with the following labels:
...
+>__NOTE__: We use nodeFeatureRules to label PCI vendor and device and it's enabled by default using `nfd.deployNodeFeatureRules` flag
+
 __Example NFD worker configurations:__
 
 ```yaml
     config:
       sources:
         pci:
           deviceClassWhitelist:
-          - "02"
```
e0ne, line 67: why do we need to drop these values?

e0ne (approved):

README.md
adrianchiris: we create the labels via nodefeaturerule CR now, so technically we dont need to have them "ON" by default.

I see @ArangoGutierrez prefers to keep as well.

hack/templates/values/values.template
adrianchiris: I suggested to waleed to remove to avoid having duplicate places that label the same thing.

if this is left here, then why create nodefeaturerule by default. i think we want to rely on one mechanism for most cases.

since we may have external NFD deployed (which we require to have nodefeaturerules enabled) which then will require us to use nodefeaturerules created. i prefer to always rely on it.


if user decides to set this one to false, he needs to be aware of the implications. IMO.
in that case the user either: 
1. has another NFD deployed with all needed configs
2. user would need to change this entry to add the values needed 
3. or simply keep deployNodeFeatureRules true

deployment/network-operator/templates/nodefeaturerules.yaml
adrianchiris: we still label if its nvidia NIC (15b3) so its still vendor specific.

if we dont need it, lets remove this rule

adrianchiris:
we need to add a section in README on the different ways NFD can be deployed.

1. externally
2. via network operator

if externally, we need to specify requirements. i.e version, expected features to be enabled and install example using helm from upstream repo

deployment/network-operator/templates/nodefeaturerules.yaml
wmousa: ACK, removing.

hack/templates/values/values.template
wmousa: ACK, Done

deployment/network-operator/README.md
```diff
@@ -101,9 +101,19 @@ $ kubectl -n network-operator get pods
 By default the network operator
 deploys [Node Feature Discovery (NFD)](https://github.com/kubernetes-sigs/node-feature-discovery)
 in order to perform node labeling in the cluster to allow proper scheduling of Network Operator resources. If the nodes
-where already labeled by other means, it is possible to disable the deployment of NFD by setting
-`nfd.enabled=false` chart parameter.
+where already labeled by other means (either deployed from master or deployed within another deployment), it is possible to disable the deployment of NFD by setting
+`nfd.enabled=false` chart parameter and make sure that the installed version is `v0.13.2` or newer and has NodeFeatureApi enabled.
 
+##### Deploy NFD from master with NodeFeatureApi enabled
```
ykulazhenkov, line 107: nit: Why "from master"? Maybe rename to "from upstream". WDYT?

README.md
```diff
@@ -57,16 +57,15 @@ NFD is used to label nodes with the following labels:
 - RDMA capability
 - GPU features*
 
+>__NOTE__: We use nodeFeatureRules to label PCI vendor and device and it's enabled by default using `nfd.deployNodeFeatureRules` flag
```
adrianchiris, line 60: nit: nodeFeatureRules should be a link.

adrianchiris: also its unclear in the context of this section what is nfd.deployNodeFeatureRules flag.

im not even sure helm values are regarded as "flags"

deployment/network-operator/README.md
wmousa: Done

README.md
wmousa: chart parameter

ykulazhenkov (approved):

adrianchiris (approved):

PR #572, opened 2023-07-18
Bump ginkgo gomega

adrianchiris:


adrianchiris:
as only tests are affected by this change and relevant CI jobs passed, we can go ahead and merge this one.

PR #573, opened 2023-07-18
Add default env vars for MOFED container

adrianchiris:
In case of older distros (kernels),
we need to explictly specify CREATE_IFNAMES_UDEV
so that netdev names are preserved.

when MOFED container loads the driver, it reports
additional information to kernel about Nvidia network devices that affect the way the OS names netdevs.

On newer distros (kernels) this is not needed as
inbox driver already reports this information.

Instruct MOFED container to install the
"name preserving udev" on the following OSes:

rhel < 9.0
ocp < 4.13
ubuntu < 22.04


adrianchiris:
created #574 the mofed state file keeps growing, at some point we need to chop it up.

adrianchiris:
@e0ne @ykulazhenkov PTAL, this is needed b4 next beta drop.

i also testing this one on kind cluster, it behaves as expected.

e0ne (approved):
LGTM

pkg/state/state_ofed.go
```diff
@@ -590,3 +593,69 @@ func (s *stateOFED) setEnvFromClusterWideProxy(cr *mellanoxv1alpha1.NicClusterPo
...
+	osName := nodeAttrs[nodeinfo.AttrTypeOSName]
+	osVer := nodeAttrs[nodeinfo.AttrTypeOSVer]
+	createIfnameUdevEnv := v1.EnvVar{Name: "CREATE_IFNAMES_UDEV", Value: "true"}
+
+	currVer, err := semver.NewVersion(osVer)
+	if err != nil {
+		return currentEnvs
+	}
+
+	ubuntuVer, err := semver.NewVersion("22.04")
```
ykulazhenkov, line 616: nit: use semver.MustParse
https://github.com/Masterminds/semver/blob/e06051f8fcc4c8b4a4990c337b9862a2448722e5/version.go#L230

pkg/state/state_ofed.go
```diff
@@ -590,3 +593,69 @@ func (s *stateOFED) setEnvFromClusterWideProxy(cr *mellanoxv1alpha1.NicClusterPo
 		)
 	}
 }
+
+// mergeWithDefaultEnvs returns env variables provided in currentEnvs merged with default
+// env variables for MOFED container, taking into account attributes of the k8s cluster.
+func (s *stateOFED) mergeWithDefaultEnvs(
+	currentEnvs []v1.EnvVar, nodeAttrs map[nodeinfo.AttributeType]string) []v1.EnvVar {
+	if envVarsWithGet(currentEnvs).Get("CREATE_IFNAMES_UDEV") != nil {
```
ykulazhenkov, line 601: nit: this construction is nice, but IMO creates unneeded complexity

maybe replace it with dumb loop?
```
	for _, v := range currentEnvs {
		if v == "CREATE_IFNAMES_UDEV" {
			// already exists don't overwrite
			return currentEnvs
		}
	}
```

ykulazhenkov (commented):
overall looks good, I added few nit comments which you can ignore

pkg/state/state_ofed.go
adrianchiris: was just thinking  its a matter of time until we need to handle additional vars. id prefer to keep as is  this one.

pkg/state/state_ofed.go
adrianchiris: nice! will rework it.

pkg/state/state_ofed.go
ykulazhenkov: ack

ykulazhenkov (approved):

PR #577, opened 2023-07-21
Configure drain skip labels explicitly on the net-op side

almaslennikov:
Previously, the logic for configuring skip labels was concealed in the k8s-operator-libs, undocumented and quite vague.
Reusing the already existing PodSelector mechanism, configure the drain skip labels in the upgrade controller

Corresponding PR in k8s-operator-libs: https://gitlab.com/nvidia/cloud-native/k8s-operator-libs/-/merge_requests/24


e0ne (approved):

controllers/upgrade_controller.go
```diff
@@ -59,6 +59,8 @@ const plannedRequeueInterval = time.Minute * 2
 // UpgradeStateAnnotation is kept for backwards cleanup TODO: drop in 2 releases
 const UpgradeStateAnnotation = "nvidia.com/ofed-upgrade-state"
 
+const UpgradeSkipDrainLabelSelector = "nvidia.com/ofed-driver-upgrade-drain.skip!=true"
```
adrianchiris, line 62: a few nits.

1. mind putting all of these consts inside a const block ? 
2. add comment about this const what it means

almaslennikov: Thanks for the feedback! Addressed it

controllers/mocks/ClusterUpgradeStateManager.go
```diff
@@ -0,0 +1,211 @@
+// Code generated by mockery v2.26.1. DO NOT EDIT.
```
rollandf, line 1: Maybe this file should be under `pkg/testing/mocks` ?

almaslennikov:
Part of https://github.com/Mellanox/network-operator/pull/600 now

PR #580, opened 2023-07-27
Limit maximum number of RDMA resources for Shared Device Plugin

e0ne:
There us hardware limitation of 128 GIDs per port in RocE mode. Since RocE v1 and v2 are enabled by default we have to limit to 63 devices per port to be allocated to workloads.

There is no such limitation for IPoIB mode, so user should update default value if needed.

This commit also makes rdmaHcaMax configurable via helm values.

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,7 +138,7 @@ spec:
         "configList": [
           {
             "resourceName": "rdma_shared_device_a",
-            "rdmaHcaMax": 1000,
+            "rdmaHcaMax": 63,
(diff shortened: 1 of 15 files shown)


adrianchiris (approved):
overall LGTM, can you update example folder as well ?

e0ne:
> overall LGTM, can you update example folder as well ?

examples updated

adrianchiris (approved):

PR #591, opened 2023-08-22
Remove unneeded scc

adrianchiris:


deploy/role.yaml
```diff
@@ -252,9 +294,113 @@ rules:
   - update
   - watch
 - apiGroups:
-  - ""
+  - mellanox.com
```
ykulazhenkov, line 297: in the ClusterRole we have access to `mellanox.com` apiGroup  with `'*'` resources. Do we really need to add each resource explicitly here?

adrianchiris: Thats auto-generated code, i did not edit this explicitly.

rollandf (approved):
LGTM

deploy/role.yaml
ykulazhenkov: It looks like at some point we manually added the wildcard rule. Ack, let's keep specific rules and wildcard as a fallback.

ykulazhenkov (approved):

deploy/role.yaml
adrianchiris: hmm lemme check this further.

adrianchiris: we got wildcards set in the following reconcilers:

https://github.com/Mellanox/network-operator/blob/84f986f33d0753e8dacfe946bca05bd994f85fe9/controllers/ipoibnetwork_controller.go#L53
https://github.com/Mellanox/network-operator/blob/84f986f33d0753e8dacfe946bca05bd994f85fe9/controllers/macvlannetwork_controller.go#L55
https://github.com/Mellanox/network-operator/blob/84f986f33d0753e8dacfe946bca05bd994f85fe9/controllers/nicclusterpolicy_controller.go#L58

hostdevice controller seems to be the only saint who doesnt use wildecards :)  

ill create an issue to remove the wildecards. it should be handled in a separate PR

PR #594, opened 2023-09-04
IPPool CRD NV-IPAM

rollandf:
NV-IPAM updates:

- Add support for IPPool CRD
- Deprecate the configmap
- Update templates with latest yamls configuration
- Update documentation
- Upgrade to v0.1.0


manifests/state-nv-ipam-cni/040-nv-ipam-node.yaml
ykulazhenkov, line 1: We have existing bug in the YAML, we missing support for custom tollerations for NVIPAM. Please, fix this as a part of this PR. 
```
        {{- if .Tolerations }}
        {{- .Tolerations | yaml | nindent 8 }}
        {{- end }}
```

rollandf: Done

config/rbac/role.yaml
```diff
@@ -248,6 +248,23 @@ rules:
   - get
   - list
   - watch
+- apiGroups:
```
ykulazhenkov, line 251: Do we need these permissions to the operator itself? When it interacts with these CRs?

controllers/nicclusterpolicy_controller.go
```diff
@@ -75,6 +75,8 @@ type NicClusterPolicyReconciler struct {
 // +kubebuilder:rbac:groups=coordination.k8s.io,resources=leases,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=batch,resources=cronjobs,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=config.openshift.io,resources=proxies,verbs=get;list;watch
+// +kubebuilder:rbac:groups=nv-ipam.nvidia.com,resources=ippools,verbs=get;list;watch;create;
```
ykulazhenkov, line 78: We will need to drop this if we will decide that we don't need access for ippools resources

deployment/network-operator/crds/nv-ipam.nvidia.com_ippools.yaml
ykulazhenkov, line 1: I don't have right answer for this, but do we really want to manage CRDs for sub-components with helm?
This can be problematic if we will decide to support nv-ipam in Openshift - we will need to include all CRs to a release bundle. As alternative we can deploy CRD with network-operator itself. WDYT?

rollandf: what do you mean by "deploy CRD with network-operator"?
I agree that for OpenShift this will not work

config/rbac/role.yaml
ykulazhenkov: ack

deployment/network-operator/crds/nv-ipam.nvidia.com_ippools.yaml
ykulazhenkov: I mean network-operator app itself can create CRD in the cluster

rollandf: We have done it with WhereAbout :
https://github.com/Mellanox/network-operator/blob/master/manifests/state-whereabouts-cni/0010-whereabouts.cni.cncf.io_ippools.yaml#L16

WDYT? Move it to there?

ykulazhenkov: If we already deploy CRDs as a part of manifests rendering for other components, then I think we should use the same method here.

rollandf: Ack

ykulazhenkov (approved):

almaslennikov (approved):

e0ne (approved):
LGTM. It colud be merged once all merge conflicts are resolved and CI passed

PR #602, opened 2023-09-21
Fix behaviour for OFED wait label to handle additional scenarios

ykulazhenkov:
New scenarios:
- set OFED wait to true when OFED is removed from CR and Pods are terminating
- clean OFED wait label from the node if it doesn't have NVIDIA nic anymore

### DIFF
--- a/controllers/nicclusterpolicy_controller.go
+++ b/controllers/nicclusterpolicy_controller.go
@@ -89,7 +89,7 @@ func (r *NicClusterPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Req
 			// Request object not found, could have been deleted after reconcile request.
 			// Owned objects are automatically garbage collected. For additional cleanup logic use finalizers.
 			// Return and don't requeue
-			err := r.clearMofedWaitLabel(ctx)
+			err := r.handleMOFEDWaitLabelsNoConfig(ctx)
 			if err != nil {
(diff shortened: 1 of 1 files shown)


e0ne (approved):

controllers/nicclusterpolicy_controller.go
```diff
@@ -182,28 +174,68 @@ func (r *NicClusterPolicyReconciler) updateNodeLabels(
...
 	for i := range nodes.Items {
-		patch := []byte(fmt.Sprintf(`{"metadata":{"labels":{%q:"false"}}}`, nodeinfo.NodeLabelWaitOFED))
-		err := r.Client.Patch(ctx, &nodes.Items[i], client.RawPatch(types.StrategicMergePatchType, patch))
-		if err != nil {
-			return errors.Wrapf(err, "unable to patch %s node label for node %s",
-				nodeinfo.NodeLabelWaitOFED, nodes.Items[i].Name)
+		node := nodes.Items[i]
+		labelValue := ""
+		if _, hasOFED := nodesWithOFEDContainer[node.Name]; hasOFED {
+			labelValue = "true"
```
almaslennikov, line 202: The name of the function is misleading in this case. Should it maybe be something like `UpdateMofedWaitLabels`?

ykulazhenkov: I renamed functions to reduce confusion

almaslennikov (approved):

e0ne (approved):

PR #612, opened 2023-09-27
Remove hostNetwork:true from nv-ipam-node spec

ykulazhenkov:
nv-ipam node has liveness and rediness probes and metrics endpoint. It is problematic to use hostNetwork: true because ports can be already used on the node by a different process. 

upstream nv-ipam node DS doesn't use hostNetwork https://github.com/Mellanox/nvidia-k8s-ipam/blob/11f148c7a7c6ed9a1daaeda927f01fda8331ec80/deploy/nv-ipam.yaml#L64


e0ne (approved):

adrianchiris:
note: later on we may need to support hostnetwork if this ipam is used for primary network cni. but for now its out of scope

PR #619, opened 2023-10-03
Add API for cluster type detection (Kubernetes/Openshift)

ykulazhenkov:
- New API for cluster type detection added
- use the API in all states

### DIFF
--- a/pkg/clustertype/cluster_type.go
+++ b/pkg/clustertype/cluster_type.go
@@ -1,0 +1,82 @@
+/*
+2023 NVIDIA CORPORATION & AFFILIATES
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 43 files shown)


adrianchiris (approved):

pkg/clustertype/cluster_type.go
```diff
@@ -0,0 +1,82 @@
...
+	osconfigv1 "github.com/openshift/api/config/v1"
+	"k8s.io/apimachinery/pkg/api/meta"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+/*
+ clustertype package provides information about the cluster type
+*/
+
+type Type string
```
e0ne, line 32: nit: Type -> ClusterType

e0ne (approved):

pkg/clustertype/cluster_type.go
ykulazhenkov: I think we can keep it as Type, because usually we will refer to this const as `clustertype.Type`

ykulazhenkov:
@e0ne @adrianchiris can we merge this?

PR #623, opened 2023-10-04
Add update permission for nicclusterpolicies/finalizers

ykulazhenkov:
Without this permission we will got `is forbidden: cannot set blockOwnerDeletion if an ownerReference refers to a resource you can't set finalizers on` error

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -266,6 +266,12 @@ rules:
   - update
   - watch
 - apiGroups:
+  - mellanox.com
+  resources:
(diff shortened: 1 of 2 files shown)


adrianchiris (approved):

adrianchiris:
@ykulazhenkov during my testing of PR #626  i was able to create a fully populated nicclusterpolicy without this change.

on k8s 1.27.2 so maybe its something related only to ocp.

adrianchiris:
also if we end up needing this:

1. please update in helm chart
2. please update in bundle (clusterserviceversion)

PR #629, opened 2023-10-05
Add support for partial resource removal (stale objects removal)

ykulazhenkov:
Add logic to cleanup objects which are not part of a state anymore


adrianchiris:
note: this introduces new functionality so its technically a feature :) (feat) and not a chore

pkg/state/state_skel.go
```diff
@@ -266,12 +266,48 @@ func (s *stateSkel) handleStateObjectsDeletion(ctx context.Context) (SyncState,
 	return SyncStateIgnore, nil
 }
 
-func (s *stateSkel) deleteStateRelatedObjects(ctx context.Context) (bool, error) {
+// remove stale object of the state, returns boolean which indicates if removal is in progress and
+// an error if failed to remove an object
+func (s *stateSkel) handleStaleStateObjects(ctx context.Context,
+	currentObjs []*unstructured.Unstructured) (bool, error) {
```
adrianchiris, line 272: nit: currentObjs -> desiredObjs ?

currentObjs sounds like the current objs we got in the cluster.

pkg/state/state_skel.go
```diff
@@ -266,12 +266,48 @@ func (s *stateSkel) handleStateObjectsDeletion(ctx context.Context) (SyncState,
...
+	reqLogger := log.FromContext(ctx)
+	reqLogger.V(consts.LogLevelInfo).Info(
+		"check state for stale objects", "State:", s.name)
+	objsToKeep := stateObjects{}
+	for _, o := range currentObjs {
+		gvkMap := objsToKeep[o.GroupVersionKind()]
```
adrianchiris, line 278: can you make "add" operation a member func of stateObjects ?

pkg/state/state_skel.go
```diff
@@ -266,12 +266,48 @@ func (s *stateSkel) handleStateObjectsDeletion(ctx context.Context) (SyncState,
...
+// is a mapping where GVK is a key and a map(set) with NamespacedNames is a value
+type stateObjects map[schema.GroupVersionKind]map[types.NamespacedName]struct{}
+
+func (s *stateSkel) deleteStateRelatedObjects(ctx context.Context, stateObjectsToKeep stateObjects) (bool, error) {
 	stateLabel := map[string]string{
 		consts.StateLabel: s.name,
 	}
 	found := false
 	for _, gvk := range getSupportedGVKs() {
+		typeObjectsToKeep := stateObjectsToKeep[gvk]
```
adrianchiris, line 307: while at it, consider adding a "Get" method as well.

mainly to reduce complexity.

adrianchiris (commented):
could you add a description in commit msg of what was done (in a more specific manner) ?

rollandf (approved):
LGTM

pkg/state/state_skel.go
ykulazhenkov: done

pkg/state/state_skel.go
ykulazhenkov: I added helper function for the stateObjects map

pkg/state/state_skel.go
ykulazhenkov: I added helper function for the stateObjects map

ykulazhenkov:
> could you add a description in commit msg of what was done (in a more specific manner) ?

@adrianchiris I update the commit message to include some details

PR #633, opened 2023-10-12
feat: add controller revision to managed resources

### LINKED ISSUE #614: Reduce amount of Kubernetes API calls
Currently network-operator does lot of update calls during each reconciliation loop. We should find a way to decrease amount of API calls and to speed up Reconcile loop.

ykulazhenkov:
Inject controller generated revision to all managed objects as annotation. 
The revision is later used to check if object update is required. 
This change should help to reduce amount of API calls to Kubernetes.

fixes #614

### DIFF
--- a/pkg/consts/consts.go
+++ b/pkg/consts/consts.go
@@ -38,4 +38,5 @@ const (
 	// that network-operator pod should be skipped during the drain operation which
 	// is executed by the upgrade controller.
 	OfedDriverSkipDrainLabelSelector = "nvidia.com/ofed-driver-upgrade-drain.skip!=true"
+	ControllerRevisionAnnotation     = "nvidia.network-operator.revision"
 )
--- a/pkg/revision/revision.go
+++ b/pkg/revision/revision.go
@@ -1,0 +1,69 @@
+/*
+ 2023 NVIDIA CORPORATION & AFFILIATES
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+     http://www.apache.org/licenses/LICENSE-2.0
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+*/
+
+package revision
+
+import (
+	"encoding/json"
+	"fmt"
+	"hash/fnv"
+	"strconv"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/Mellanox/network-operator/pkg/consts"
+)
+
+// ControllerRevision is the data to be stored as checkpoint
+type ControllerRevision uint32
+
+// CalculateRevision calculates controller revision for the object
+func CalculateRevision(o client.Object) (ControllerRevision, error) {
+	rev, err := getHash(o)
+	if err != nil {
+		return 0, fmt.Errorf("failed to compute controller revision for the object: %v", err)
+	}
+	return ControllerRevision(rev), nil
+}
+
+// GetRevision returns controller revision which is saved in the object.
+// returns 0 if revision is not set for the object
+func GetRevision(o client.Object) ControllerRevision {
+	val := o.GetAnnotations()[consts.ControllerRevisionAnnotation]
+	rev, err := strconv.ParseUint(val, 10, 32)
+	if err != nil {
+		return 0
+	}
+	return ControllerRevision(rev)
+}
+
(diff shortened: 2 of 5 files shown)


ykulazhenkov:
@e0ne @rollandf @adrianchiris Please take a look and share your thoughts regarding this change.

pkg/consts/consts.go
```diff
@@ -32,4 +32,5 @@ const (
 	NicClusterPolicyResourceName = "nic-cluster-policy"
 	OfedDriverLabel              = "nvidia.com/ofed-driver"
 	StateLabel                   = "nvidia.network-operator.state"
+	ControllerRevisionAnnotation = "nvidia.network-operator.revision"
```
adrianchiris, line 35: nit: object-revision ?

pkg/revision/revision.go
```diff
@@ -0,0 +1,69 @@
...
+func SetRevision(o client.Object, rev ControllerRevision) {
+	annotations := o.GetAnnotations()
+	if annotations == nil {
+		annotations = make(map[string]string)
+	}
+	annotations[consts.ControllerRevisionAnnotation] = strconv.FormatUint(uint64(rev), 10)
+	o.SetAnnotations(annotations)
+}
+
+// Get returns calculated checksum for the Root object
```
adrianchiris, line 60: nit: fix comment

pkg/revision/revision.go
```diff
@@ -0,0 +1,69 @@
...
+func getHash(o client.Object) (uint32, error) {
+	h := fnv.New32a()
+	data, err := json.Marshal(o)
+	if err != nil {
+		return 0, err
+	}
+	_, _ = h.Write(data)
```
adrianchiris, line 67: nit: check error / written bytes to be on the safe side ?

pkg/state/state_skel.go
```diff
@@ -214,27 +215,39 @@ func (s *stateSkel) createOrUpdateObjs(
...
-			// object created successfully
-			continue
-		}
-		if !k8serrors.IsAlreadyExists(err) {
-			// Some error occurred
+		desiredRev, err := revision.CalculateRevision(desiredObj)
+		if err != nil {
 			return err
 		}
+		revision.SetRevision(desiredObj, desiredRev)
```
adrianchiris, line 222: nit: you only need to set this if an update is required.

that said, im fine with either keeping it here or moving it to after L247

adrianchiris (commented):
Approach looks good to me.

the only caveat i can think of is someone tinkering with the obj after it was created.

while some fields may be changed (and some are assured to be changed) i think what we care bout is the spec field. 

should we also save the new object generation in annotation ?
that way if for some reason spec was updated we would catch it.

thoughts ?

rollandf (approved):
LGTM

ykulazhenkov:
@adrianchiris 

> should we also save the new object generation in annotation ?
that way if for some reason spec was updated we would catch it.

How you want to retrieve the latest object spec? (some changes to object are applied by k8s components in async manner and you will not get them immediately in the API resp after create/update call) I don't know reliable method to understand which changes are "legit" (applied by k8s control plane in async mode) and which are not (applied by an user).  Do you have an idea how to handle this?

adrianchiris:
> how you want to retrieve the latest object spec? (some changes to object are applied by k8s components in async manner and you will not get them immediately in the API resp after create/update call) 

these changes affect  the object  generation ? do you have a specific example in mind ?

ykulazhenkov:
@adrianchiris 

> these changes affect the object generation ?

`meta.generation` is not available for all object types. 

> do you have a specific example in mind ?

pod.spec.nodeName, CSI things such as PersistentVolumeClaim, something else

adrianchiris (commented):
@ykulazhenkov im Fine with the proposed changes.
we dont have a use-case of user changing the created objects after creation
so the limitation of the mechanism is fine IMO.
(and may be useful for debugging)

if you could address the nits, would be great.

personally id prefer to avoid the term revision as to not confuse with k8s obj revision which is an incrementing number. here its more of a hash of the object as originally created by the operator. (each time i read this PR i face the same issue)

pkg/revision/revision.go
ykulazhenkov: the doc says: 
```
// Write (via the embedded io.Writer interface) adds more data to the running hash.
// It never returns an error.
```

pkg/revision/revision.go
ykulazhenkov: fixed

pkg/state/state_skel.go
ykulazhenkov: We also need to set revision when we create the object at line 236.

adrianchiris (approved):

ykulazhenkov:
@adrianchiris I think I addressed all nits in this PR. All other things were discussed with you offline. Can you check this PR one more time?

PR #639, opened 2023-10-16
chore: Removed deploy directory

rollandf:
Removing `deploy` directory to avoid duplications.

Deployment should be done using `make deploy`,
using the `config` dir and kustomize.


adrianchiris (approved):
LGTM! thx for cleaning this 🚀

PR #646, opened 2023-10-18
feat: allow to configure CNI bin directory

almaslennikov:
We need to add the ability to configure CNI bin directories,
as the default `/opt/cni/bin` might not be available in some use-cases (OCP)


hack/templates/values/values.template
```diff
@@ -25,6 +25,9 @@ psp:
 
 upgradeCRDs: true
 
+global:
```
adrianchiris, line 28: why would you not put it after deployCR with the rest of nicclusterpolicy parameters.

also id drop the "global" as we didnt  add it to begin with for the top level items of the CR deployed by the chart.

api/v1alpha1/nicclusterpolicy_types.go
```diff
@@ -235,6 +235,10 @@ type NicClusterPolicySpec struct {
 	NvIpam                 *NVIPAMSpec              `json:"nvIpam,omitempty"`
 	PSP                    *PSPSpec                 `json:"psp,omitempty"`
 	NicFeatureDiscovery    *NICFeatureDiscoverySpec `json:"nicFeatureDiscovery,omitempty"`
+	// Configure the directory where cni binaries would be deployed to
```
adrianchiris, line 238: this ends up in the description of the crd.

please add a "." at  the end of the first sentence so it reads properly.

also we do have a default today dont we ? defined today in each of the states.
all assuming `/opt/cni/bin` IIRC.

what logic is broken related to openshift ?
also i dont think the end user needs to see this in CRD definition.

deployment/network-operator/charts/sriov-network-operator/templates/operator.yaml
```diff
@@ -88,6 +88,7 @@ spec:
             - name: RELEASE_VERSION
               value: {{ .Release.AppVersion }}
             - name: SRIOV_CNI_BIN_PATH
-              value: {{ .Values.operator.cniBinPath }}
+              # If global value is available, use that, otherwise, use the local one for the operator
```
adrianchiris, line 91: This chart should be kept aligned with upstream.
starting to keep changes locally is bad IMO.

the chart already defines cniBinPath

manifests/state-ipoib-cni/0050-ipoib-cni-ds.yml
```diff
@@ -74,9 +74,12 @@ spec:
       volumes:
         - name: cnibin
           hostPath:
+            # First we try to set the user-set value, then fallback to defaults for Openshift / K8s
```
adrianchiris, line 77: lets do this logic in the state and not the template.
i think it should be applied generically for all relevant (CNI related) states, then we  just template the value.

1. if  provided - take the value
2. else
2.1  if openshift -> /var/lib/cni/bin
2.2 else -> /opt/cni/bin

adrianchiris (changes requested):

deployment/network-operator/charts/sriov-network-operator/templates/operator.yaml
adrianchiris: that means -> end user is in charge of providing this path twice in the chart 
since helm does not support templating in values.yaml previous values.

hack/templates/values/values.template
almaslennikov: Removed the 'global' section and put it together with NicClusterPolicy values

adrianchiris (commented):
Thinking  on this more, this should really be a static configuration, (e.g passed to operator via env var)
same as being done for sriov  operator. I see no use-case where one might want to "update" this path after cluster deployment so better avoid API changes in CR for now.

ykulazhenkov:
> Thinking on this more, this should really be a static configuration, (e.g passed to operator via env var)
> same as being done for sriov operator. I see no use-case where one might want to "update" this path after cluster deployment so better avoid API changes in CR for now.

I agree with this point

ykulazhenkov (commented):
> Thinking on this more, this should really be a static configuration, (e.g passed to operator via env var)
> same as being done for sriov operator. I see no use-case where one might want to "update" this path after cluster deployment so better avoid API changes in CR for now.

I agree with this point

deployment/network-operator/templates/delete-cr-hook.yaml
ykulazhenkov, line 1: Please, rebase the PR

pkg/staticinfo/static_info.go
```diff
@@ -0,0 +1,44 @@
...
+
+package staticinfo
+
+/*
+ static package provides static cluster information, required by network-operator
+*/
+
+// Provider provides static cluster attributes
+type Provider interface {
+	GetCniBinDirectory() string
```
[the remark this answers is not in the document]
almaslennikov: Makes sense, remade the interface in according to your suggestion

almaslennikov:
@adrianchiris @ykulazhenkov I've addressed your comments, PTAL

pkg/state/state_nv_ipam_cni.go
```diff
@@ -62,8 +63,8 @@ type stateNVIPAMCNI struct {
 
 type nvipamRuntimeSpec struct {
 	runtimeSpec
-	// is true if cluster type is Openshift
-	IsOpenshift bool
+	IsOpenshift     bool
+	CniBinDirectory string
```
adrianchiris, line 67: here its in runntimeSpec while the reset is in the *RenderData ?
is there a particular reason ? should be align with the rest of the  states  ?

adrianchiris (changes requested):
one comment regarding aligning where cni bin dir parameter is located in the state "render data"
apart o f that need to fix lint and we are good 2 go !

pkg/state/state_nv_ipam_cni.go
almaslennikov: Addressed. Now all CNIs use a common `cniRuntimeSpec`

deployment/network-operator/README.md
ykulazhenkov, line 1: Are these changes related to the PR?

ykulazhenkov (approved):

ykulazhenkov:
LGTM, only question about README.md update

deployment/network-operator/README.md
almaslennikov: Thanks for catching, left only the related changes in README

adrianchiris (approved):

PR #593, opened 2023-08-31
feat: Add support for Admission Controller

wmousa:
This change implements validating admission controller for supported CRDs NicClusterPolicy, HostDeviceNetwork, and MacvlanNetwork, and IPoIBNetwork.
Currently, this commit covers:
 - Using operator-sdk to scaffold the webhook for the CRDS
 - Implementing validation rules for NicClusterPolicy.
   - IBKubernetes.PKeyGUIDPoolRangeStart
   - IBKubernetes.PKeyGUIDPoolRangeEnd
   - OFEDDriver.Version
   - RdmaSharedDevicePlugin.Config
     - Configuration is a valid JSON and check its schema
     - Resource name is valid for k8s.
     - At least one of the supported selectors exists. 
     - All selectors are strings.
   - SriovNetworkDevicePlugin.Config.
     - Configuration is a valid JSON and check its schema.
     - Resource name is valid for k8s.
     - At least one of the supported selectors exists.
     - All selectors are strings.
 - Implemented validation rules for HostDeviceNetwork.
   - ResourceName

PR #647, opened 2023-10-18
bug: delete NCP on Helm uninstall

rollandf:
In PR #593, the creation of the NicClusterPolicy CR via Helm has been marked to be done with Helm "post-install,post-upgrade".

As a result, when Helm was uninstalled, the CR was not deleted.

In order to keep previous behavior, a new post-delete hook is introduced to delete the CR if it was created by Helm.


adrianchiris:
nit: could you rename the file to be suffixed with "-hook"
just so its  clearer what this file is.

(we should rename the one on insstall as well)

e0ne (approved):
LGTM. It could be merged once CI succeed

PR #649, opened 2023-10-23
chore: Bump MOFED version to 23.10-0.4.1.0

e0ne:
Bump MOFED version to 23.10-0.4.1.0


e0ne:
image_scan CI isn't related to this changes

PR #650, opened 2023-10-23
fix: validate resourcePrefix in NCP DP

rollandf:
In the admission controller, the `resourcePrefix` field in Device Plugins configuration is now validated to be a valid FQDN.

### DIFF
--- a/api/v1alpha1/nicclusterpolicy_webhook.go
+++ b/api/v1alpha1/nicclusterpolicy_webhook.go
@@ -34,6 +34,12 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
 )
 
+const (
+	fqdnRegex              = `^[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\.[a-zA-Z]{2,})+$`
+	sriovResourceNameRegex = `^([A-Za-z0-9][A-Za-z0-9_.]*)?[A-Za-z0-9]$`
+	rdmaResourceNameRegex  = `^([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9]$`
+)
+
 // log is for logging in this package.
 var nicClusterPolicyLog = logf.Log.WithName("nicclusterpolicy-resource")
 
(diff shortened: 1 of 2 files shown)


ykulazhenkov (approved):

api/v1alpha1/nicclusterpolicy_webhook.go
```diff
@@ -229,6 +226,28 @@ func (dp *DevicePluginSpec) validateSriovNetworkDevicePlugin(fldPath *field.Path
 	return allErrs
 }
 
+func validateResourceNamePrefix(resource map[string]interface{},
+	allErrs field.ErrorList, fldPath *field.Path, dp *DevicePluginSpec) (bool, field.ErrorList) {
+	resourceName := resource["resourceName"].(string)
+	if !isValidSriovNetworkDevicePluginResourceName(resourceName) {
```
adrianchiris, line 232: can we replace this one with isValidFQDN as well ?

api/v1alpha1/nicclusterpolicy_webhook.go
```diff
@@ -347,6 +375,12 @@ func isValidRdmaSharedDevicePluginResourceName(resourceName string) bool {
 	return resourceNameRegex.MatchString(resourceName)
 }
 
+func isValidFQDN(input string) bool {
+	pattern := `^[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\.[a-zA-Z]{2,})+$`
```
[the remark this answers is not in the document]
rollandf: It is for the Top Level Domain (.com, .io , .ca)

api/v1alpha1/nicclusterpolicy_webhook.go
rollandf: I am not sure that the resource name has the same pattern as an FQDN

api/v1alpha1/nicclusterpolicy_webhook.go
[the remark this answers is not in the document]
adrianchiris: cool thx.

api/v1alpha1/nicclusterpolicy_webhook.go
adrianchiris: looking  at device plugin : https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin/blob/ea2e52c4ae8bf48c17ec666f605cb66fd03a3d49/pkg/utils/utils.go#L274

lets keep this as is. im not sure k8s will actually complain if it was fqdn.

api/v1alpha1/nicclusterpolicy_webhook.go
adrianchiris, line 379: could you put this one into a const ? 
then use it here and in the err msg when checking resource name ?

you can try to use raw string for it: https://www.golangprograms.com/golang-raw-string-literals-and-interpreted-string-literals.html

if it doesnt work, im ok with keeping it as is.

also while at it,  would be great if you did the same for the resource name regex as well.

rollandf: Done

adrianchiris (approved):

PR #657, opened 2023-11-03
chore: Update MOFED and OCP version

e0ne:
Update MOFED and OCP version  for RC1


e0ne:
nic_operator_helm CI   failure will be fixed with a new sriov operator version

PR #667, opened 2023-11-13
fix: OFED wait label behaviour

ykulazhenkov:
Fix behaviour of OFED wait label when
there is no OFED spec in the NicClusterPolicy.

### DIFF
--- a/controllers/nicclusterpolicy_controller.go
+++ b/controllers/nicclusterpolicy_controller.go
@@ -97,10 +97,13 @@ func (r *NicClusterPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Req
 			// Request object not found, could have been deleted after reconcile request.
 			// Owned objects are automatically garbage collected. For additional cleanup logic use finalizers.
 			// Return and don't requeue
-			err := r.handleMOFEDWaitLabelsNoConfig(ctx)
+			shouldRequeue, err := r.handleMOFEDWaitLabelsNoConfig(ctx)
 			if err != nil {
 				reqLogger.V(consts.LogLevelError).Error(err, "Fail to clear Mofed label on CR deletion.")
 				return reconcile.Result{}, err
+			}
+			if shouldRequeue {
+				return r.requeue()
 			}
 			return reconcile.Result{}, nil
 		}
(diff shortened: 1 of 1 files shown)


controllers/nicclusterpolicy_controller.go
```diff
@@ -212,24 +225,35 @@ func (r *NicClusterPolicyReconciler) handleMOFEDWaitLabelsNoConfig(ctx context.C
 		} else if node.GetLabels()[nodeinfo.NodeLabelMlnxNIC] == "true" {
 			labelValue = "false"
 		}
-		if node.GetLabels()[nodeinfo.NodeLabelWaitOFED] == labelValue {
-			// already has the right value
-			continue
-		}
 		if err := setOFEDWaitLabel(ctx, r.Client, node.Name, labelValue); err != nil {
```
adrianchiris, line 228: btw should we still attempt to patch node obj if its already patched to the correct value ?
or you want to avoid cache edge cases when node comes back to cluster with/without mlnx nic ?

controllers/nicclusterpolicy_controller.go
```diff
@@ -141,58 +144,68 @@ func (r *NicClusterPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
-func (r *NicClusterPolicyReconciler) handleMOFEDWaitLabelsNoConfig(ctx context.Context) error {
+// on the node (probably in the terminating state).
+// returns true if requeue (resync) is required
+func (r *NicClusterPolicyReconciler) handleMOFEDWaitLabelsNoConfig(ctx context.Context) (bool, error) {
+	reqLogger := log.FromContext(ctx)
 	nodesWithOFEDContainer := map[string]struct{}{}
 	pods := &corev1.PodList{}
 	if err := r.Client.List(ctx, pods, client.MatchingLabels{consts.OfedDriverLabel: ""}); err != nil {
-		return errors.Wrap(err, "failed to list pods")
+		return true, errors.Wrap(err, "failed to list pods")
```
adrianchiris, line 208: in case of error first return param should be ignored by the caller, maybe just return false in case of error as its the default for bool ?

adrianchiris: some thought drill: what would you return in case the resync val was *bool ? nil ?

controllers/nicclusterpolicy_controller.go
```diff
@@ -212,24 +225,35 @@ func (r *NicClusterPolicyReconciler) handleMOFEDWaitLabelsNoConfig(ctx context.C
 		} else if node.GetLabels()[nodeinfo.NodeLabelMlnxNIC] == "true" {
 			labelValue = "false"
 		}
-		if node.GetLabels()[nodeinfo.NodeLabelWaitOFED] == labelValue {
-			// already has the right value
-			continue
```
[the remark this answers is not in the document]
ykulazhenkov: I decided to always update labels here not to rely on the cached Pod info

controllers/nicclusterpolicy_controller.go
ykulazhenkov: yes, I decided not to rely on cached data for Pods. We don't watch Pods in the API that means that we may have no chance to fix the state on the Pod update event.

controllers/nicclusterpolicy_controller.go
ykulazhenkov: Agree. I update the code to return false in case of error

controllers/nicclusterpolicy_controller.go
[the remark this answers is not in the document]
ykulazhenkov: this is to avoid cache races/edge cases

controllers/nicclusterpolicy_controller.go
ykulazhenkov: we always patch the object to avoid possible cache edge cases

rollandf (approved):

adrianchiris (approved):

e0ne (approved):

PR #672, opened 2023-11-14
fix:  preserve NVIPAM CM after upgrade

rollandf:
In order to allow migration from config map to IP Pool CRs in NV IPAM, the config map needs to be kept.
    
 - Removing the state label will make sure it is not deleted as a stale object.
 - Remove the helm hook annotation on NiClusterPolicy CR template, that causes delete and recreate of CR and the delete-cr hook. Note that this will prevent CR validation in install stage for admission controller.
  - Make sure to stop existing Network Operator before updating CR, so that it does not delete the config map since config is removed.

### DIFF
--- a/deployment/network-operator/templates/mellanox.com_v1alpha1_nicclusterpolicy_cr.yaml
+++ b/deployment/network-operator/templates/mellanox.com_v1alpha1_nicclusterpolicy_cr.yaml
@@ -18,8 +18,6 @@ apiVersion: mellanox.com/v1alpha1
 kind: NicClusterPolicy
 metadata:
   name: nic-cluster-policy
-  annotations:
-    helm.sh/hook: post-install,post-upgrade
 spec:
   {{- if .Values.nodeAffinity }}
   nodeAffinity:
--- a/pkg/migrate/migrate.go
+++ b/pkg/migrate/migrate.go
@@ -18,11 +18,14 @@ package migrate
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/go-logr/logr"
 	v1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
 	apiErrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/Mellanox/network-operator/pkg/config"
@@ -46,6 +49,11 @@ func migrate(ctx context.Context, log logr.Logger, c client.Client) error {
 		// not critical for the operator operation, safer to ignore
 		log.V(consts.LogLevelWarning).Info("ignore error during whereabouts CronJob removal")
 	}
+	if err := removeStateLabelFromNVIpamConfigMap(ctx, log, c); err != nil {
+		// critical for the operator operation, will fail NVIPAM migration
+		log.V(consts.LogLevelError).Error(err, "error trying to remove state label on NV IPAM configmap")
+		return err
+	}
 	return nil
 }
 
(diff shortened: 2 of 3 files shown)


ykulazhenkov (dismissed):

pkg/migrate/migrate.go
```diff
@@ -46,6 +49,10 @@ func migrate(ctx context.Context, log logr.Logger, c client.Client) error {
 		// not critical for the operator operation, safer to ignore
 		log.V(consts.LogLevelWarning).Info("ignore error during whereabouts CronJob removal")
 	}
+	if err := removeStateLabelFromNVIpamConfigMap(ctx, log, c); err != nil {
```
e0ne, line 52: Why do we need to return error if we ignore it? We've got logging inside  removeStateLabelFromNVIpamConfigMap so no need to return an error

e0ne (changes requested):

pkg/migrate/migrate.go
rollandf: I followed the same pattern as `removeWhereaboutsIPReconcileCronJob`.
WDYT?

ykulazhenkov: I think we should keep it as is. `removeStateLabelFromNVIpamConfigMap` function do the action and an error can happen during this action - it is fair to return the error from the function.  It is up to caller to decide what to do with the error.

ykulazhenkov: and this should not block the PR for sure

e0ne (commented):
remove 'request changes' label since we've got the same patter in other part of code

pkg/migrate/migrate.go
ykulazhenkov: OK, I have one more thing. It is pretty safe to run `removeWhereaboutsIPReconcileCronJob` in a "best effort" mode and ignore any error if it happens (we will have an unneeded Job in the cluster, but this will not break anything). But with `removeStateLabelFromNVIpamConfigMap`, we are in a bit different situation - if we fail to remove the label, we will probably lose all nv-ipam allocations. Maybe we should crash in this case and rely on k8s to restart the Pod? WDYTH?

ykulazhenkov (commented):
I tested the PR on my env. I tried to upgrade from 23.7 to the version from this PR. All nv-ipam allocations were preserved.  I added one comment about the error handling for annotation removal call - maybe we should react on an error

pkg/migrate/migrate.go
rollandf: Agreed, returning error to main.

ykulazhenkov (approved):

e0ne (approved):
LGTM. It could be merged once CI pass

deployment/network-operator/templates/mellanox.com_v1alpha1_nicclusterpolicy_cr.yaml
```diff
@@ -18,8 +18,6 @@ apiVersion: mellanox.com/v1alpha1
 kind: NicClusterPolicy
 metadata:
   name: nic-cluster-policy
-  annotations:
-    helm.sh/hook: post-install,post-upgrade
```
adrianchiris, line 22: Idea (needs to be checked):

introduce scale down of netop deployment on post upgrade. deploy CR as hook with higher weight, scale back up as hook with even higher weight 

on post install we just deploy CR 

on post delete we delete CR

use  helm.sh/resource-policy: keep  -> should not delete existing resource (i hope it will do resouce update)

migration logic - remove the label from config map to prevent operator from removing the config map


with this i believe we can keep admission controller logic.

adrianchiris: Update:

after internal discussion and testing with @rollandf this will not work.

pkg/migrate/migrate.go
```diff
@@ -68,3 +76,36 @@ func removeWhereaboutsIPReconcileCronJob(ctx context.Context, log logr.Logger, c
...
+	}
+	err := c.Get(ctx, key, cfg)
+	if apiErrors.IsNotFound(err) {
+		log.V(consts.LogLevelDebug).Info("NVIPAM config map not found, skip remove state label")
+		return nil
+	} else if err != nil {
+		log.V(consts.LogLevelError).Error(err, "fail to get NVIPAM configmap")
+		return err
+	}
+	log.V(consts.LogLevelDebug).Info("clear State label from NVIPAM configmap")
```
adrianchiris, line 98: can we first check the label exists ? or will patch op suceed if label is not there ?

rollandf: Add a check before the patch

adrianchiris (approved):

PR #675, opened 2023-11-20
docs: Add force flag to Helm upgrade

rollandf:
Add force flag  to Helm upgrade

### DIFF
--- a/deployment/network-operator/README.md
+++ b/deployment/network-operator/README.md
@@ -266,7 +266,7 @@ Known limitations:
 These limitations will be addressed in future releases.
 
 > __NOTE__: changes which were made directly in NicClusterPolicy CR (e.g. with `kubectl edit`)
-> will be overwritten by Helm upgrade
+> will be overwritten by Helm upgrade due to the `force` flag.
 
 ### Temporary disable network-operator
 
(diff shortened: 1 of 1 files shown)


ykulazhenkov (approved):

deployment/network-operator/README.md
```diff
@@ -285,7 +285,7 @@ You have to wait for network-operator POD to remove before proceeding.
 ### Apply Helm chart update
 
 ```
-helm upgrade -n network-operator  network-operator mellanox/network-operator --version=<VERSION> -f values-<VERSION>.yaml
+helm upgrade -n network-operator  network-operator mellanox/network-operator --version=<VERSION> -f values-<VERSION>.yaml --force
```
adrianchiris, line 288: @rollandf since helm will use a replace when force is provided. it will use a replace strategy[1]
how will this affect MOFED container ? will the ds be deleted and recreated ?

if so, i still believe we will be OK, with caveat that we will go through inbox driver first.
(driver container deleted, new ds starts, mofed upgrade flow from inbox kicks in)

that said we need to merge this first.

[1] https://github.com/helm/helm/blob/82195652495d757d87c1a01f640973f20ca141b0/pkg/kube/client.go#L672

rollandf: Not sure I see why it should affect the MOFED container.
If the MOFED version change, the image will be updated, but the pods are not restarted (only onDelete).

adrianchiris: this is not an issue you are right.

adrianchiris (approved):

PR #679, opened 2023-11-20
chore: Remove config from NVIPAM spec

rollandf:
Remove config from NVIPAM spec. IPPool CR is used to configure NVIPAM


api/v1alpha1/nicclusterpolicy_types.go
```diff
@@ -211,8 +211,8 @@ type IBKubernetesSpec struct {
 type NVIPAMSpec struct {
 	// Enable deployment of the validation webhook
 	// +kubebuilder:default:=false
-	EnableWebhook       bool `json:"enableWebhook,omitempty"`
-	ImageSpecWithConfig `json:""`
+	EnableWebhook bool `json:"enableWebhook,omitempty"`
+	ImageSpec     `json:""`
```
adrianchiris, line 215: k8s is OK with removing a field from CRD while there are objects with this field present right ?
what would be the behaviour ?(config field disappears ?)

rollandf: The config field disappears

adrianchiris (approved):

adrianchiris:
@ykulazhenkov can we merge this one ?

PR #685, opened 2023-11-21
chore: update all dependencies for the operator

ykulazhenkov:
This PR includes following changes:
-  all dependencies are upgraded to the latest version
- `controler-runtime` package changed the API, this PR also includes migration to the new API
- `k8s-operator-libs` package changed the API, this PR removes logic which is not required anymore and does migration to the new API
- move from deprecated `wait.Poll` function to `wait.PollUntilContextTimeout`


ykulazhenkov:
This PR is prerequisite for #600 (`k8s-operator-libs` update required)

pkg/consts/consts.go
```diff
@@ -34,4 +34,7 @@ const (
 	StateLabel                   = "nvidia.network-operator.state"
 	DefaultCniBinDirectory       = "/opt/cni/bin"
 	OcpCniBinDirectory           = "/var/lib/cni/bin"
+	// OfedDriverSkipDrainLabelSelector contains labelselector which is used to indicate
+	// that OFED pod should be skipped during the drain operation which is executed by the upgrade controller.
```
adrianchiris, line 38: we use this for the operator pod no ?

ykulazhenkov: Yes, I will fix the coment

adrianchiris (approved):
Overall LGTM,

one small comment RE comment in new const once addressed can be merged

pkg/consts/consts.go
ykulazhenkov: fixed

rollandf (approved):

PR #686, opened 2023-11-22
fix: remove stale objects handling from *network CR states

ykulazhenkov:
All CR instances are sharing the same state name, but they handled by a separate Reconcile loop. 
Stale objects removal logic doesn't support this scenario.


e0ne (approved):
LGTM

rollandf (approved):

PR #696, opened 2023-11-29
feat: Allow to configure resource requirements of individual containers

almaslennikov:
Allow to configure [resource requirements](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/) for each of individuals containers deployed by network-operator.
Each state might deploy one or more containers, user needs to specify the name of the specific container to set its resource requirements.
By default, requirements are not applied or set to previously set defaults.


deployment/network-operator/templates/mellanox.com_v1alpha1_nicclusterpolicy_cr.yaml
```diff
@@ -69,7 +72,7 @@ spec:
   {{- end }}
   {{- if .Values.rdmaSharedDevicePlugin.deploy }}
   rdmaSharedDevicePlugin:
-    # {{ required "A valid value for .Values.rdmaSharedDevicePlugin.resources is required" .Values.rdmaSharedDevicePlugin.resources }}
+    # {{ required "A valid value for .Values.rdmaSharedDevicePlugin.containerResources is required" .Values.rdmaSharedDevicePlugin.containerResources }}
```
ykulazhenkov, line 75: Is this auto renaming glitch or an expected change?

deployment/network-operator/templates/mellanox.com_v1alpha1_nicclusterpolicy_cr.yaml
```diff
@@ -82,8 +85,8 @@ spec:
     config: |
       {
         "configList": [
-          {{- $length := len .Values.rdmaSharedDevicePlugin.resources }}
-          {{- range $index, $element := .Values.rdmaSharedDevicePlugin.resources }}
+          {{- $length := len .Values.rdmaSharedDevicePlugin.containerResources }}
```
ykulazhenkov, line 88: Is this auto renaming glitch or an expected change?

deployment/network-operator/templates/mellanox.com_v1alpha1_nicclusterpolicy_cr.yaml
```diff
@@ -111,8 +114,8 @@ spec:
     config: |
       {
         "resourceList": [
-          {{- $length := len .Values.sriovDevicePlugin.resources }}
-          {{- range $index, $element := .Values.sriovDevicePlugin.resources }}
+          {{- $length := len .Values.sriovDevicePlugin.containerResources }}
```
ykulazhenkov, line 117: Is this auto renaming glitch or an expected change?

deployment/network-operator/values.yaml
```diff
@@ -112,6 +112,7 @@ sriov-network-operator:
 # General Operator related values
 # The operator element allows to deploy network operator from an alternate location
 operator:
+  resources: {}
```
ykulazhenkov, line 115: Do we need this?

deployment/network-operator/values.yaml
```diff
@@ -174,6 +175,10 @@ ofedDriver:
   # env:
   #   - name: EXAMPLE_ENV_VAR
   #     value: example_env_var_value
+  containerResources:
+    - name: "mofed-container"
```
ykulazhenkov, line 179: Do we want to include all resources here explicitly? Maybe It is better to provide a single commented example? I want to avoid situation when we will need to always update the list when adding a new container

deployment/network-operator/values.yaml
almaslennikov: This is a placeholder left to remind that operator resources can also be customized

deployment/network-operator/values.yaml
almaslennikov: Do you mean a single example like this?
```
containerResources:
  - name: "mofed-container"
     // requests:
     //   "cpu": "500Mi"
     //   "memory": "300Mi"
     // limits:
     //   "cpu": "500Mi"
     //   "memory": "300Mi"
 ```

**Comment by almaslennikov** [comment]
retest-nic_operator_helm

File: api/v1alpha1/nicclusterpolicy_types.go [hunk 6]
```diff
@@ -184,6 +185,21 @@ type SecondaryNetworkSpec struct {
 	IpamPlugin *ImageSpec `json:"ipamPlugin,omitempty"`
 }
 
+// ResourceRequirements describes the compute resource requirements.
+type ResourceRequirements struct {
+	Name string `json:"name"`
```
vasrem, line 190: nit: would be nice to add a comment that this is the name of the container

api/v1alpha1/validator/nicclusterpolicy_webhook.go
```diff
@@ -472,69 +479,129 @@ func validateRepository(repo string, allErrs field.ErrorList, fp *field.Path, ch
...
 	}
 	return allErrs
 }
 
-func validateContainerResourcesIfNotNil(resource interface{}, allErrs field.ErrorList,
+func validateContainerResourcesIfNotNil(
+	resource *stateRenderData, policy *v1alpha1.NicClusterPolicy, allErrs field.ErrorList,
 	fp *field.Path, fieldName string) field.ErrorList {
-	if resource == nil {
+	if resource.spec == nil {
```
vasrem, line 559: Should we keep checking if resource is nil like we used to?

api/v1alpha1/validator/nicclusterpolicy_webhook.go
```diff
@@ -435,6 +479,159 @@ func validateRepository(repo string, allErrs field.ErrorList, fp *field.Path, ch
...
+
+func validateResourceRequirements(resources []v1alpha1.ResourceRequirements,
+	policy *v1alpha1.NicClusterPolicy, manifestRenderer state.ManifestRenderer,
+	allErrs field.ErrorList, fp *field.Path, child string) field.ErrorList {
+	if len(resources) == 0 {
+		allErrs = append(allErrs, field.Invalid(fp.Child(child).Child("containerResources"),
+			resources, "should not be empty if declared"))
+	}
+
+	supportedContainerNames, err := state.ParseContainerNames(manifestRenderer, policy, nicClusterPolicyLog)
```
rollandf, line 592: So we are rendering all the manifests on each webhook calls?
Isn't it a little heavy?

manifests/state-container-networking-plugins/0010-container-networking-plugins-ds.yml
```diff
@@ -55,13 +55,30 @@ spec:
           imagePullPolicy: IfNotPresent
           securityContext:
             privileged: true
+          {{- $containerResourcesFound := false }}
```
ykulazhenkov, line 58: Maybe we can try to create a template function which will contain this logic
something like 
`addResources(.CrSpec.ContainerResources,  "cni-plugins",  "100m", "50Mi", "100m", 50Mi")`
or 
`addResources(.CrSpec.ContainerResources,  "cni-plugins",  objectWithDefaults)`

adrianchiris (commented):
maybe we need to split to several PRs.

1. move validator to its own pkg
2. add container resources to operator
3. add validation to container resources (webhook)
4. add container resources to helm (with the continuity test)

for me, its hard to review this as a single bulk.

manifests/state-container-networking-plugins/0010-container-networking-plugins-ds.yml
adrianchiris: i think, if we use for the template a map keyed by container name then we can check for key existance then with if/else use req/limits from map entry or default as provided by template.

for those templates that dont need default then we drop the "else" part

thoughts ?


presonally;
1. i want to avoid inventing sort of a DSL with special functions
2. i dont want to move template information to code (e.g defaults will be part of state)

deployment/network-operator/values.yaml
```diff
@@ -229,6 +234,10 @@ rdmaSharedDevicePlugin:
   # it must be provided by the user when deploying the chart
   # each entry in the resources element will create a resource with the provided <name> and list of devices
   # example:
+  containerResources:
+    - name: "rdma-shared-dp"
```
adrianchiris, line 238: why do we want to provide empty default value ?
why not comment it out ?

empty may actually mean: i want empty req and limits and let container-runtime do whatever it wants

deployment/network-operator/README.md
```diff
@@ -428,17 +429,19 @@ imagePullSecrets:
 | `ofedDriver.upgradePolicy.drain.deleteEmptyDir`             | bool   | `false`                           | continue even if there are pods using emptyDir                                                                                                                                |
 | `ofedDriver.upgradePolicy.waitForCompletion.podSelector`    | string | not set                           | specifies a label selector for the pods to wait for completion before starting the driver upgrade                                                                             |
 | `ofedDriver.upgradePolicy.waitForCompletion.timeoutSeconds` | int    | not set                           | specify the length of time in seconds to wait before giving up for workload to finish, zero means infinite                                                                    |
+| `ofedDriver.containerResources`                 | []     | `[{"name": "mofed-container", "requests": {}, "limits": {}}]` | Optional [resource requests and limits](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/) for the container                                |
```
adrianchiris, line 432: i would keep that empty. and in description say: resource requirements for individual containers (or something of this sort)
then add separate section about container resources (and link to it from the type )

rinse and repeat for all relevant sections

PR #704, opened 2023-12-11
fix: add label changed predicate for node obj in upgrade controller

ykulazhenkov:
This change is needed because common upgrade lib is using both labels and annotations to store upgrade state. 
This change increase performance of the upgrade controller.

### DIFF
--- a/controllers/upgrade_controller.go
+++ b/controllers/upgrade_controller.go
@@ -222,8 +222,10 @@ func (r *UpgradeReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		return ok
 	}))
 
-	// Watch for spec and annotation changes
-	nodePredicates := builder.WithPredicates(predicate.AnnotationChangedPredicate{})
+	// react only on label and annotation changes
+	nodePredicates := builder.WithPredicates(
+		predicate.Or(predicate.AnnotationChangedPredicate{},
+			predicate.LabelChangedPredicate{}))
 
 	return ctrl.NewControllerManagedBy(mgr).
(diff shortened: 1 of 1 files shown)


controllers/upgrade_controller.go
```diff
@@ -223,7 +223,9 @@ func (r *UpgradeReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	}))
 
 	// Watch for spec and annotation changes
```
rollandf, line 225: Please update comment to include labels

rollandf (approved):
small nit, LGTM

controllers/upgrade_controller.go
```diff
@@ -223,7 +223,9 @@ func (r *UpgradeReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
-	nodePredicates := builder.WithPredicates(predicate.AnnotationChangedPredicate{})
+	nodePredicates := builder.WithPredicates(
```
adrianchiris, line 226: for node predicates, we care about node labels only no ?
when do we want to enqueue based on annotation change ?

the upgrade state is stored in label today.

adrianchiris: from controller side we care about nvidia.com/mofed-driver-upgrade-sate label only no ?

controllers/upgrade_controller.go
ykulazhenkov: fixed

adrianchiris (approved):

e0ne (approved):

PR #705, opened 2023-12-14
feat: enable mofed upgrade flow by default

adrianchiris:
enable mofed upgrade flow by default in helm and ocp deployments.


ykulazhenkov (approved):

adrianchiris:
one thing to note is that in NicClusterPolicy CRD we are keeping it disabled by default.

I believe we must not change it as it may be problematic for existing deployments where CR is managed separately from operator deployment.

rollandf (approved):

PR #706, opened 2023-12-18
feat: Move validation web hook to a subpackage

almaslennikov:
To add validation logic based on manifest rendering,we need to decouple
the validation web hook from the api package to avoid loop dependencies.


ykulazhenkov (approved):

rollandf (approved):

killianmuldoon (commented):
One high level question - if we're moving the webhooks to a new package does it have to be included under `api`?

almaslennikov:
> One high level question - if we're moving the webhooks to a new package does it have to be included under `api`? The only real downside is it adds a bunch of dependencies for anyone who wants to import this API, but I'm not sure if that's a concern.

It's a personal preference. I think it's handy to have a validation next to a type. So far we don't have a use case where network-operator API would be imported into another project

api/v1alpha1/validator/hostdevicenetwork_webhook.go
```diff
@@ -25,50 +27,66 @@ import (
...
 	if skipValidations {
 		nicClusterPolicyLog.Info("skipping CR validation")
 		return nil, nil
 	}
 
+	hostDeviceNetwork, ok := obj.(*v1alpha1.HostDeviceNetwork)
+	if !ok {
+		return nil, errors.New("failed to unmarshal HostDeviceNetwork object to validate")
+	}
+	w.HostDeviceNetwork = *hostDeviceNetwork
```
adrianchiris, line 63: can we have issues with concurrency ?

that is, webhook gets called concurently with two different requests

almaslennikov: I think theoretically we might, just as we probably might have had issues with the previous implementation (the nicclusterpolicy object might change during validation). Let's assess the risks and address them in a separate PR.

WDYT?

adrianchiris: ack, please open a task about it to check.

almaslennikov: @adrianchiris https://github.com/Mellanox/network-operator/issues/714

PR #709, opened 2023-12-19
chore: Update SR-IOV Network Operator chart to consume new cert logic

vasrem:
Today it's cumbersome to enable admission controllers for SRIOV Network Operator because the user needs to create a secret manually. With this PR https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/561, it's possible to generate a self signed certificate so the user doesn't need to do manual steps to enable those admission controllers.

This PR just updates the chart to the latest chart found in the `master` branch of https://github.com/k8snetworkplumbingwg/sriov-network-operator which at the time was https://github.com/k8snetworkplumbingwg/sriov-network-operator/commit/233b99abed2645f626dea43b02179fcec0d661b4.

Another PR will enable proper support for admission controllers.

### DIFF
--- a/deployment/network-operator/charts/sriov-network-operator/templates/certificate.yaml
+++ b/deployment/network-operator/charts/sriov-network-operator/templates/certificate.yaml
@@ -1,0 +1,71 @@
+{{- if .Values.operator.admissionControllers.enabled }}
+{{- if and (.Values.operator.admissionControllers.certificates.certManager.enabled) (.Values.operator.admissionControllers.certificates.certManager.generateSelfSigned) }}
+---
+apiVersion: cert-manager.io/v1
+kind: Certificate
+metadata:
+  name: {{ .Values.operator.admissionControllers.certificates.secretNames.operator }}
+  namespace: {{ .Release.Namespace }}
+spec:
+  dnsNames:
+  - operator-webhook-service.{{ .Release.Namespace }}.svc
+  - operator-webhook-service.{{ .Release.Namespace }}.svc.cluster.local
+  issuerRef:
+    kind: Issuer
+    name: operator-webhook-selfsigned-issuer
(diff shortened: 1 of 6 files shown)


deployment/network-operator/charts/sriov-network-operator/crds/sriovnetwork.openshift.io_sriovoperatorconfigs.yaml
```diff
@@ -65,11 +65,6 @@ spec:
                   provision switchdev-configuration.service and enable OpenvSwitch
                   hw-offload on nodes.
                 type: boolean
-              featureGates:
-                additionalProperties:
-                  type: boolean
-                description: FeatureGates to enable experimental features
-                type: object
```
vasrem, line 72: Drift that never existed in https://github.com/k8snetworkplumbingwg/sriov-network-operator. Not sure why this is here and if it's needed.

rollandf: It is from a PR that is not yet merged upstream:
https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/497/files#diff-d2b2b1b622c09c90d67bef474fd91cd2291a086465d1c75615f15a20a1d0133dR49

It is part of our downstream image, so I believe we need it here.

vasrem: Thanks for the pointers. Reverting this removal.

rollandf:
Overall LGTM, let's wait for dependent PR in SRIOV-NO

adrianchiris (approved):

PR #715, opened 2023-12-20
fix: Race condition in webhooks

### LINKED ISSUE #714: Review and fix possible concurrency issues with the admission webhook
**What would you like to be added**:
The current implementation of the admission webhook might have issues when run concurrently. Need to review the code and make sure it's safe
**Why is this needed**:
If two or more validation requests are issued at the same time for different objs, the behavior may be undefined

killianmuldoon:
Fix an issue where webhooks were carrying a global state which could lead to concurrency bugs.

Fixes #714

### DIFF
--- a/api/v1alpha1/validator/hostdevicenetwork_webhook.go
+++ b/api/v1alpha1/validator/hostdevicenetwork_webhook.go
@@ -27,6 +27,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/validation/field"
 	ctrl "sigs.k8s.io/controller-runtime"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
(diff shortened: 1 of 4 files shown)


api/v1alpha1/validator/hostdevicenetwork_webhook.go
```diff
@@ -35,9 +34,7 @@ import (
 // log is for logging in this package.
 var hostDeviceNetworkLog = logf.Log.WithName("hostdevicenetwork-resource")
 
-type hostDeviceNetworkValidator struct {
-	v1alpha1.HostDeviceNetwork
-}
+type hostDeviceNetworkValidator struct{}
```
killianmuldoon, line 37: Second commit is the fix.

adrianchiris (approved):
Thx @killianmuldoon  !

almaslennikov (approved):

rollandf (approved):

ykulazhenkov (approved):

PR #716, opened 2023-12-20
feat: Validate container resources in the webhook

almaslennikov:
Add webhook validation for:
- Container names used in resource requirements
- Resource requests should not exceed limits
- Requirements should not be zero
- Requirements should not be empty if declared

### DIFF
--- a/api/v1alpha1/validator/nicclusterpolicy_webhook.go
+++ b/api/v1alpha1/validator/nicclusterpolicy_webhook.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"math/big"
 	"os"
+	"path/filepath"
 	"regexp"
 	"strings"
 
@@ -30,15 +31,21 @@ import (
 
 	"github.com/containers/image/v5/docker/reference"
 	"github.com/xeipuuv/gojsonschema"
+	"golang.org/x/exp/slices"
+	v1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	apiresource "k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 
 	"github.com/Mellanox/network-operator/api/v1alpha1"
+	"github.com/Mellanox/network-operator/pkg/config"
+	"github.com/Mellanox/network-operator/pkg/state"
 )
 
 const (
@@ -53,6 +60,8 @@ var nicClusterPolicyLog = logf.Log.WithName("nicclusterpolicy-resource")
 var schemaValidators *schemaValidator
 
 var skipValidations = false
+
+var envConfig = config.FromEnv().State
 
 type nicClusterPolicyValidator struct{}
 
@@ -152,6 +161,7 @@ func (w *nicClusterPolicyValidator) validateNicClusterPolicy(in *v1alpha1.NicClu
 	var allErrs field.ErrorList
 	// Validate Repository
 	allErrs = w.validateRepositories(in, allErrs)
+	allErrs = w.validateContainerResources(in, allErrs)
 	// Validate IBKubernetes
 	ibKubernetes := in.Spec.IBKubernetes
 	if ibKubernetes != nil {
@@ -329,15 +339,15 @@ func (dp *devicePluginSpecWrapper) validateRdmaSharedDevicePlugin(fldPath *field
 		configListInterface := rdmaSharedDevicePluginConfigJSON["configList"]
 		configList, _ := configListInterface.([]interface{})
 		for _, configInterface := range configList {
-			config := configInterface.(map[string]interface{})
-			resourceName := config["resourceName"].(string)
+			dpConfig := configInterface.(map[string]interface{})
+			resourceName := dpConfig["resourceName"].(string)
 			if !isValidRdmaSharedDevicePluginResourceName(resourceName) {
 				allErrs = append(allErrs, field.Invalid(fldPath.Child("Config"),
 					dp.Config, "Invalid Resource name, it must consist of alphanumeric characters, "+
 						"'-', '_' or '.', and must start and end with an alphanumeric character "+
 						"(e.g. 'MyName',  or 'my.name',  or '123-abc') regex used for validation is "+rdmaResourceNameRegex))
 			}
-			resourcePrefix, ok := config["resourcePrefix"]
+			resourcePrefix, ok := dpConfig["resourcePrefix"]
 			if ok {
 				if !isValidFQDN(resourcePrefix.(string)) {
 					allErrs = append(allErrs, field.Invalid(fldPath.Child("Config"), dp.Config,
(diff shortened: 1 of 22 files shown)


almaslennikov:
Once https://github.com/Mellanox/network-operator/pull/715 is merged I will rebase my PR

api/v1alpha1/validator/nicclusterpolicy_webhook.go
```diff
@@ -474,6 +485,159 @@ func validateRepository(repo string, allErrs field.ErrorList, fp *field.Path, ch
...
+
+func validateContainerResourcesIfNotNil(
+	resource *stateRenderData, policy *v1alpha1.NicClusterPolicy, allErrs field.ErrorList,
+	fp *field.Path, fieldName string) field.ErrorList {
+	if resource.spec == nil {
+		return allErrs
+	}
+
+	// Use reflection to check if ContainerResources is nil
+	val := reflect.ValueOf(resource.spec)
```
ykulazhenkov, line 570: Maybe it is better to define a getter for ContainerResources fields for imageSpec struct and use it as an interface in stateRenderData?

```
// ImageSpec Contains container image specifications
type ImageSpec struct {
	// +kubebuilder:validation:Pattern=[a-zA-Z0-9\-]+
	Image string `json:"image"`
	// +kubebuilder:validation:Pattern=[a-zA-Z0-9\.\-\/]+
	Repository string `json:"repository"`
	// +kubebuilder:validation:Pattern=[a-zA-Z0-9\.-]+
	Version string `json:"version"`
	// +optional
	// +kubebuilder:default:={}
	ImagePullSecrets   []string               `json:"imagePullSecrets"`
	ContainerResources []ResourceRequirements `json:"containerResources,omitempty"`
}

func (is *ImageSpec) GetContainerResources() []ResourceRequirements {
	if is == nil {
		return nil
	}
	return is.ContainerResources
}
```
```
type containerResourceProvider interface {
	GetContainerResources() []v1alpha1.ResourceRequirements 
}
```

```
type stateRenderData struct {
	res containerResourceProvider
	newState    newStateFunc
	manifestDir string
}
```
WDYT?

rollandf (approved):
LGTM
See Yuri's comment

ykulazhenkov (approved):
I would prefer to work with/types interfaces explicitly (see my comment), but this is not critical. LGTM

api/v1alpha1/validator/nicclusterpolicy_webhook.go
almaslennikov: Agreed, updated the PR

pkg/state/dummy_provider.go
```diff
@@ -0,0 +1,58 @@
...
+
+func (d *dummyProvider) GetClusterType() clustertype.Type {
+	return clustertype.Kubernetes
+}
+
+func (d *dummyProvider) IsKubernetes() bool {
+	return true
+}
+
+func (d *dummyProvider) IsOpenshift() bool {
```
adrianchiris, line 34: this may be problematic if a state renders different containers in case of openshift ?

e.g DTK

pkg/state/state_cni_plugins.go
```diff
@@ -33,29 +33,31 @@ import (
...
-	"github.com/Mellanox/network-operator/pkg/staticconfig"
 	"github.com/Mellanox/network-operator/pkg/utils"
 )
 
 const stateCNIPluginsName = "state-container-networking-plugins"
 const stateCNIPluginsDescription = "Container Networking CNI Plugins deployed in the cluster"
 
 // NewStateCNIPlugins creates a new state for secondary container networking CNI plugins
-func NewStateCNIPlugins(k8sAPIClient client.Client, scheme *runtime.Scheme, manifestDir string) (State, error) {
+func NewStateCNIPlugins(
```
adrianchiris, line 43: we can embed MainfestRenderer into State and avoid these changes (returning same obj ptr twice)

then relevant code can just accept MainfestRenderer and not whole State obj.

WDYT ?

api/v1alpha1/validator/nicclusterpolicy_webhook.go
```diff
@@ -474,6 +484,168 @@ func validateRepository(repo string, allErrs field.ErrorList, fp *field.Path, ch
 	return allErrs
 }
 
+type containerResourcesProvider interface {
+	GetContainerResources() []v1alpha1.ResourceRequirements
+}
```
adrianchiris, line 489: nit: missing newline

api/v1alpha1/validator/nicclusterpolicy_webhook.go
```diff
@@ -474,6 +484,168 @@ func validateRepository(repo string, allErrs field.ErrorList, fp *field.Path, ch
...
+type newStateFunc func(
+	k8sAPIClient client.Client, scheme *runtime.Scheme, manifestDir string) (state.State, state.ManifestRenderer, error)
+type stateRenderData struct {
```
adrianchiris, line 492: newline

api/v1alpha1/validator/nicclusterpolicy_webhook.go
```diff
@@ -474,6 +484,168 @@ func validateRepository(repo string, allErrs field.ErrorList, fp *field.Path, ch
 	return allErrs
 }
 
+type containerResourcesProvider interface {
```
adrianchiris, line 487: please add docstring for types/funcs

api/v1alpha1/validator/nicclusterpolicy_webhook.go
```diff
@@ -474,6 +484,168 @@ func validateRepository(repo string, allErrs field.ErrorList, fp *field.Path, ch
...
+				filepath.Join(manifestBaseDir, "state-multus-cni"),
+			}
+		}
+		if policy.Spec.SecondaryNetwork.IpamPlugin != nil {
+			states["ipamPlugin"] = stateRenderData{
+				policy.Spec.SecondaryNetwork.IpamPlugin, state.NewStateWhereaboutsCNI,
+				filepath.Join(manifestBaseDir, "state-whereabouts-cni"),
+			}
+		}
+		for stateName, renderData := range states {
```
adrianchiris, line 575: is there a reason to do this in 2 steps ?

that is , why not populate states map with all states then loop over once.

api/v1alpha1/validator/nicclusterpolicy_webhook.go
```diff
@@ -474,6 +484,168 @@ func validateRepository(repo string, allErrs field.ErrorList, fp *field.Path, ch
...
+		}
+	}
+
+	return allErrs
+}
+
+func validateResources(resources map[v1.ResourceName]apiresource.Quantity, allErrs field.ErrorList, fp *field.Path,
+	child, resourceType string, requests ...map[v1.ResourceName]apiresource.Quantity) field.ErrorList {
+	for resourceName, quantity := range resources {
+		if resourceName == v1.ResourceCPU || resourceName == v1.ResourceMemory {
```
adrianchiris, line 627: i believe we want to limit req/limits to just CPU and Memory no ?

if its something else which was provided we fail, WDYT ?

pkg/state/state_cni_plugins.go
almaslennikov: Yeah I think in this case the code will be easier to extend

api/v1alpha1/validator/nicclusterpolicy_webhook.go
almaslennikov: Because we want to provide the proper field path in the error message. Some states are configured under SecondaryNetwork section

api/v1alpha1/validator/nicclusterpolicy_webhook.go
almaslennikov: Let’s do it. We probably don’t care about other resources

api/v1alpha1/validator/nicclusterpolicy_webhook.go
adrianchiris: ack. so its either moving field path into the render data struct or keeping as is.

thx.

pkg/state/dummy_provider.go
adrianchiris: ah, this is just to get current UT working right ? if so then disregard this comment.

pkg/state/state_cni_plugins.go
almaslennikov: Looking into it, we can't really do that as the interface requires NicClusterPolicy object to be passed into the GetManifestObjects func. I think it's not worth adding such a method to the states not related to NicClusterPolicy

api/v1alpha1/validator/nicclusterpolicy_webhook.go
almaslennikov: Done

api/v1alpha1/validator/nicclusterpolicy_webhook.go
almaslennikov: Done

PR #717, opened 2023-12-20
feat: Configure container resources in the helm chart

almaslennikov:
Update README
Add continuity check unit test:
* This test renders all manifests to extract deployable containers' names and parses values.yaml from the helm chart to make sure that all containers are represented in the helm chart under the ResourceRequirements section

### DIFF
--- a/pkg/state/continuity_check_test.go
+++ b/pkg/state/continuity_check_test.go
@@ -1,0 +1,156 @@
+/*
+2023 NVIDIA CORPORATION & AFFILIATES
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package state
+
+import (
+	"bufio"
+	"bytes"
+	"os"
+	"path/filepath"
+	"sort"
+	"strings"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"gopkg.in/yaml.v3"
+
+	mellanoxv1alpha1 "github.com/Mellanox/network-operator/api/v1alpha1"
+	"github.com/Mellanox/network-operator/pkg/config"
+)
+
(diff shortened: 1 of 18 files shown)


almaslennikov:
Build&UT are failing here because the continuity check depends on the changes introduced here: https://github.com/Mellanox/network-operator/pull/716

pkg/state/continuity_check_test.go
```diff
@@ -0,0 +1,150 @@
...
+	var names []string
+
+	switch v := data.(type) {
+	case []interface{}:
+		for _, item := range v {
+			names = append(names, extractContainerNamesFromSubsection(item)...)
+		}
+	case map[string]interface{}:
+		for key, value := range v {
+			if key == "containerResources" {
```
rollandf, line 62: The 'containerResources' are commented out in the values.yaml. I believe it won't find any.

almaslennikov: Thank you! Fixed that by adding a function that uncomments containerResources sections

rollandf (approved):
Overall LGTM.
Need rebase after other PR is merged

pkg/state/continuity_check_test.go
```diff
@@ -0,0 +1,179 @@
...
+
+	state.SetRenderer(render.NewRenderer(files))
+
+	return ParseContainerNames(state, cr, logger)
+}
+
+var _ = Describe("Continuity check", func() {
+
+	Context("Resource requirements", func() {
+		It("Resource requirements from helm chart should cover all deployable containers", func() {
```
adrianchiris, line 125: if i add a new state, but i forget to update helm and forget to update this test will it fail ?

almaslennikov: Yes it will. It compares the list of states rendered from manifests with the list in the helm chart. You don’t need to edit this test when adding a new state

adrianchiris: names from manifests depends on stateMap which is part of the test 
names from chart are extracted from helm chart

not having a state in both (map in test and chart) means that state will be ignored by the test.
so i dont understand how it will work. maybe its the time :)

almaslennikov: Right 😅 of course you need to add the new state to the test as well. If you don’t update both, the test won’t fail.

adrianchiris: This guy returns all nicclusterpolicy states: `newNicClusterPolicyStates`

u can use it to get the expected states then if a state is missing from the map, fail

almaslennikov: Thanks, adapted the test. Not it will fail if you add a new state and not change both the test and helm chart. You still need to edit the NicClusterPolicy object in the test but maybe we can adopt reflection to address that in the future

PR #718, opened 2023-12-20
fix: remove apiserver from go.mod

killianmuldoon:
Remove unneeded dependency from go.mod.

This was accidentally introduced as part of #715


almaslennikov (approved):

ykulazhenkov (approved):

rollandf:
Do you think it is worth to add a CI check to see if 'go mod tidy' generates changes?
Like the ones we have with 'make generate' and 'make manifests'?

killianmuldoon:
> Do you think it is worth to add a CI check to see if 'go mod tidy' generates changes?

Working on it now 😄

PR #719, opened 2023-12-20
fix: Add check for go modules

killianmuldoon:
Add a check in the CI to ensure that go modules are up to date.

This will catch issues like #718 from recurring.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -46,6 +46,16 @@ jobs:
         uses: actions/checkout@v3
       - name: check manifests
         run: make check-manifests
+  diff-modules:
+    name: check go modules
+    runs-on: ubuntu-20.04
+    permissions:
+      contents: read
+    steps:
+      - name: checkout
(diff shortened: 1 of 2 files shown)


killianmuldoon (commented):
Failing on the CI here: https://github.com/Mellanox/network-operator/actions/runs/7275136310/job/19822555302?pr=719

Should pass on current version.

ykulazhenkov (approved):

rollandf (approved):

.github/workflows/build.yml
```diff
@@ -46,6 +46,16 @@ jobs:
         uses: actions/checkout@v3
       - name: check manifests
         run: make check-manifests
+  diff-modules:
```
e0ne, line 49: We need to switch to Ubuntu 22.04 for all workflows. It should be done in a separate PR

e0ne (approved):

adrianchiris (approved):

PR #726, opened 2024-01-01
chore: CHECK CI

rollandf:
Do not merge


rollandf:
Making sure the changes in CI for mofed multiple DS works also with previous DS

PR #727, opened 2024-01-01
feat: Support ForcePrecompiled flag

rollandf:
Mofed images support some OS/Kernel with precompiled
drivers.

In case that a Precompiled image is available in the image registry,
it will be used for MOFED, otherwise the image with sources will be
used.

The user can specify that in case the precompiled image does not exists,
the Ofed State will fail.
This can be done with setting ForcePrecompiled to "true" in "ofedDriver"
spec in NicClusterPolicy.

The default for ForcePrecompiled is false.


rollandf:
Depends on #691

pkg/docadriverimages/doca_drivers.go
```diff
@@ -0,0 +1,141 @@
...
+		if t == tag {
+			return true
+		}
+	}
+	return false
+}
+
+// SetImageSpec sets the Container registry details
+func (p *provider) SetImageSpec(spec *mellanoxv1alpha1.ImageSpec) {
+	if spec == nil {
```
ykulazhenkov, line 88: I think we need to protect this method with mutex as well. 

retrieveTags is running in a background mode and  in theory line 111 can panic if SetImageSpec was called between check on line 103 and loop on line 111

ykulazhenkov (approved):

pkg/docadriverimages/doca_drivers.go
```diff
@@ -0,0 +1,141 @@
...
+		logger.Error(err, "failed to create registry auth from secrets")
+		return
+	}
+	image := fmt.Sprintf("%s/%s", p.docaImageSpec.Repository, p.docaImageSpec.Image)
+	repo, err := name.NewRepository(image)
+	if err != nil {
+		logger.Error(err, "failed to create repo")
+		return
+	}
+	p.tags, err = remote.List(repo, remote.WithAuthFromKeychain(auth))
```
ykulazhenkov, line 136: One more thought. Maybe we should assign tags to p.tags after the err check to make sure that we will not rewrite p.tags with an empty list in case if remote.List call failed with some temp error. WDYT?

killianmuldoon:
@rollandf can you rebase this one?

pkg/docadriverimages/doca_drivers.go
rollandf: I added the mutex.
Since it calls, `retrieveTags` the mutex needs to be explicitly release, and not in a `defer` call

e0ne (approved):

PR #732, opened 2024-01-02
feat: change to doca driver

adrianchiris:
change default driver container to doca-driver
this container provides same functionality but
is part of the doca host software suite


rollandf (approved):

adrianchiris:
going through failed CI logs, i dont think failure related to this PR.

PR #734, opened 2024-01-02
feat: Add support for multi arch images build

e0ne:
Add support for multi arch images build

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,7 +15,9 @@
 # Build the manager binary
 FROM golang:1.20 as builder
 
-ARG TARGETARCH=amd64
+ARG ARCH ?= $(shell go env GOARCH)
+ARG TARGETPLATFORM
+ENV TARGETPLATFORM=${TARGETPLATFORM:-linux/amd64}
 
 WORKDIR /workspace
 # Copy the Go Modules manifests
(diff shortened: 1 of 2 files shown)


ykulazhenkov (approved):

rollandf (approved):

Dockerfile
```diff
@@ -15,7 +15,8 @@
 # Build the manager binary
 FROM golang:1.20 as builder
 
-ARG TARGETARCH=amd64
+ARG ARCH ?= $(shell go env GOARCH)
```
killianmuldoon, line 18: If I understand right multi-arch builds are then done somewhere else - this change is just enabling them. 

Could we make the docker build in the Makefile multi-arch? Or is it better to just have that as a detail of the release process?

almaslennikov (approved):

Dockerfile
e0ne: @killianmuldoon I'll add a note into README to to build such images. I don't want to change make target to not complicate develpment process

rollandf (changes requested):
`make image` does not work now with this Dockerfile, seems to break some CI.

rollandf:
> `make image` does not work now with this Dockerfile, seems to break some CI.

Seems podman related issue

rollandf (approved):

rollandf (approved):

PR #737, opened 2024-01-05
fix: Update both chart and app version when bulding helm chart with Makefile

almaslennikov:
When deploying network-operator helm chart, the templates rely on chart.AppVersion when constructing the image url. To allow to automate chart publishing, consistent AppVersions need to be provided when chart is built.
Add a flag to helm package command in the make chart-build target

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -212,7 +212,11 @@ image-push:
 
 .PHONY: chart-build
 chart-build: $(HELM) ; $(info Building Helm image...)  @ ## Build Helm Chart
-	$Q $(HELM) package deployment/network-operator/ --version $(VERSION)
+	@if [ -z "$(APP_VERSION)" ]; then \
+		echo "APP_VERSION is not set, skipping a part of the command."; \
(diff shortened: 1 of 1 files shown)


rollandf:
I think it may cause an issue with "official" releases as for `appVersion` we need to have a `v` prefix. (https://github.com/Mellanox/network-operator/blob/master/deployment/network-operator/Chart.yaml#L5)

almaslennikov:
> I think it may cause an issue with "official" releases as for `appVersion` we need to have a `v` prefix. (https://github.com/Mellanox/network-operator/blob/master/deployment/network-operator/Chart.yaml#L5)
> 
> Maybe we need it as an additional variable and fix also in jenkins (`cloud-orchestration/cicd/jenkins-confs/-/blob/master/jenkins-files/network-operator/charts-push.jenkinsfile`)

@rollandf You're right. I've added a new `APP_VERSION` parameter to the target for CI cases. Also made those two cases separate so that the absence of `APP_VERSION` doesn't mess up the release build

rollandf (approved):

PR #742, opened 2024-01-08
 chore: Remove unused files and functions

killianmuldoon:
Remove some unused files and functions across the codebase.

AFAICT all of these can be safely removed.


tests/crds/k8s.cni.cncf.io_networkattachmentdefinitions_crd.yaml
```diff
@@ -1,57 +0,0 @@
-# Copyright 2020 NVIDIA
```
rollandf, line 1: This one is need for the unit-tests:
https://github.com/Mellanox/network-operator/blob/master/controllers/suite_test.go#L85
https://github.com/Mellanox/network-operator/blob/master/controllers/suite_test.go#L75

rollandf (approved):

ykulazhenkov (approved):

version/version.go
ykulazhenkov, line 1: I created an issue to add support for displaying version information. https://github.com/Mellanox/network-operator/issues/744

adrianchiris: fields are populated during build (via LDFLAGS).

lt seems we never used this :\, but i believe we do want to use it.

so im for not removing this (except the TODO, because... now we know), and #744 would just print this information 
(thinking logging the version before we start controlller manager plus a version flag)

killianmuldoon:
Going to close this while I do fixups to not put pressure on the CI.

PR #746, opened 2024-01-09
chore: Cleanup unused code and fix build version metadata

killianmuldoon:
Remove some unused files and code from the repo.

Fix the build version implementation to log the version information when the controller starts.


rollandf (approved):

killianmuldoon:
Can see from the operator logs here: https://nvidia-nbu-ci-logs.s3.us-east-2.amazonaws.com/nic_operator_helm-ci/1501/logs/pods_logs/network-operator-helm-ci-56c9cbd579-nggj4-network-operator.log that the version is now being reported on start up.

Dockerfile
```diff
@@ -46,7 +48,7 @@ RUN mkdir crds && \
     cp -r chart/charts/node-feature-discovery/crds /workspace/crds/node-feature-discovery/
 
 # Build
-RUN CGO_ENABLED=0 GOOS=linux GOARCH=${ARCH} go build -a -o manager main.go
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=${ARCH} go build -ldflags="${LDFLAGS}" -a -o manager main.go
```
adrianchiris, line 51: not related to this PR but we should probably copy everything and just use `make build` for this, seems cleaner

Makefile
```diff
@@ -29,7 +29,7 @@ ENVTEST_K8S_VERSION=1.28
 VERSION?=master
 DATE=`date -Iseconds`
 COMMIT?=`git rev-parse --verify HEAD`
-LDFLAGS="-X github.com/Mellanox/network-operator/version.Version=$(VERSION) -X github.com/Mellanox/network-operator/version.Commit=$(COMMIT) -X github.com/Mellanox/network-operator/version.Date=$(DATE)"
+LDFLAGS="-X github.com/Mellanox/network-operator/version.Version=$(BUILD_VERSION) -X github.com/Mellanox/network-operator/version.Commit=$(COMMIT) -X github.com/Mellanox/network-operator/version.Date=$(DATE)"
```
adrianchiris, line 32: when we build from tag, what will we get as build version ? (i think we would want the tag in this case)
when we build from master what will be get as build version (non tagged commit) ?
when we build from dirty tree what will we get as build version ?

if we have indication that it was one of the three above then im ok with the changes

on another note, perhaps we should follow(ish) what was done in nic-feature-discovery: 

[1] https://github.com/Mellanox/nic-feature-discovery/blob/9c97588cf3e377c08668deba366d4114ba2689ca/taskfiles/Version.yaml#L3
[2] https://github.com/Mellanox/nic-feature-discovery/blob/main/pkg/utils/version/version.go#L1

killianmuldoon: This uses `t describe --always --tags --dirty` under the hood. The output at the current HEAD is:

```
## At a non-tagged commit return the most recent tag, the number of steps to the tag and a hash.
bash-4.4$ git describe --always --tags --dirty
v24.1.0-beta.2-4-gd6eb2c5

## At a tagged commit this returns the tag.
bash-4.4$ git tag  v1.9.3
bash-4.4$ git describe --always --tags --dirty
v1.9.3

## With a dirty tree we get a `-dirty` suffix.
bash-4.4$ touch file && git add file
bash-4.4$ git describe --always --tags --dirty
v1.9.3-dirty
```

I think we get all the same information here as we would from the version in nic feature discovery. We collect the commit ID and date separately. `releaseStatus` can be deduced from the state of the tag i.e. whether or not it's semVer.

Would be happy to standardize to the other format if you prefer it though.

adrianchiris: this is perfectly fine. thx for checking it !

adrianchiris (approved):

PR #761, opened 2024-01-16
chore: Bump helm, hadolint, and golangci-lint

killianmuldoon:
Bumps to some tooling in the Makefile. None of them bring major changes.

`helm v3.5.3` -> `helm v3.13.3`
`hadolint v1.23.0` -> `hadolint v2.12.0`
`golangci-lint v1.52.2` -> `golangci-lint v1.55.2`

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -55,7 +55,6 @@ linters:
   disable-all: true
   enable:
     - bodyclose
-    - depguard
     - dogsled
     - dupl
     - errcheck
@@ -95,6 +94,11 @@ issues:
 
   # Excluding configuration per-path, per-linter, per-text and per-source
   exclude-rules:
+    - path: _test\.go
+      linters:
+      - revive
+      - stylecheck
+      text: "should not use dot imports"
     - path: _test\.go
       linters:
         - gomnd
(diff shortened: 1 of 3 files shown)


.golangci.yml
```diff
@@ -55,7 +55,6 @@ linters:
   disable-all: true
   enable:
     - bodyclose
-    - depguard
```
killianmuldoon, line 58: Depguard has become extremely strict to the point of needing an allowlist for all packages if I understand it right. I think we shouldn't bother for the time being, and there's other dependency solutions that could be better if we care in future.

.golangci.yml
```diff
@@ -95,6 +94,11 @@ issues:
 
   # Excluding configuration per-path, per-linter, per-text and per-source
   exclude-rules:
+    - path: _test\.go
+      linters:
+      - revive
+      - stylecheck
+      text: "should not use dot imports"
```
killianmuldoon, line 101: Disabling this on test packages as it catches each `ginkgo` and `gomega` import.

ykulazhenkov (approved):
LGTM

rollandf (approved):

adrianchiris (approved):

PR #767, opened 2024-01-18
fix: Populate coverage in all packages when generating test coverage

vasrem:
This change ensures that all the packages are taken into account when generating test coverage for a particular package e.g. integration tests on controllers end up marking the `Sync()` methods of pkg/state as covered which is not the case without this change.

The drawback of this change is that unit-test coverage is not clearly visible as it used to be.


killianmuldoon (commented):
While you're working with the test packages WDYT about removing the `PKGS` and `TESTPKGS` variables?

I think the `test` target should be run well with `./...` (same as the coverage target).

Adding `coverpkg` is a good change IMO

vasrem:
> While you're working with the test packages WDYT about removing the `PKGS` and `TESTPKGS` variables?
> 
> I think the `test` target should be run well with `./...` (same as the coverage target).

Added that in https://github.com/Mellanox/network-operator/pull/767/commits/137adc95aea34a139d8525e1648baf73c37b0a64. In general tests are running fast (looking at integration tests) and personally I would just run `make test` when developing, but it might be the case that people use those env variables when developing to run unit-tests on a particular package which can make the run significantly faster.

Let's see what others have to say.

killianmuldoon:
> but it might be the case that people use those env variables when developing to run unit-tests on a particular package which can make the run significantly faster.

That makes sense. Could we default it to all packages instead of that TESTPKGS command?

vasrem:
> That makes sense. Could we default it to all packages instead of that `TESTPKGS` command?

Makes total sense. Addressed in https://github.com/Mellanox/network-operator/pull/767/commits/b87a3a069a604611834a468b608ea663e22dae93.

ykulazhenkov (dismissed):

rollandf (dismissed):

Makefile
```diff
@@ -232,7 +231,7 @@ COVERAGE_MODE = count
 
 test-coverage: COVERAGE_DIR := $(CURDIR)/test
 test-coverage: setup-envtest; $(info  running coverage tests...) @ ## Run coverage tests
-	KUBEBUILDER_ASSETS=`$(SETUP_ENVTEST) use --use-env -p path $(ENVTEST_K8S_VERSION)` $(GO) test -covermode=$(COVERAGE_MODE) -coverprofile=network-operator.cover ./...
+	KUBEBUILDER_ASSETS=`$(SETUP_ENVTEST) use --use-env -p path $(ENVTEST_K8S_VERSION)` $(GO) test -covermode=$(COVERAGE_MODE) -coverpkg=./... -coverprofile=network-operator.cover ./...
```
adrianchiris, line 234: can you use TESTPKGS instead of ./... ?

vasrem: Adjusted

adrianchiris (approved):

PR #768, opened 2024-01-18
feat: Add DOCA Telemetry Service deployment

killianmuldoon:
Add DOCA Telemetry Service deployment to the network operator.

Signed-off-by: killianmuldoon <cilliancapi@gmail.com>


killianmuldoon:
PR is a bit of a mess as I rebased it onto a couple of older branches. Not for review in its current state.

manifests/state-doca-telemetry-service/0010-doca-telemetry-service.yaml
```diff
@@ -0,0 +1,140 @@
...
+      volumes:
+        - name: doca-telemetry-service-configmap
+          configMap:
+            name: {{ .ConfigMapName }}
+        - name: pod-device-resources
+          hostPath:
+            path: /var/lib/kubelet/pod-resources
+        - name: cni
+          hostPath:
+            path: /var/run/k8s.cni.cncf.io/devinfo/cni/
```
adrianchiris, line 77: This may not exist which i guess is ok.

if it becomes available later will this behave as expected ?

hack/templates/values/values.template
```diff
@@ -441,6 +441,20 @@ nicFeatureDiscovery:
   #       cpu: "300m"
   #       memory: "150Mi"
 
+docaTelemetryService:
+  deploy: false
+  image: {{ .DOCATelemetryService.Image }}
+  repository: {{ .DOCATelemetryService.Repository }}
+  version: {{ .DOCATelemetryService.Version }}
+  # containerResources:
```
adrianchiris, line 449: nit: add image pull secrets comment like above

manifests/state-doca-telemetry-service/0010-doca-telemetry-service.yaml
```diff
@@ -0,0 +1,140 @@
+---
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: doca-telemetry-service
+  namespace: {{ .RuntimeSpec.Namespace }}
+spec:
+  # for DaemonSet
```
adrianchiris, line 8: nit: remove comment

example/crs/mellanox.com_v1alpha1_nicclusterpolicy_cr_telemetry.yaml
```diff
@@ -0,0 +1,23 @@
...
+
+apiVersion: mellanox.com/v1alpha1
+kind: NicClusterPolicy
+metadata:
+  name: nic-cluster-policy
+spec:
+  docaTelemetryService:
+    image: doca-telemetry
+    repository: nvcr.io/nvidia/doca
+    version: 1.15.5-doca2.5.0-host
```
adrianchiris, line 23: we generate updated examples from here : https://github.com/Mellanox/network-operator/tree/master/hack/templates/crs

so this needs to go into templates/crs

pkg/state/state_doca_telemetry_service.go
```diff
@@ -0,0 +1,196 @@
...
+	CrSpec        *mellanoxv1alpha1.DOCATelemetryServiceSpec
+	ConfigMapName string
+	RuntimeSpec   *dtsRuntimeSpec
+	Tolerations   []v1.Toleration
+	NodeAffinity  *v1.NodeAffinity
+}
+
+// Sync attempt to get the system to match the desired state which State represents.
+// a sync operation must be relatively short and must not block the execution thread.
+func (d docaTelemetryServiceState) Sync(
```
adrianchiris, line 64: any plans to add unit test for it ? like we started to do with different states ?

docs/local-development.md
```diff
@@ -72,7 +72,7 @@ From the root directory of the `network-operator` repository run:
 
 Export a registry to push images to:
 
-`export REGISTRY="localhost:5000`
+`export REGISTRY="localhost:5000"`
```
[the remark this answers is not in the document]
killianmuldoon: I'll fix these in a separate PR - they should be fixed sooner rather than later.

manifests/state-doca-telemetry-service/0010-doca-telemetry-service.yaml
```diff
@@ -0,0 +1,140 @@
...
+        {{- end }}
+        volumeMounts:
+        - name: doca-telemetry-service-configmap
+          mountPath: /configmap
+        - name: pod-device-resources
+          mountPath: /var/lib/kubelet/pod-resources
+        - name: cni
+          mountPath: /var/run/k8s.cni.cncf.io/devinfo/cni
+        env:
+        - name: MY_POD_UID
```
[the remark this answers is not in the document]
killianmuldoon: It's in their yaml. After the converstaions around configuration and who is ultimately responsible for these yamls I only made the changes that were absolutely necessary to get it running.

I can absolutely do another round of trying to clear up the yaml if we think that's the right approach.

pkg/state/state_doca_telemetry_service.go
killianmuldoon: Yeah - it's on the backlog. Right now I'd like to get #814 merged before moving ahead with unit tests here as that could simplify things alot.

pkg/state/state_doca_telemetry_service.go
```diff
@@ -0,0 +1,196 @@
...
+	}
+
+	// Render objects related to the DOCATelemetryService
+	reqLogger.V(consts.LogLevelDebug).Info("Rendering objects", "data:", renderData)
+	renderedObjects, err := d.renderer.RenderObjects(&render.TemplatingData{Data: renderData})
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to render objects")
+	}
+
+	// Remove the configMap from the deployment if an explicit config is set.
```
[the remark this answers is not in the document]
killianmuldoon: We can - is it preferable? 

My preference is to do any logic that's unique to this state in go code because debugging templates is annoying - but I can move this logic to the template if we want.

hack/templates/values/values.template
killianmuldoon: Can do - but the helm part of this isn't working right now. I need to remove helm changes in this PR and add it as a follow-up.

example/crs/mellanox.com_v1alpha1_nicclusterpolicy_cr_telemetry.yaml
killianmuldoon: I don't think we really need this template though - mostly kept it around for dev purposes. Will just remove.

pkg/state/state_doca_telemetry_service.go
killianmuldoon: Happy to do it as part of this PR - but would like to know if I can do them in the style from 814 first.

manifests/state-doca-telemetry-service/0010-doca-telemetry-service.yaml
killianmuldoon: I'm really not sure - the DTS container will fail silently in most cases where things don't work at startup - this might be one of those cases.

killianmuldoon: I'll follow up and try to figure out if this prevents the pod devices collector from running.

pkg/state/state_doca_telemetry_service.go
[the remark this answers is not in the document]
rollandf: Better to do it in template.
We have already existing logic that will remove the stale objects based on the state label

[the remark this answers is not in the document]
killianmuldoon: That logic is being used here - this code doesn't delete the object - it just removes it from the list of desired objects.

I can definitely move this to the template file though

hack/templates/values/values.template
killianmuldoon: It should be done now - this version of the PR has helm support.

manifests/state-doca-telemetry-service/0010-doca-telemetry-service.yaml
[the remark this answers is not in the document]
killianmuldoon: I'm going to push this beyond the scope of this PR - will tackle another round of review of the yaml and use cases in a follow-up.

manifests/state-doca-telemetry-service/0010-doca-telemetry-service.yaml
killianmuldoon: I'm going to push this beyond the scope of this PR - will tackle another round of review of the yaml and use cases in a follow-up.

rollandf (approved):
LGTM

manifests/state-doca-telemetry-service/0010-doca-telemetry-service.yaml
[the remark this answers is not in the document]
killianmuldoon: This is being used by DTS in a metric. AFAICT it's not correctly implemented as it doesn't make sense. I reported it to them, but they're currently using it so we need to plumb it through like this.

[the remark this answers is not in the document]
adrianchiris: cool, thx for checking this out !

PR #769, opened 2024-01-19
fix: add unit tests for multus cni state

vasrem:
This PR adds additional unit tests for the Multus CNI state. 

The `Sync()` function will be tested using envtest on the controller pkg. This will happen in a later PR.

In addition to that, this PR fixes a bug that was discovered while writing tests where default resources of Multus CNI wouldn't render correctly even when ContainerResources were not specified in the NicClusterPolicy. Additional PR will come to fix the behavior in other states that might have the same issue.

### DIFF
--- a/pkg/state/state_multus_cni_test.go
+++ b/pkg/state/state_multus_cni_test.go
@@ -1,0 +1,398 @@
+/*
+2024 NVIDIA CORPORATION & AFFILIATES
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package state
+
+import (
+	"context"
+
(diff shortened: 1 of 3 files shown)


pkg/state/state_multus_cni_test.go
```diff
@@ -0,0 +1,398 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+
+	mellanoxv1alpha1 "github.com/Mellanox/network-operator/api/v1alpha1"
+	"github.com/Mellanox/network-operator/pkg/config"
+	"github.com/Mellanox/network-operator/pkg/render"
+	"github.com/Mellanox/network-operator/pkg/testing/mocks"
+	"github.com/Mellanox/network-operator/pkg/utils"
+)
+
+var _ = Describe("Multus CNI state", func() {
```
killianmuldoon, line 37: I know the general pattern is to use ginkgo for unit tests in network operator, but is there any reason not to use simpler `go test` tests? 

I find those much easier to run and debug locally than ginkgo, and most of the time we're not really using any of the useful features of ginkgo in these tests.

adrianchiris:
IMO we should:

1. have external pkg for testing multus state (multus_state_test)
2. test multus state public interface (State) methods

that means, you need to use a fake kclient for Sync method then query created/updated objs

rollandf:
We can refactor unit-test in follow up PR
LGTM

rollandf (approved):

rollandf:
@almaslennikov PTAL

almaslennikov (approved):

PR #772, opened 2024-01-19
[WIP] Pr multi arch build

killianmuldoon:
Testing out some ideas for building multi-arch images to help with arm Mac dev. No need to review.


killianmuldoon:
@e0ne this is a lot cleaner now - actually possible to review 😄

Makefile
```diff
@@ -237,16 +242,47 @@ test-coverage: setup-envtest; $(info  running coverage tests...) @ ## Run covera
...
+		--build-arg LDFLAGS=$(LDFLAGS) \
+		--build-arg ARCH="$(ARCH)" \
+		--build-arg GCFLAGS="$(GCFLAGS)" \
+		-t $(CONTROLLER_IMAGE)-$(ARCH):$(VERSION) -f $(DOCKERFILE)  $(CURDIR) $(IMAGE_BUILD_OPTS)
+
+image-build-%:
+	make ARCH=$* image-build
+
+.PHONY: image-build-multiarch
+image-build-multiarch: $(addprefix image-build-,$(BUILD_ARCH))
```
e0ne, line 273: this target builds two images `network-operator-amd64:master` and `network-operator-arm64:master` instead of one multi-arch image like `docker buildx` does.  That's not desirable pattern because we can't deliver single helm chart and OCP bundle for different platforms

e0ne (changes requested):

Makefile
killianmuldoon: It's the target `image-push-multiarch` that creates a single tag for multiarch platforms. `image-build-multiarch` + `image-push-multiarch` should produce something identical to what `buildx` produces. We can also put both into the same target, or at least put the `manifest create` part into this target so that it's created locally.

e0ne (approved):

e0ne (approved):

PR #774, opened 2024-01-21
chore: add test to macvlan controller

rollandf:
Add env-test for macvlan network controller.

### DIFF
--- a/controllers/macvlannetwork_controller_test.go
+++ b/controllers/macvlannetwork_controller_test.go
@@ -1,0 +1,174 @@
+/*
+2024 NVIDIA CORPORATION & AFFILIATES
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


controllers/macvlannetwork_controller_test.go
```diff
@@ -0,0 +1,174 @@
...
+			verifyNADCreated(cr.GetName(), testNetworkNamespace)
+
+			By("Verify network namespace annotation")
+			verifyNetworkNamespaceAnnot(testNetworkNamespace)
+
+			By("Verify MacvlanNetwork Status")
+			verifyMacvlanStatus(testNetworkNamespace)
+
+			By("Update NetworkNamespace")
+			Eventually(func() error {
```
ykulazhenkov, line 80: Will this work?

I think it is required to use Gomega instance for Expect calls inside the Eventually block

```
	Eventually(func(g Gomega) {
		model, err := client.Find(1138)
		g.Expect(err).NotTo(HaveOccurred())
		g.Expect(model.Reticulate()).To(Succeed())
		g.Expect(model.IsReticulated()).To(BeTrue())
		g.Expect(model.Save()).To(Succeed())
	}).Should(Succeed())
```

rollandf: Updated.

ykulazhenkov (approved):
Thx, LGTM

e0ne (approved):

PR #775, opened 2024-01-23
feat: Allow to configure persistent driver inventory for the OFED driver to avoid recompilation between runs

almaslennikov:
Introduce DriverInventoryPath to the OFED spec.

It represents a path to a persistent writable folder where OFED container can store its build artifacts. If the artifacts for the correct version and arch exist in this folder when the OFED container is started, they will be reused without driver being recompiled.

### DIFF
--- a/docs/mofed-container-env-vars.md
+++ b/docs/mofed-container-env-vars.md
@@ -23,6 +23,7 @@ In most deployments its expected to not require setting those.
 | UNLOAD_STORAGE_MODULES |N|`"false"`| unload host storage modules prior to loading mofed modules  |
 | ENABLE_NFSRDMA |N|`"false"`| enable loading of nfs relates storage modules from mofed container|
 | RESTORE_DRIVER_ON_POD_TERMINATION |N|`"true"`| restore host drivers when container is gracefully stopped |
+| NVIDIA_NIC_DRIVERS_INVENTORY_PATH | N | `"/mnt/drivers-inventory"` | enable use of a persistent directory to store drivers' build artifacts to avoid recompilation between runs. Keep the default value or set to "" to disable. |
 
 In addition, the user can specify essentially any environment variables to be exposed to the MOFED container such as
 the standard `"HTTP_PROXY"`, `"HTTPS_PROXY"`, `"NO_PROXY"`
--- a/manifests/state-ofed-driver/0050_ofed-driver-ds.yaml
+++ b/manifests/state-ofed-driver/0050_ofed-driver-ds.yaml
@@ -112,6 +112,8 @@ spec:
               readOnly: true
             - name: host-lib-modules
               mountPath: /host/lib/modules
+            - name: drivers-inventory
+              mountPath: /mnt/drivers-inventory
             {{- if.AdditionalVolumeMounts.VolumeMounts }}
             {{- range .AdditionalVolumeMounts.VolumeMounts }}
             - name: {{ .Name }}
@@ -219,6 +221,10 @@ spec:
         - name: host-lib-modules
           hostPath:
             path: /lib/modules
+        - name: drivers-inventory
+          hostPath:
+            path: /var/opt/mofed-container/inventory
+            type: DirectoryOrCreate
         {{- range .AdditionalVolumeMounts.Volumes }}
         - name: {{ .Name }}
           configMap:
--- a/pkg/state/state_ofed.go
+++ b/pkg/state/state_ofed.go
@@ -80,9 +80,12 @@ const (
 
 // names of environment variables which used for OFED proxy configuration
 const (
-	envVarNameHTTPProxy  = "HTTP_PROXY"
-	envVarNameHTTPSProxy = "HTTPS_PROXY"
-	envVarNameNoProxy    = "NO_PROXY"
+	envVarNameHTTPProxy         = "HTTP_PROXY"
+	envVarNameHTTPSProxy        = "HTTPS_PROXY"
+	envVarNameNoProxy           = "NO_PROXY"
+	envVarCreateIfNamesUdev     = "CREATE_IFNAMES_UDEV"
+	envVarDriversInventoryPath  = "NVIDIA_NIC_DRIVERS_INVENTORY_PATH"
+	defaultDriversInventoryPath = "/mnt/drivers-inventory"
 )
 
 // CertConfigPathMap indicates standard OS specific paths for ssl keys/certificates.
@@ -647,13 +650,18 @@ func (s *stateOFED) setEnvFromClusterWideProxy(cr *mellanoxv1alpha1.NicClusterPo
 // mergeWithDefaultEnvs returns env variables provided in currentEnvs merged with default
 // env variables for MOFED container.
 func (s *stateOFED) mergeWithDefaultEnvs(currentEnvs []v1.EnvVar) []v1.EnvVar {
-	if envVarsWithGet(currentEnvs).Get("CREATE_IFNAMES_UDEV") != nil {
-		// already exists dont overwrite
-		return currentEnvs
-	}
+	envs := currentEnvs
 
 	// CREATE_IFNAMES_UDEV: should be set to true if not provided.
-	return append(currentEnvs, v1.EnvVar{Name: "CREATE_IFNAMES_UDEV", Value: "true"})
+	if envVarsWithGet(currentEnvs).Get(envVarCreateIfNamesUdev) == nil {
+		envs = append(envs, v1.EnvVar{Name: envVarCreateIfNamesUdev, Value: "true"})
+	}
+	// NVIDIA_NIC_DRIVERS_INVENTORY_PATH: should be set to true if not provided.
+	if envVarsWithGet(currentEnvs).Get(envVarDriversInventoryPath) == nil {
+		envs = append(envs, v1.EnvVar{Name: envVarDriversInventoryPath, Value: defaultDriversInventoryPath})
+	}
+
+	return envs
 }
 
 // envVarsWithGet is a wrapper type for []EnvVar to extend with additional functionality
--- a/pkg/state/state_ofed_test.go
+++ b/pkg/state/state_ofed_test.go
@@ -214,16 +214,31 @@ var _ = Describe("MOFED state test", func() {
 			Expect(mergedEnvs).To(BeEquivalentTo(expectedEnvs))
 		},
 		Entry("add defaults when no env vars",
-			[]v1.EnvVar{}, []v1.EnvVar{{Name: "CREATE_IFNAMES_UDEV", Value: "true"}}),
+			[]v1.EnvVar{}, []v1.EnvVar{
+				{Name: envVarCreateIfNamesUdev, Value: "true"},
+				{Name: envVarDriversInventoryPath, Value: defaultDriversInventoryPath}}),
 		Entry("add defaults when env vars provided",
 			[]v1.EnvVar{{Name: "Foo", Value: "Bar"}},
-			[]v1.EnvVar{{Name: "Foo", Value: "Bar"}, {Name: "CREATE_IFNAMES_UDEV", Value: "true"}}),
+			[]v1.EnvVar{
+				{Name: "Foo", Value: "Bar"},
+				{Name: envVarCreateIfNamesUdev, Value: "true"},
+				{Name: envVarDriversInventoryPath, Value: defaultDriversInventoryPath}}),
 		Entry("override defaults by user",
-			[]v1.EnvVar{{Name: "CREATE_IFNAMES_UDEV", Value: "false"}},
-			[]v1.EnvVar{{Name: "CREATE_IFNAMES_UDEV", Value: "false"}}),
+			[]v1.EnvVar{
+				{Name: envVarCreateIfNamesUdev, Value: "false"},
+				{Name: envVarDriversInventoryPath, Value: ""}},
+			[]v1.EnvVar{
+				{Name: envVarCreateIfNamesUdev, Value: "false"},
+				{Name: envVarDriversInventoryPath, Value: ""}}),
 		Entry("override defaults by user with additional env vars",
-			[]v1.EnvVar{{Name: "Foo", Value: "Bar"}, {Name: "CREATE_IFNAMES_UDEV", Value: "false"}},
-			[]v1.EnvVar{{Name: "Foo", Value: "Bar"}, {Name: "CREATE_IFNAMES_UDEV", Value: "false"}}),
+			[]v1.EnvVar{
+				{Name: "Foo", Value: "Bar"},
+				{Name: envVarCreateIfNamesUdev, Value: "false"},
+				{Name: envVarDriversInventoryPath, Value: ""}},
+			[]v1.EnvVar{
+				{Name: "Foo", Value: "Bar"},
+				{Name: envVarCreateIfNamesUdev, Value: "false"},
+				{Name: envVarDriversInventoryPath, Value: ""}}),
 	)
 
 	Context("Render Manifests", func() {


deployment/network-operator/values.yaml
```diff
@@ -280,6 +280,8 @@ ofedDriver:
       # podSelector: "app=myapp"
       # specify the length of time in seconds to wait before giving up for workload to finish, zero means infinite
       # timeoutSeconds: 300
+  # OFED container will store its build artifacts between runs to avoid driver recompilation on container rerun
+  driverInventoryPath: "/opt/mofed-inventory"
```
almaslennikov, line 284: Do we want to enable this feature in the helm chart while the corresponding OFED changes are still in progress?
Or we can postpone merging this PR until before April release

api/v1alpha1/nicclusterpolicy_types.go
```diff
@@ -106,6 +106,11 @@ type OFEDDriverSpec struct {
 	// +kubebuilder:default:=300
 	// +kubebuilder:validation:Minimum:=0
 	TerminationGracePeriodSeconds int64 `json:"terminationGracePeriodSeconds,omitempty"`
+	// Path to a persistent writable folder where OFED container can store its build artifacts.
+	// If the artifacts for the correct version and arch exist in this folder when the OFED container is started,
+	// they will be reused without driver being recompiled.
+	// +optional
+	DriverInventoryPath string `json:"driverInventoryPath,omitempty"`
```
ykulazhenkov, line 113: nit: OFED env variable contains "drivers". Do we want to rename the CR option to DriversInventoryPath WDYT?

deployment/network-operator/values.yaml
ykulazhenkov: For me it looks quite safe to merge this change now.

ykulazhenkov (approved):
LGTM, one small nit

deployment/network-operator/values.yaml
rollandf: I think we should wait at least for after having a branch for January

manifests/state-ofed-driver/0050_ofed-driver-ds.yaml
```diff
@@ -230,6 +238,11 @@ spec:
         - name: shared-doca-driver-toolkit
           emptyDir: {}
         {{- end }}
+        {{- if .CrSpec.DriverInventoryPath }}
```
adrianchiris, line 241: do we have an explicit requirement to make this configurable ?

almaslennikov: No, we can also add this a true/false switch

adrianchiris: do you (or others) see issue with always defining the volume and its volume mount ?

then we can have the env variable set by default (there is some code in state that handles default env vars)
if we need to disable for some reason, user can override this env var with empty string

manifests/state-ofed-driver/0050_ofed-driver-ds.yaml
```diff
@@ -124,6 +128,10 @@ spec:
             - name: shared-doca-driver-toolkit
               mountPath: /mnt/shared-doca-driver-toolkit
             {{- end}}
+            {{- if .CrSpec.DriverInventoryPath }}
```
rollandf, line 131: Is it needed for the DTK container?

manifests/state-ofed-driver/0050_ofed-driver-ds.yaml
almaslennikov: Sounds good to me. Should be documented, otherwise I see no issues with this approach.

@rollandf @ykulazhenkov @e0ne @killianmuldoon @vasrem what do you think?

rollandf: I agree that we should hide things from user where possible.
Also, need to verify on OpenShift where this dir should be created and if it will be kept after reboot. (Maybe there is no issue)

ykulazhenkov: > then we can have the env variable set by default (there is some code in state that handles default env vars)
if we need to disable for some reason, user can override this env var with empty string

Set by default in the code not in helm values, right? +1 to handle this in the code.

adrianchiris: > Set by default in the code not in helm values, right? +1 to handle this in the code.
yes

> Also, need to verify on OpenShift where this dir should be created
we can ask RH folks in the weekly meeting. in my openshift single node cluster with coreOS, /opt is mapped to /var/opt 
this is the mount im seeing (not tmpfs):
/dev/sda4 /var xfs rw,seclabel,relatime,attr2,inode64,logbufs=8,logbsize=32k,prjquota 0 0

almaslennikov: So far I've deduced that /var should be writable. We could create this dir in /var/opt if the RH folks confirm it's indeed writable with privileged SCC

manifests/state-ofed-driver/0050_ofed-driver-ds.yaml
almaslennikov: Per discussion with Vlad, the main ofed container gathers the build artifacts from the DTK container. The current idea is to let the main container handle the inventory

manifests/state-ofed-driver/0050_ofed-driver-ds.yaml
almaslennikov: Updated the PR and used `/var/opt/ofed-inventory` dir

manifests/state-ofed-driver/0050_ofed-driver-ds.yaml
rollandf: So no need for that mount on the DTK container

docs/mofed-container-env-vars.md
```diff
@@ -23,6 +23,7 @@ In most deployments its expected to not require setting those.
 | UNLOAD_STORAGE_MODULES |N|`"false"`| unload host storage modules prior to loading mofed modules  |
 | ENABLE_NFSRDMA |N|`"false"`| enable loading of nfs relates storage modules from mofed container|
 | RESTORE_DRIVER_ON_POD_TERMINATION |N|`"true"`| restore host drivers when container is gracefully stopped |
+| USE_DRIVERS_INVENTORY |N|`"true"`| use a persistent directory to store drivers' build artifacts to avoid recompilation between runs |
```
rollandf, line 26: So, this is an Operator variable, not a MODED container one.
I don't think we have a doc for Operator variables, maybe we should

manifests/state-ofed-driver/0050_ofed-driver-ds.yaml
almaslennikov: Yep

docs/mofed-container-env-vars.md
almaslennikov: `USE_DTK` env var can be controlled by a helm chart and is documented as part of it. Do we want to do the same with this feature?

rollandf: Yes, I think we could align the same as `USE_DTK`

shivamerla:
@almaslennikov thanks for letting me know, this change looks good to me. I cannot think of OSes where `/var/opt` is not writable (including Google COS and Azure CBL-Mariner), so hard coding this seems fine to me. Making this configurable also open up unnecessary security concerns.

e0ne (commented):
LGTM once CI pass

killianmuldoon (approved):

e0ne (approved):

manifests/state-ofed-driver/0050_ofed-driver-ds.yaml
```diff
@@ -232,6 +240,12 @@ spec:
         - name: shared-doca-driver-toolkit
           emptyDir: {}
         {{- end }}
+        {{- if .RuntimeSpec.UseDriversInventory }}
+        - name: drivers-inventory
+          hostPath:
+            path: /var/opt/ofed-inventory
```
adrianchiris, line 246: i would prefer this to not be flat as this might expand later to other dirs

WDYT about the following path :  `/var/opt/mofed-container/cache`

pkg/config/config.go
```diff
@@ -56,6 +56,10 @@ type OFEDStateConfig struct {
 	InitContainerImage string `env:"OFED_INIT_CONTAINER_IMAGE"`
 	// UseDTK Enable use of Driver ToolKit side car to compile OFED drivers (Relevant for OpenShift only)
 	UseDTK bool `env:"USE_DTK" envDefault:"true"`
+	// UseDriversInventory enables use of the special persistent directory on the host filesystem to store OFED build
+	// artifacts. If the artifacts for the correct version and arch exist in this folder
+	// when the OFED container is started, they will be reused without driver being recompiled.
+	UseDriversInventory bool `env:"USE_DRIVERS_INVENTORY" envDefault:"true"`
```
adrianchiris, line 62: I dont think we need this to be derived from env var of the operator.

what i had in mind: in `state_ofed.go` just extend `mergeWithDefaultEnvs()` to always set this.

manifests/state-ofed-driver/0050_ofed-driver-ds.yaml
```diff
@@ -124,6 +128,10 @@ spec:
             - name: shared-doca-driver-toolkit
               mountPath: /mnt/shared-doca-driver-toolkit
             {{- end}}
+            {{- if .RuntimeSpec.UseDriversInventory }}
+            - name: drivers-inventory
```
adrianchiris, line 132: dont we want to always have these mounts ?

adrianchiris: we could also set it if  driver inventory env var is set to true  WDYT ?

pkg/config/config.go
almaslennikov: But it's worth noting that this `USE_DRIVERS_INVENTORY` env var will not translate directly to the env var mofed container is expecting. We still need to transform it into:
```
name: NVIDIA_NIC_DRIVERS_INVENTORY_PATH
value: "/mnt/drivers-inventory"
```

docs/mofed-container-env-vars.md
```diff
@@ -23,6 +23,7 @@ In most deployments its expected to not require setting those.
 | UNLOAD_STORAGE_MODULES |N|`"false"`| unload host storage modules prior to loading mofed modules  |
 | ENABLE_NFSRDMA |N|`"false"`| enable loading of nfs relates storage modules from mofed container|
 | RESTORE_DRIVER_ON_POD_TERMINATION |N|`"true"`| restore host drivers when container is gracefully stopped |
+| USE_DRIVERS_INVENTORY | N | `"true"` | enable use of a persistent directory to store drivers' build artifacts to avoid recompilation between runs |
```
adrianchiris, line 26: why wouldnt we want to direclty document: NVIDIA_NIC_DRIVERS_INVENTORY_PATH ?
if empty we can say that that build artifacts will not be stored and recompile will always happen

lets hear others opinions on this

manifests/state-ofed-driver/0050_ofed-driver-ds.yaml
adrianchiris: im ok with having this defined by default.

if we want this to be defined by the operator, perhaps path can be provided via template value ? or inferred from the relevant env var if it exists ?

lets see what other think.

pkg/state/state_ofed.go
```diff
@@ -647,13 +653,18 @@ func (s *stateOFED) setEnvFromClusterWideProxy(cr *mellanoxv1alpha1.NicClusterPo
...
-	}
+	envs := currentEnvs
 
 	// CREATE_IFNAMES_UDEV: should be set to true if not provided.
-	return append(currentEnvs, v1.EnvVar{Name: "CREATE_IFNAMES_UDEV", Value: "true"})
+	if envVarsWithGet(currentEnvs).Get(envVarCreateIfNamesUdev) == nil {
+		envs = append(envs, v1.EnvVar{Name: envVarCreateIfNamesUdev, Value: "true"})
+	}
+	// USE_DRIVERS_INVENTORY: should be set to true if not provided.
+	if envVarsWithGet(currentEnvs).Get(envVarUseDriversInventory) == nil {
```
adrianchiris, line 663: hmmm,
i thought we will add : NVIDIA_NIC_DRIVERS_INVENTORY_PATH here with value pointing to`/mnt/drivers-inventory`

if provided with empty value ("") then it will override the default value above and disable the feature (which we may want if things go south with this feature.

pkg/state/state_ofed.go
```diff
@@ -445,6 +448,8 @@ func (s *stateOFED) GetManifestObjects(
 		return nil, err
 	}
 
+	useDriversInventory := envVarsWithGet(cr.Spec.OFEDDriver.Env).Get(envVarUseDriversInventory).Value == "true"
```
adrianchiris, line 451: thats not what i had in mind tbh.

docs/mofed-container-env-vars.md
adrianchiris: the rest of the values here are mapped directly to the container right ?

almaslennikov: Yes, I've changed it to use directly the `NVIDIA_NIC_DRIVERS_INVENTORY_PATH` env var

pkg/state/state_ofed.go
almaslennikov: I see now! :) Updated the code to use the `NVIDIA_NIC_DRIVERS_INVENTORY_PATH` env var directly

manifests/state-ofed-driver/0050_ofed-driver-ds.yaml
```diff
@@ -112,6 +112,8 @@ spec:
               readOnly: true
             - name: host-lib-modules
               mountPath: /host/lib/modules
+            - name: drivers-inventory
```
e0ne, line 115: Should we mount this only if `NVIDIA_NIC_DRIVERS_INVENTORY_PATH` is set?

adrianchiris (dismissed):
LGTM!

killianmuldoon:
I think the most recent failures on the CI aren't flakes - there's something in the change that causes the OFED container to not become ready.

docs/mofed-container-env-vars.md
```diff
@@ -23,6 +23,7 @@ In most deployments its expected to not require setting those.
 | UNLOAD_STORAGE_MODULES |N|`"false"`| unload host storage modules prior to loading mofed modules  |
 | ENABLE_NFSRDMA |N|`"false"`| enable loading of nfs relates storage modules from mofed container|
 | RESTORE_DRIVER_ON_POD_TERMINATION |N|`"true"`| restore host drivers when container is gracefully stopped |
+| NVIDIA_NIC_DRIVERS_INVENTORY_PATH | N | `"/mnt/drivers-inventory"` | enable use of a persistent directory to store drivers' build artifacts to avoid recompilation between runs. Set to "" to disable |
```
rollandf, line 26: What will happen if user set this to `"/foo"` for example?

From the code it will always get the default value, no?

almaslennikov: > From the code it will always get the default value, no?

You are right and I'd say it's desired behavior. The only control option we'd like to expose is to disable the feature

pkg/state/state_ofed.go
```diff
@@ -647,13 +650,18 @@ func (s *stateOFED) setEnvFromClusterWideProxy(cr *mellanoxv1alpha1.NicClusterPo
...
-		return currentEnvs
-	}
+	envs := currentEnvs
 
 	// CREATE_IFNAMES_UDEV: should be set to true if not provided.
-	return append(currentEnvs, v1.EnvVar{Name: "CREATE_IFNAMES_UDEV", Value: "true"})
+	if envVarsWithGet(currentEnvs).Get(envVarCreateIfNamesUdev) == nil {
+		envs = append(envs, v1.EnvVar{Name: envVarCreateIfNamesUdev, Value: "true"})
+	}
+	// NVIDIA_NIC_DRIVERS_INVENTORY_PATH: should be set to true if not provided.
```
adrianchiris, line 659: nit: comment is wrong.

docs/mofed-container-env-vars.md
adrianchiris: lets clarify documentation that changing the path has no effect when used with the operator

adrianchiris (approved):
LGTM.

rollandf (approved):

PR #776, opened 2024-01-23
chore: Bump commit check github action

killianmuldoon:
Bump conventional commit github action. This action was giving a security warning because of outdated dependencies.

Release: https://github.com/agenthunt/conventional-commit-checker-action/releases/tag/v2.0.0

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -1,0 +1,29 @@
+version: 2
+updates:
+  # GitHub Actions
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+      day: "monday"
+    commit-message:
+      prefix: "chore:"
+
+  - package-ecosystem: "gomod"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+      day: "tuesday"
+    groups:
+      kubernetes:
+        patterns: [ "k8s.io/*" ]
+    ignore:
+      # Ignore k8s-operator-libs, controller-runtime, and Kubernetes major and minor updates. These should be done manually.
+      - dependency-name: "sigs.k8s.io/controller-runtime"
+        update-types: [ "version-update:semver-major", "version-update:semver-minor" ]
+      - dependency-name: "github.com/NVIDIA/k8s-operator-libs"
+        update-types: [ "version-update:semver-major", "version-update:semver-minor" ]
+      - dependency-name: "k8s.io/*"
+        update-types: [ "version-update:semver-major", "version-update:semver-minor" ]
+    commit-message:
+      prefix: "chore:"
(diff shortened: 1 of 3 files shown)


.github/workflows/commit-msg-chk.yaml
```diff
@@ -11,4 +11,4 @@ jobs:
     steps:
       - name: check-for-cc
         id: check-for-cc
-        uses: agenthunt/conventional-commit-checker-action@v1.0.0
+        uses: agenthunt/conventional-commit-checker-action@v2.0.0
```
killianmuldoon, line 14: Note: We could get dependabot to automate these updates - have we considered using dependabot previously?

ykulazhenkov (approved):

.github/workflows/commit-msg-chk.yaml
ykulazhenkov: Interesting. You mean for all PRs? Can you share a link to the doc which describes this?

killianmuldoon: I've added a new commit with a dependabot config

.github/dependabot.yaml
```diff
@@ -0,0 +1,29 @@
...
+      kubernetes:
+        patterns: [ "k8s.io/*" ]
+    ignore:
+      # Ignore k8s-operator-libs, controller-runtime, and Kubernetes major and minor updates. These should be done manually.
+      - dependency-name: "sigs.k8s.io/controller-runtime"
+        update-types: [ "version-update:semver-major", "version-update:semver-minor" ]
+      - dependency-name: "github.com/NVIDIA/k8s-operator-libs"
+        update-types: [ "version-update:semver-major", "version-update:semver-minor" ]
+      - dependency-name: "k8s.io/*"
+        update-types: [ "version-update:semver-major", "version-update:semver-minor" ]
```
killianmuldoon, line 27: These ignores mean we will have to manually update (for major and minor releases) K8s-operator-libs, controller runtime and all K8s dependencies. I think this makes sense as this is a huge part of our dependency tree and we might prefer to take it on manually.

Controller runtime also regularly enough requires some kind of code changes for a minor release so the likelihood is we'd have to do this anyway. And controller runtime has one release per K8s release.

.github/workflows/commit-msg-chk.yaml
ykulazhenkov: Ah, I see. I misunderstood your first comment. I thought that you said that we could use Dependabot to replace the conventional-commit checker somehow.

Thanks!
+ 1 from me to manage updates for dependencies automatically

ykulazhenkov (approved):

adrianchiris (approved):

PR #785, opened 2024-01-25
chore: add cni plugins test

rollandf:
Add unit-test to CNI plugins state.

The test is a different package so that only public APIs are used.

### DIFF
--- a/pkg/state/common_test.go
+++ b/pkg/state/common_test.go
@@ -1,0 +1,67 @@
+/*
+2024 NVIDIA CORPORATION & AFFILIATES
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package state_test
+
+import (
+	"github.com/Mellanox/network-operator/pkg/clustertype"
+	"github.com/Mellanox/network-operator/pkg/nodeinfo"
+	"github.com/Mellanox/network-operator/pkg/state"
+	"github.com/Mellanox/network-operator/pkg/staticconfig"
+)
+
+type testProvider struct {
+	isOpenshift bool
+	cniBinDir   string
+}
+
+func (tp *testProvider) GetClusterType() clustertype.Type {
+	if tp.isOpenshift {
+		return clustertype.Openshift
+	}
+	return clustertype.Kubernetes
+}
+
+func (tp *testProvider) IsKubernetes() bool {
+	return !tp.isOpenshift
+}
+
+func (tp *testProvider) IsOpenshift() bool {
+	return tp.isOpenshift
+}
+
+func (tp *testProvider) GetStaticConfig() staticconfig.StaticConfig {
+	return staticconfig.StaticConfig{CniBinDirectory: tp.cniBinDir}
+}
+
+func (tp *testProvider) GetNodesAttributes(...nodeinfo.Filter) []nodeinfo.NodeAttributes {
+	nodeAttr := make(map[nodeinfo.AttributeType]string)
+	nodeAttr[nodeinfo.AttrTypeCPUArch] = "amd64"
+	nodeAttr[nodeinfo.AttrTypeOSName] = "ubuntu"
+	nodeAttr[nodeinfo.AttrTypeOSVer] = "20.04"
+
+	return []nodeinfo.NodeAttributes{{Attributes: nodeAttr}}
+}
+
+func getTestCatalog() state.InfoCatalog {
+	catalog := state.NewInfoCatalog()
+	tp := &testProvider{isOpenshift: false, cniBinDir: ""}
+	catalog.Add(state.InfoTypeNodeInfo, tp)
+	catalog.Add(state.InfoTypeStaticConfig, tp)
+	catalog.Add(state.InfoTypeClusterType, tp)
+
+	return catalog
+}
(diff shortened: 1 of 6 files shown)


pkg/state/state_cni_plugins.go
```diff
@@ -146,13 +146,17 @@ func (s *stateCNIPlugins) GetManifestObjects(
...
+		return nil, errors.New("clusterInfo provider required")
+	}
 	renderData := &CNIPluginsManifestRenderData{
 		CrSpec:       cr.Spec.SecondaryNetwork.CniPlugins,
 		Tolerations:  cr.Spec.Tolerations,
 		NodeAffinity: cr.Spec.NodeAffinity,
 		RuntimeSpec: &cniRuntimeSpec{
 			runtimeSpec:        runtimeSpec{config.FromEnv().State.NetworkOperatorResourceNamespace},
-			CniBinDirectory:    utils.GetCniBinDirectory(staticConfig, nil),
+			CniBinDirectory:    utils.GetCniBinDirectory(staticConfig, clusterInfo),
```
vasrem, line 159: Was there a bug before that ?

pkg/state/state_cni_plugins_test.go
```diff
@@ -0,0 +1,346 @@
...
+		By("Set spec to nil and Sync")
+		cr.Spec.SecondaryNetwork.CniPlugins = nil
+		status, err = cniPluginsState.Sync(context.Background(), cr, catalog)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(status).To(BeEquivalentTo(state.SyncStateNotReady))
+		By("Verify DaemonSet is deleted")
+		ds = &appsv1.DaemonSet{}
+		err = client.Get(context.Background(), types.NamespacedName{Namespace: namespace, Name: "cni-plugins-ds"}, ds)
+		Expect(errors.IsNotFound(err)).To(BeTrue())
+	})
```
vasrem, line 195: nit: maybe add 2 contexts, one for the rendered manifests and one for the Sync flow without what is rendered.

pkg/state/common_test.go
```diff
@@ -0,0 +1,67 @@
...
+package state_test
+
+import (
+	"github.com/Mellanox/network-operator/pkg/clustertype"
+	"github.com/Mellanox/network-operator/pkg/nodeinfo"
+	"github.com/Mellanox/network-operator/pkg/state"
+	"github.com/Mellanox/network-operator/pkg/staticconfig"
+)
+
+type testProvider struct {
```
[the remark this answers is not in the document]
rollandf: If I want to use the dummy provider, I will need to make it public.
Eventually, if we go with the different package for the test, we can have a single "test" provider that can be configurable.
WDYT?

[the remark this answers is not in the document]
vasrem: > If I want to use the dummy provider, I will need to make it public.

True. I see that dummy provider is only used in tests, so we need to make it public to be used by tests that are in the `state` and `state_test` pkgs.

> Eventually, if we go with the different package for the test, we can have a single "test" provider that can be configurable.

Right.

---

Overall, it's a decision if we want to have a duplicate test provider until we move all the tests in a separate pkg vs making the existing provider public in the interim.

I think both are fine, just wanted to point out this dummy provider :)

pkg/state/state_cni_plugins.go
rollandf: Not really, as CNI plugins are not supposed to be used in OpenShift.
It would choose CNI default path instead of OpenShift one.

pkg/state/state_cni_plugins_test.go
rollandf: Not sure which test should go where, all of them are using `Sync`

vasrem: Basically, what I was thinking is:
* Add context for tests checking the flow of Sync (i.e. error handling, correct garbage collection etc)
  * https://github.com/Mellanox/network-operator/blob/30d9595d02cd033e57543f9f20bae02f4d03023f/pkg/state/state_cni_plugins.go#L93-L95
  * https://github.com/Mellanox/network-operator/blob/30d9595d02cd033e57543f9f20bae02f4d03023f/pkg/state/state_cni_plugins.go#L98-L100
  * https://github.com/Mellanox/network-operator/blob/30d9595d02cd033e57543f9f20bae02f4d03023f/pkg/state/state_cni_plugins.go#L112-L114
  * etc
* Add context for tests checking the manifests rendered by Sync
  * Did it render the image pull secret correctly?
  * Did it render container resources correctly?
  * etc

pkg/state/common_test.go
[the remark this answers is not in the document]
killianmuldoon: We could make the dummy provider Public under /`internal/dummy` or similar

pkg/state/state_cni_plugins_test.go
rollandf: Done

pkg/state/common_test.go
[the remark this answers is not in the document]
rollandf: I would prefer to keep both until we move all the tests.

vasrem (approved):

pkg/state/common_test.go
```diff
@@ -0,0 +1,67 @@
...
+	nodeAttr[nodeinfo.AttrTypeOSName] = "ubuntu"
+	nodeAttr[nodeinfo.AttrTypeOSVer] = "20.04"
+
+	return []nodeinfo.NodeAttributes{{Attributes: nodeAttr}}
+}
+
+func getTestCatalog() state.InfoCatalog {
+	catalog := state.NewInfoCatalog()
+	tp := &testProvider{isOpenshift: false, cniBinDir: ""}
+	catalog.Add(state.InfoTypeNodeInfo, tp)
```
adrianchiris, line 62: this works ofc, but any reason why not to use mocks ?

adrianchiris (approved):

adrianchiris (approved):

pkg/state/common_test.go
rollandf: Open an issue to change to mock in a follow up:
https://github.com/Mellanox/network-operator/issues/791

PR #792, opened 2024-02-05
chore: use mocks for providers

### LINKED ISSUE #791: Use mock for catalog providers in unit tests
Instead of using test/dummy provider, generate mocks and use in tests.

rollandf:
Fix #791


.mockery.yaml
```diff
@@ -0,0 +1,4 @@
+inpackage: False
+testonly: False
+with-expecter: True
+keeptree: True
```
adrianchiris, line 4: nit: newline

adrianchiris (dismissed):

.mockery.yaml
rollandf: done

vasrem (approved):

rollandf:
Copy Right Validation is checking also auto-generated files...
Maybe we should update this job to filter those somehow.

e0ne (approved):

adrianchiris (approved):

PR #794, opened 2024-02-08
chore: Update MOFED to v24.01-0.3.3.1

e0ne:
Update MOFED to v24.01-0.3.3.1


killianmuldoon (commented):
/lgtm

I killed the kind CI as it's still being repaired - not sure if we want to wait for it to be fixed before merging.

ykulazhenkov (approved):

PR #825, opened 2024-02-20
fix: hostdevicenetwork IPAM

rollandf:
An empty IPAM string in the HostDeviceNetwork
Spec was rendered without brackets, making the
NAD config a not valid json.

### DIFF
--- a/manifests/state-hostdevice-network/0010-hostdevice-net-cr.yml
+++ b/manifests/state-hostdevice-network/0010-hostdevice-net-cr.yml
@@ -10,5 +10,5 @@ spec:
   "cniVersion":"0.3.1",
   "name":"{{.HostDeviceNetworkName}}",
   "type":"host-device",
-  "ipam": {{.CrSpec.IPAM}}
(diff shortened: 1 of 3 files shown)


manifests/state-hostdevice-network/0010-hostdevice-net-cr.yml
```diff
@@ -10,5 +10,5 @@ spec:
   "cniVersion":"0.3.1",
   "name":"{{.HostDeviceNetworkName}}",
   "type":"host-device",
-  "ipam": {{.CrSpec.IPAM}}
+  {{.Ipam}}
```
ykulazhenkov, line 13: We can use `"ipam":{{or .CrSpec.IPAM "{}"}}` here and avoid extra logic in the state. WDYT?

rollandf: Done

ykulazhenkov (approved):
LGTM

e0ne (approved):

PR #827, opened 2024-02-21
chore: Bump go to v1.21

killianmuldoon:
chore: Bump go to v1.21

v1.21 is in use in the gomod and toolchain of some of our dependencies - at least controller runtime. Starting from Go 1.21 the go mod will be automatically updated to ensure we're working with the right toolchain.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,7 +11,7 @@ jobs:
       - name: Set up Go 1.x
         uses: actions/setup-go@v5
         with:
-          go-version: '1.20'
+          go-version: '1.21'
       - name: checkout
         uses: actions/checkout@v4
       - name: lint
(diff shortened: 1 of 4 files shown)


ykulazhenkov (dismissed):

ykulazhenkov:
I C that k8s repo contains 1.21 in the go.mod, but doesn't include the toolchain line. Do we want to keep it?

Also we need to upgrade tools (see tests that failed) to be compatible with the toolchain setting.

killianmuldoon:
> I C that k8s repo contains 1.21 in the go.mod, but doesn't include the toolchain line. Do we want to keep it? Also we need to upgrade tools (see tests that failed) to be compatible with the toolchain setting.

I think that as long as any dependency has a toolchain line we should stay at least equal to it. If we don't specify the toolchain it's implicitly v1.21.0 in this case anyway - so it's better to set it explicitly IMO.

killianmuldoon:
Let me take a look at the failures and try to fix them

e0ne (approved):
LGTM

PR #831, opened 2024-02-25

maze88:

Example of old version string: `23.10-0.2.2.0`
Example of new version string: `24.01-0.3.3.1-0`

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,7 +8,7 @@ require (
 	github.com/caarlos0/env/v6 v6.10.1
 	github.com/containers/image/v5 v5.29.2
 	github.com/go-logr/logr v1.4.1
-	github.com/k8snetworkplumbingwg/network-attachment-definition-client v1.4.0
+	github.com/k8snetworkplumbingwg/network-attachment-definition-client v1.6.0
 	github.com/onsi/ginkgo/v2 v2.15.0
 	github.com/onsi/gomega v1.31.1
 	github.com/openshift/api v0.0.0-20231120222239-b86761094ee3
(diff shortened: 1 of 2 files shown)


e0ne (approved):
LGTM

rollandf:
@maze88 LGTM , but why is there 2 commits?

api/v1alpha1/validator/nicclusterpolicy_webhook.go
```diff
@@ -659,7 +659,7 @@ func validateResources(resources map[v1.ResourceName]apiresource.Quantity, allEr
 
 // isValidOFEDVersion is a custom function to validate OFED version
 func isValidOFEDVersion(version string) bool {
-	versionPattern := `^(\d+\.\d+-\d+(\.\d+)*)$`
+	versionPattern := `^(\d+\.\d+-\d+(\.\d+)*(-\d+)?)$`
```
adrianchiris, line 662: nit: * -> +

we could even do it more strict e.g {3,4}

i think this release we have like 5 numbers as the patch version right @e0ne ?

so we got:

"old" version - container version not embedded: e.g 23.10-0.2.2.0
"current" version - container version embedded, dot separated: e.g 24.04-0.3.3.1.1
"future" version - container version embedded, dash separated: e.g 24.07-0.4.5.6-0

adrianchiris: nothing is blocker, ill let @e0ne decide.

killianmuldoon: One thing we should definitely do is enable the webhook in the CI so issues like this are caught in future - that has to be done on the CI repo though.

adrianchiris: added to backlog

maze88:
I recreated the [PR here](https://github.com/Mellanox/network-operator/pull/835), with clean commit history.

PR #832, opened 2024-02-25
chore: update official doc version

rollandf:
Update doc link to 24.1

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ RDMA and GPUDirect RDMA workloads in a kubernetes cluster including:
 * Kubernetes secondary network for Network intensive workloads
 
 ## Documentation
-For more information please visit the official [documentation](https://docs.nvidia.com/networking/display/kubernetes2310/Network+Operator).
+For more information please visit the official [documentation](https://docs.nvidia.com/networking/display/kubernetes2410).
 
 
 ## Prerequisites


killianmuldoon (approved):
/lgtm

README.md
```diff
@@ -43,7 +43,7 @@ RDMA and GPUDirect RDMA workloads in a kubernetes cluster including:
 * Kubernetes secondary network for Network intensive workloads
 
 ## Documentation
-For more information please visit the official [documentation](https://docs.nvidia.com/networking/display/kubernetes2310/Network+Operator).
+For more information please visit the official [documentation](https://docs.nvidia.com/networking/display/kubernetes2410).
```
rollandf, line 46: Should we use just a most general landing page, listing the versions? It will save us updating this each release.

https://docs.nvidia.com/networking/software/cloud-orchestration/index.html

killianmuldoon: I'm not sure how the routing works for these - but could we just have the base page point to latest? A first change would be to make changing this part of the release task so that we don't have to do it in a separate PR.

e0ne: LGTM. We'll probably update in to some 'latest' version once we'll switch to new documentation process

e0ne (approved):

README.md
rollandf: Not sure about having routing from latest. We should check with technical writers team about it.

PR #835, opened 2024-02-29

maze88:

Example of old version string: `23.10-0.2.2.0`
Example of new version string: `24.01-0.3.3.1-0`

### DIFF
--- a/api/v1alpha1/validator/nicclusterpolicy_webhook.go
+++ b/api/v1alpha1/validator/nicclusterpolicy_webhook.go
@@ -659,7 +659,7 @@ func validateResources(resources map[v1.ResourceName]apiresource.Quantity, allEr
 
 // isValidOFEDVersion is a custom function to validate OFED version
 func isValidOFEDVersion(version string) bool {
-	versionPattern := `^(\d+\.\d+-\d+(\.\d+)*)$`
+	versionPattern := `^(\d+\.\d+-\d+(\.\d+)*(-\d+)?)$`
 	versionRegex := regexp.MustCompile(versionPattern)
 	return versionRegex.MatchString(version)
 }
--- a/api/v1alpha1/validator/nicclusterpolicy_webhook_test.go
+++ b/api/v1alpha1/validator/nicclusterpolicy_webhook_test.go
@@ -100,7 +100,7 @@ var _ = Describe("Validate", func() {
 				ContainSubstring("pKeyGUIDPoolRangeStart must be a valid GUID format"),
 				ContainSubstring("pKeyGUIDPoolRangeEnd must be a valid GUID format")))
 		})
-		It("Valid MOFED version", func() {
+		It("Valid MOFED version (old scheme)", func() {
 			validator := nicClusterPolicyValidator{}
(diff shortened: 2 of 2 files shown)


adrianchiris:
git rebase didnt work ?

can you please move unresolved comments to this PR ?

maze88:
@adrianchiris [wrote](https://github.com/Mellanox/network-operator/pull/831#discussion_r1507178967):

> nit: * -> +
> 
> we could even do it more strict e.g {3,4}
> 
> i think this release we have like 5 numbers as the patch version right @e0ne ?
> 
> so we got:
> 
> "old" version - container version not embedded: e.g 23.10-0.2.2.0 "current" version - container version embedded, dot separated: e.g 24.04-0.3.3.1.1 "future" version - container version embedded, dash separated: e.g 24.07-0.4.5.6-0
> 
> im thinking to add explicit tests for each so we have some documentation around the expected format

adrianchiris:
@adrianchiris also wrote in closed PR:

> nothing is blocker, ill let @e0ne decide.

e0ne:
> @adrianchiris [wrote](https://github.com/Mellanox/network-operator/pull/831#discussion_r1507178967):
> 
> > nit: * -> +
> > we could even do it more strict e.g {3,4}
> > i think this release we have like 5 numbers as the patch version right @e0ne ?
> > so we got:
> > "old" version - container version not embedded: e.g 23.10-0.2.2.0 "current" version - container version embedded, dot separated: e.g 24.04-0.3.3.1.1 "future" version - container version embedded, dash separated: e.g 24.07-0.4.5.6-0
> > im thinking to add explicit tests for each so we have some documentation around the expected format

I don't think we need to be too strict here. From other side it's a good idea to have unit-tests for three patterns:
* 23.10-0.2.2.0 (old scheme)
* 23.10-0.2.2.0.1 (old scheme with container version suffix)
* 23.10-0.2.2.0-1 (new scheme)

docs/mofed-container-env-vars.md
```diff
@@ -23,6 +23,7 @@ In most deployments its expected to not require setting those.
 | UNLOAD_STORAGE_MODULES |N|`"false"`| unload host storage modules prior to loading mofed modules  |
 | ENABLE_NFSRDMA |N|`"false"`| enable loading of nfs relates storage modules from mofed container|
 | RESTORE_DRIVER_ON_POD_TERMINATION |N|`"true"`| restore host drivers when container is gracefully stopped |
+| NVIDIA_NIC_DRIVERS_INVENTORY_PATH | N | `"/mnt/drivers-inventory"` | enable use of a persistent directory to store drivers' build artifacts to avoid recompilation between runs. Keep the default value or set to "" to disable. |
```
killianmuldoon, line 26: Looks a bad rebase - could you clean this up? I think we should only have webhook + tests in this change.

maze88: corrected, thank you @killianmuldoon!

e0ne (approved):
Thanks for addressing my comments, Michael!

rollandf (approved):

PR #837, opened 2024-03-05
chore: use stdlib slices

rollandf:
Use stdlib slices

### DIFF
--- a/api/v1alpha1/validator/nicclusterpolicy_webhook.go
+++ b/api/v1alpha1/validator/nicclusterpolicy_webhook.go
@@ -25,13 +25,13 @@ import (
 	"os"
 	"path/filepath"
 	"regexp"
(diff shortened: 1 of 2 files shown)


killianmuldoon (approved):
/lgtm

Thanks for following up with this!

killianmuldoon:
Looks like we're building with the wrong version of go - `make deploy` is failing in the kind test. Not sure how it's building as the Docker file is from go 1.21

rollandf:
> Looks like we're building with the wrong version of go - `make deploy` is failing in the kind test. Not sure how it's building as the Docker file is from go 1.21

CI worker is 1.19
# go version
go version go1.19.2 linux/amd64

killianmuldoon:
go compiler is used in `controller-gen` so it makes sense.

PR #844, opened 2024-03-07

rollandf:

Example of old version string: 23.10-0.2.2.0
Example of old version string: 23.10-0.2.2.0.1 (with container version suffix)
Example of new version string: 24.01-0.3.3.1-0

Signed-off-by: Michael Zeevi <mzeevi@nvidia.com>
(cherry picked from commit e35704f1fea50b33926d2dc583ab51dff676bd26)


killianmuldoon (approved):
/lgtm

ykulazhenkov (approved):

e0ne (approved):

PR #881, opened 2024-04-08
feat: update chart for sriov-network-operator

ykulazhenkov:
Update chart for sriov-network-operator to match downstream version of sriov-network-operator

\+ use downstream version of ovs-cni


hack/release.yaml
```diff
@@ -71,5 +71,5 @@ docaTelemetryService:
   version: 1.16.5-doca2.6.0-host
 ovsCni:
   image: ovs-cni-plugin
-  repository: quay.io/kubevirt
-  version: v0.32.0
+  repository: nvcr.io/nvstaging/mellanox
```
killianmuldoon, line 74: will this break some users who previously had access to the kubevirt repo? Is this just temperorary?

ykulazhenkov: We are temporally switching to downstream version of ovs-cni to include new functionality which is not merged to the upstream yet. This functionality is fully backward compatible and should not break anything. We expect that needed features will be eventually part of the upstream and we will switch to the upstream image.

e0ne (approved):

PR #885, opened 2024-04-08
Remove test data from helm values

e0ne:
We don't use 'helm test' in our CI for a while so it' safe to remove it now.


ykulazhenkov (approved):

rollandf:
Can you please check if README need update about the test:

https://github.com/Mellanox/network-operator/blob/master/deployment/network-operator/README.md#helm-tests

PR #890, opened 2024-04-15
chore: Update SR-IOV Network Operator chart to deploy RDMA CNI

e0ne:
Update SR-IOV Network Operator chart to deploy RDMA CNI


ykulazhenkov (approved):
LGTM

PR #895, opened 2024-04-17
feat: Support errors on NCP status

### LINKED ISSUE #533: Errors are not reported in CR status
**What happened**:

if an error occurs when syncing state, we should update global status with the first error encountered (in controller)

while this is true for some controllers, it is not for others
in addition the different states  in most cases that an error occured return `SyncStateNotReady` instead of `SyncStateError`
this should be reviewed and fixed for all states.

rollandf:
Fixes #533

Example status:
```
status:
  appliedStates:
  - name: state-multus-cni
    state: ignore
  - name: state-container-networking-plugins
    state: ignore
  - name: state-ipoib-cni
    state: ignore
  - name: state-whereabouts-cni
    state: ignore
  - error: 'failed to create k8s objects from manifest: failed to render objects:
      ForcePrecompiled is enabled and precompiled image was not found'
    name: state-OFED
    state: notReady
  - name: state-SRIOV-device-plugin
    state: ignore

```

### DIFF
--- a/api/v1alpha1/nicclusterpolicy_types.go
+++ b/api/v1alpha1/nicclusterpolicy_types.go
@@ -291,6 +291,9 @@ type AppliedState struct {
 	Name string `json:"name"`
 	// +kubebuilder:validation:Enum={"ready", "notReady", "ignore", "error"}
 	State State `json:"state"`
+	// Message is a human readable message indicating details about why
+	// the state is in this condition
+	Message string `json:"message,omitempty"`
 }
 
 // NicClusterPolicyStatus defines the observed state of NicClusterPolicy
--- a/controllers/nicclusterpolicy_controller_test.go
+++ b/controllers/nicclusterpolicy_controller_test.go
@@ -138,6 +138,84 @@ var _ = Describe("NicClusterPolicyReconciler Controller", func() {
 			Expect(err).NotTo(HaveOccurred())
 		})
 	})
+	Context("When MOFED precompiled tag does not exists", func() {
+		It("should set error message in status", func() {
+			By("Create Node")
+			node := &corev1.Node{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "test-node",
+					Labels: map[string]string{
+						nodeinfo.NodeLabelMlnxNIC:       "true",
+						nodeinfo.NodeLabelOSName:        "ubuntu",
+						nodeinfo.NodeLabelCPUArch:       "amd64",
+						nodeinfo.NodeLabelKernelVerFull: "generic-9.0.1",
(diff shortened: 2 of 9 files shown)


rollandf:
Need to add unit tests, and check how the different states are behaving

rollandf:
> Need to be a bit careful with this as any error which is different on each reconcile - e.g. which contains the generated name of an object - will cause infinite super-fast reconciles

AFAIK, the generated yamls are the same on each reconcile. Otherwise, we would have tons of k8s objects created.

api/v1alpha1/nicclusterpolicy_types.go
```diff
@@ -291,6 +291,8 @@ type AppliedState struct {
 	Name string `json:"name"`
 	// +kubebuilder:validation:Enum={"ready", "notReady", "ignore", "error"}
 	State State `json:"state"`
+	// Error string in case the state is error
+	Error string `json:"error,omitempty"`
```
adrianchiris, line 295: can we call this "message"  ? 

// message is a human readable message indicating details about why the state is in this condition.

similar to what we have in pod status

adrianchiris (commented):
small comment about naming

api/v1alpha1/nicclusterpolicy_types.go
rollandf: Done

adrianchiris (dismissed):

controllers/nicclusterpolicy_controller_test.go
```diff
@@ -138,6 +138,84 @@ var _ = Describe("NicClusterPolicyReconciler Controller", func() {
...
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "nic-cluster-policy",
+					Namespace: "",
+				},
+				Spec: mellanoxv1alpha1.NicClusterPolicySpec{
+					OFEDDriver: &mellanoxv1alpha1.OFEDDriverSpec{
+						ForcePrecompiled: true,
+						ImageSpec: mellanoxv1alpha1.ImageSpec{
+							Image:            "mofed",
+							Repository:       "nvcr.io/nvidia/mellanox",
```
ykulazhenkov, line 169: The test is depends on the facts that NGC is available and that tag for Kernel generic-9.0.1 is not exist? Right?
I don't like the fact that we depend on uncontrolled external state in tests. Maybe we can mock DocaDriverImagesProvider? I think it is fine to create single mocked instance of DocaDriverImagesProvider which will work for entire tests suite. WDYT?

rollandf: Done

ykulazhenkov (approved):
LGTM

adrianchiris (approved):
LGTM!

PR #896, opened 2024-04-19
feat: Allow to configure and deploy RDMA CNI image

e0ne:
Allow to configure and deploy RDMA CNI image

### DIFF
--- a/hack/release.yaml
+++ b/hack/release.yaml
@@ -73,3 +73,7 @@ ovsCni:
   image: ovs-cni-plugin
   repository: nvcr.io/nvstaging/mellanox
   version: 8b231bb
+rdmaCni:
+  image: rdma-cni
+  repository: ghcr.io/k8snetworkplumbingwg
+  version: latest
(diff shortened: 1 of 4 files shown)


hack/release.yaml
```diff
@@ -73,3 +73,7 @@ ovsCni:
   image: ovs-cni-plugin
   repository: nvcr.io/nvstaging/mellanox
   version: 61fd27c
+rdmaCni:
+  image: rdma-cni
+  repository: ghcr.io/k8snetworkplumbingwg
+  version: latest
```
rollandf, line 79: I think we should have a specific version here, no?

e0ne: I would like to keep `latest` for beta releases to simplify testing as it contains multi-arch images. If you think it's better to use some tag for beta release I'll update PR

rollandf: Ack.
Just let's make sure that we do not release with `latest`

rollandf (approved):

ykulazhenkov (dismissed):

PR #897, opened 2024-04-19
chore: Update base image to ubi8-micro:8.9-15

e0ne:
Update base image to ubi8-micro:8.9-15


adrianchiris:
@e0ne could you rebase and re-trigger CI, i think it should all pass now.

PR #900, opened 2024-04-22
chore: Add coverage report to CI checks

almaslennikov:
A separate PR to follow for the README badge
Might need to provide access to Mellanox org

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -92,3 +92,12 @@ jobs:
         uses: actions/checkout@v4
       - name: test
         run: make test-coverage
+      - name: Generate LCOV report
+        run: |
+          go install github.com/jandelgado/gcov2lcov@latest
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -92,3 +92,12 @@ jobs:
         uses: actions/checkout@v4
       - name: test
         run: make test-coverage
+      - name: Generate LCOV report
+        run: |
+          go install github.com/jandelgado/gcov2lcov@latest
```
ykulazhenkov, line 97: do we want to add Makefile target for this command?

ykulazhenkov: ok. let's keep this as a part of GH action manifest

ykulazhenkov (approved):

rollandf (approved):

PR #929, opened 2024-05-10
feat: Add RHEL support in ofed config

killianmuldoon:
Add support for RHEL filepaths in the ofed manifest generation. This enables the `repoConfig` and `certConfig` options when running OFED on RHEL nodes.


almaslennikov (approved):

e0ne (approved):

rollandf:
@e0ne Do we need this one to be backported?

e0ne:
> @e0ne Do we need this one to be backported?

we do. here is it #933

PR #935, opened 2024-05-14
feat: add github actions job for helm publishing to CI workflow

almaslennikov:
this job will replace our legacy jenkins job, and be dependent on the new github action job for the network operator docker image.


e0ne:
merged in #963

PR #945, opened 2024-05-20
fix: refactor docker tag and helm version determination via triggers

maze88:
Git tag events (regardless of the branch) would not trigger the old workflow, and push branch events would trigger it - including the Helm CI, which would sometimes not be able to determine any version to set, thus failing.

This commit fixes the above, by:
- Add a trigger for _pushed tag_ events.
- Determine docker tags and chart versions based on trigger event _type_.
- Add conditions to Helm CI job steps, allowing it to be skipped (but _not_ failed!) in cases where it is not required to run (such as an untagged intermediate commits on release branches).

### DIFF
--- a/.github/workflows/network-operator-docker-and-helm-ci.yaml
+++ b/.github/workflows/network-operator-docker-and-helm-ci.yaml
@@ -5,26 +5,31 @@ on:
     branches:
     - "master"
     - "v[0-9]+.[0-9]+.x"
+    tags:
+    - "v[0-9]+.[0-9]+.[0-9]+*"
 
 env:
-  REGISTRY: nvcr.io/nvstaging/mellanox  # for docker
-  IMAGE_NAME: network-operator
-  VALID_TAG_REGEX: 'v[0-9]+\.[0-9]+\.[0-9]+(-(beta|rc)\.[0-9]+)?'
-  NGC_REPO: nvstaging/mellanox/network-operator  # for helm
+  DEFAULT_BRANCH: master
 
 jobs:
   docker-build-push:
     runs-on: ubuntu-latest
+    env:
+      REGISTRY: nvcr.io/nvstaging/mellanox
+      IMAGE_NAME: network-operator  # used in makefile
     steps:
     - uses: actions/checkout@v4
-      with:
-        fetch-tags: true
-    - name: Determine docker tags
+    - if: github.ref_type == 'branch'
+      name: Determine docker tags (when git branch)
       run: |
         git_sha=$(git rev-parse --short HEAD)  # short git commit hash
-        git_tag=$(git describe --tags --exact-match 2>/dev/null | grep -E '${{ env.VALID_TAG_REGEX}}' || true)  # git tag, if it exists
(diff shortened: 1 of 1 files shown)


.github/workflows/network-operator-docker-and-helm-ci.yaml
```diff
@@ -5,26 +5,31 @@ on:
     branches:
     - "master"
     - "v[0-9]+.[0-9]+.x"
+    tags:
+    - 'v[0-9]+.[0-9]+.[0-9]+*'
```
almaslennikov, line 9: nit: Probably better for code style consistency to use the same type of quotes throughout the file

.github/workflows/network-operator-docker-and-helm-ci.yaml
```diff
@@ -5,26 +5,31 @@ on:
...
 
 env:
-  REGISTRY: nvcr.io/nvstaging/mellanox  # for docker
-  IMAGE_NAME: network-operator
-  VALID_TAG_REGEX: 'v[0-9]+\.[0-9]+\.[0-9]+(-(beta|rc)\.[0-9]+)?'
-  NGC_REPO: nvstaging/mellanox/network-operator  # for helm
+  STABLE_BRANCH: master
```
almaslennikov, line 12: Not the stale branch, probably, maybe just `MAINLINE_BRANCH`/`CD_BRANCH` or smth like this?

.github/workflows/network-operator-docker-and-helm-ci.yaml
```diff
@@ -40,29 +45,34 @@ jobs:
...
+        app_version=$(git rev-parse --short HEAD)  # short git commit hash
+        current_chart_version=$(yq '.version' deployment/network-operator/Chart.yaml)
+        echo APP_VERSION=""$app_version""                    >> $GITHUB_ENV
+        echo VERSION=""$current_chart_version-$app_version"" >> $GITHUB_ENV
+    - if: github.ref_type == 'tag'
+      name: Determine versions (when git tag)
+      run: |
+        git_tag=${{ github.ref_name }}
+        app_version=$git_tag
+        chart_version=${git_tag:1}  # without the 'v' prefix
```
almaslennikov, line 65: (Can do after the merge, just so we don't forget) Need to make a test release to check that all versions in the helm chart / image versions are correct

almaslennikov (approved):
A few nits, otherwise LGTM

.github/workflows/network-operator-docker-and-helm-ci.yaml
maze88: some parts it's syntactically important, for github actions. but not here - i'll change it, ty!

.github/workflows/network-operator-docker-and-helm-ci.yaml
maze88: you're right, we develop there so it's _not_ stable. i'll call it default - since that's the term used also in the github repository settings.

almaslennikov (approved):

e0ne (approved):

e0ne:
CI failures are not related to this change, so it's safe to merge

PR #946, opened 2024-05-20
feat: MOFED subscription mounts

rollandf:
Add needed mounts for subscription in RHEL with non CRIO container runtime.

### DIFF
--- a/pkg/nodeinfo/node_info.go
+++ b/pkg/nodeinfo/node_info.go
@@ -22,6 +22,7 @@ package nodeinfo
 
 import (
 	"fmt"
+	"strings"
 
 	corev1 "k8s.io/api/core/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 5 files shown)


pkg/nodeinfo/node_info.go
```diff
@@ -127,6 +129,8 @@ func (p *provider) GetNodePools(filters ...Filter) []NodePool {
 		}
 		nodePool.Kernel = kernel
 
+		nodePool.ContainerRuntime = getContainerRuntime(node)
```
adrianchiris, line 132: do you assume that container runtime is the same for all nodes that are part of the pool ?

technically every attribute that is not part of the key in nodePoolMap may be different and you just save the last node that match.

pkg/state/state_ofed.go
```diff
@@ -783,6 +825,40 @@ func (s *stateOFED) handleCertConfig(
 	return nil
 }
 
+// handleSubscriptionVolumes handles additional mounts required for subscriptions
+func (s *stateOFED) handleSubscriptionVolumes(
+	ctx context.Context, osname string, runtime string, mounts *additionalVolumeMounts) error {
+	reqLogger := log.FromContext(ctx)
+	if osname == "rhel" && runtime != nodeinfo.CRIO {
```
adrianchiris, line 832: do we want to ignore an empty runtime value ?

rollandf: Can an empty runtime happen?

rollandf:
Tested manually on cluster with RHEL 8.8, containerd 1.7.16
MOFED pods compiles successfully

pkg/state/state_ofed.go
adrianchiris: field should be set. just wondering if we should add check for completness.

pkg/nodeinfo/node_info.go
adrianchiris: ack.

adrianchiris (approved):

PR #954, opened 2024-05-24
fix: Align ib-kubernetes RBAC permissions with those of network-operator

almaslennikov:
Network-operator doesn't allow to get * resources from k8s.cni.cncf.io. When ib-k8s deploys, it asks for more permissions thus failing to start


rollandf (approved):

e0ne (approved):
Copy Right Validation CI failure isn't related to this change

PR #958, opened 2024-05-27
chore: Bump RDMA CNI to v1.1.0 and deploy enable it in helm values

e0ne:


e0ne:
We don't deploy RDMA CNI on our CI so it's safe to merge

PR #970, opened 2024-06-11
chore: Use NVIDIA IPAM Plugin by default

e0ne:


ykulazhenkov (approved):

example/crs/nvidia.com_v1alpha1_ nv_ipam_ippool_cr.yaml
```diff
@@ -0,0 +1,23 @@
+# 2024 NVIDIA CORPORATION & AFFILIATES
```
rollandf, line 1: Do you have a space in the file name here?

rollandf (approved):

example/crs/nvidia.com_v1alpha1_ nv_ipam_ippool_cr.yaml
e0ne: yes :(, it's fixed now

rollandf (approved):

ykulazhenkov (approved):

adrianchiris (approved):

example/rdma-gpu-test-pod1.yml
```diff
@@ -6,7 +6,7 @@ metadata:
     k8s.v1.cni.cncf.io/networks: rdma-net-ipam
```
krembu, line 6: @e0ne seems like a tyope and change should be on this line

krembu:
@e0ne  if we default nv-ipam for vanilla kubernetes i think we should enable it by default in openshift

PR #975, opened 2024-06-19
[DNM] chore: Bump multus version thick plugin

adrianchiris:
- deploy multus thick plugin. update manifests accordingly
- use sha256 digest as there is no recent release version
  of multus for now.

Note: we should first update PR to point to a version of multus (preferably a release tag) which supports multiarch ([PR1297](https://github.com/k8snetworkplumbingwg/multus-cni/pull/1297))

### DIFF
--- a/deployment/network-operator/values.yaml
+++ b/deployment/network-operator/values.yaml
@@ -362,9 +362,9 @@ secondaryNetwork:
     #       memory: "50Mi"
   multus:
     deploy: true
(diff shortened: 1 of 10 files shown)


adrianchiris:
one thing to note: due to the lack of a recent multus release, we use a sha256 digest of a recent commit.

rollandf:
closing, see #3191

PR #995, opened 2024-07-16
chore: Add unit tests for nv-ipam-cni objects rendering and sync

souleb:


pkg/state/common_test.go
```diff
@@ -25,18 +25,28 @@ import (
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 )
 
+const (
+	defaultTestRepository = "myRepo"
+	defaultTestImage      = "myImage"
+	defaultTestVersion    = "myVersion"
+)
+
```
tobiasgiese, line 49: In https://github.com/Mellanox/network-operator/pull/996 I am moving the consts to the test scope. Otherwise, the complete `state_test` package will have these constants. It is pretty confusing if we define consts here and we can use them anyway. We should define and pass them explicitly, WDYT

pkg/state/state_nv_ipam_cni_test.go
```diff
@@ -21,33 +21,347 @@ import (
...
+		nvIpamState = s
+		renderer = r
+		catalog = getTestCatalog()
+		catalog.Add(state.InfoTypeStaticConfig,
+			staticconfig.NewProvider(staticconfig.StaticConfig{CniBinDirectory: "custom-cni-bin-directory"}))
+		namespace = config.FromEnv().State.NetworkOperatorResourceNamespace
+	})
+
+	Context("Verify objects rendering", func() {
+		It("should create Daemonset - minimal spec", func() {
```
tobiasgiese, line 73: I know we don't have this in our other tests, but should we start adding negative tests?

tobiasgiese: NB: this is just a reminder, nothing related to this PR :)

pkg/state/common_test.go
souleb: I have those three global because many tests need to define an image in `.spec`.

tobiasgiese: Ah, it's in the `common_test.go`, overlooked that. Thought I was in the state test

pkg/state/state_nv_ipam_cni_test.go
```diff
@@ -21,33 +21,389 @@ import (
...
+	return cr
+}
 
-	_, s, err := state.NewStateNVIPAMCNI(fake.NewClientBuilder().Build(), "../../manifests/state-nv-ipam-cni")
-	Expect(err).NotTo(HaveOccurred())
-	It("should test that manifests are rendered and fields are set correctly", func() {
-		GetManifestObjectsTest(ctx, cr, catalog, imageSpec, s)
-	})
-})
+func getExpectedNVIpamClusterRoleBinding(name, namespace string) *rbacv1.ClusterRoleBinding {
```
killianmuldoon, line 331: I don't think we should have tests for fields that don't have any logic around them as it will make the manifests difficult to change and update. We should only have tests for fields which are templated and particularly those which have logic associated with them.

killianmuldoon: Similare with getExpectedNVIpamNodeClusterRole and getExpectedNVIpamControllerClusterRole

killianmuldoon (commented):
Looks good - just some comments around simplifying it a bit

pkg/state/info_source.go
```diff
@@ -45,31 +45,37 @@ func NewInfoCatalog() InfoCatalog {
 // InfoSource represents an object that is a source of information.
 type InfoSource interface{}
 
-// InfoCatalog is an information catalog to be used to retrieve infoSources. used for State implementation that require
-// additional helping functionality to perfrom the Sync operation. As more States are added,
-// more infoSources may be added to aid them. for any infoSource if not present in the catalog, nil will be returned.
+// InfoCatalog is an information catalog to be used to retrieve infoSources.
```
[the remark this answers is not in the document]
souleb: right, I'll revert it. This can be a separate PR.

pkg/state/state_nv_ipam_cni.go
```diff
@@ -71,7 +71,7 @@ type NVIPAMManifestRenderData struct {
 //
 //nolint:dupl
 func (s *stateNVIPAMCNI) Sync(
-	ctx context.Context, customResource interface{}, infoCatalog InfoCatalog) (SyncState, error) {
+	ctx context.Context, customResource any, infoCatalog InfoCatalog) (SyncState, error) {
```
e0ne, line 74: As we use `interface{}` over all our codebase I prefer to use the same approach across all the code so we don't need this change

pkg/state/state_nv_ipam_cni_test.go
```diff
@@ -21,33 +21,266 @@ import (
...
+			},
+		},
+	))
+}
+
+func getMinimalNicClusterPolicyWithNVIpam(name string) *mellanoxv1alpha1.NicClusterPolicy {
 	cr := getTestClusterPolicyWithBaseFields()
-	cr.Spec.NvIpam = &mellanoxv1alpha1.NVIPAMSpec{
-		// TODO(killianmuldoon): Test with the webhook enabled.
+	cr.Name = "nic-cluster-policy"
```
e0ne, line 276: Why do we need Name field only for these new tests?

e0ne (changes requested):

pkg/state/state_nv_ipam_cni_test.go
souleb: right, we don't need it in the tests here.

e0ne (approved):

rollandf (approved):

PR #998, opened 2024-07-17
fix: add timeout to mofed probes

adrianchiris:
we have observed on some systems lsmod command timesout. current default is 1s. increase to 10s.


adrianchiris:
@e0ne lets get this one in :)

noama-nv (approved):

e0ne (approved):
LGTM

PR #999, opened 2024-07-18
chore: implement new assert function for NAD unit tests

tobiasgiese:
Todos:

- [x] create new assert func
- [x] adopt for hostdevice network
- [x] adopt for ipoib
- [x] adopt for macvlan
- [x] verify if everything covered (only refactoring necessary IMO)

### DIFF
--- a/pkg/state/common_test.go
+++ b/pkg/state/common_test.go
@@ -23,6 +23,8 @@ import (
 
 	. "github.com/onsi/gomega"
 
+	netattdefv1 "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
+
 	mellanoxv1alpha1 "github.com/Mellanox/network-operator/api/v1alpha1"
 	clustertype_mocks "github.com/Mellanox/network-operator/pkg/clustertype/mocks"
 	"github.com/Mellanox/network-operator/pkg/state"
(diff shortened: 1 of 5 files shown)


pkg/state/common_test.go
```diff
@@ -169,6 +193,25 @@ func assertCNIBinDirForDS(u *unstructured.Unstructured) {
 	}
 }
 
+func assertNetworkAttachmentDefinition(c client.Client, expectedNadConfig *nadConfig,
```
ykulazhenkov, line 196: Q: Should we use here `GinkgoHelper()` or `ExpectWithOffset`?

tobiasgiese: This would be helpful for sure. We are never using `GinkgoHelper` or `ExpectWithOffset` in our tests.
Should we create a separate PR to align our tests?

ykulazhenkov: yes, I think we can update all helper functions in a separate PR

e0ne (dismissed):
LGTM

rollandf (dismissed):

tobiasgiese:
Had to update the first commit, change is [between `3e7c24c` and `8a07cad`](https://github.com/Mellanox/network-operator/compare/3e7c24ca91e080652534bb20de195fff477fbe9a..8a07cad31fc13b6e20229b12fbc218e346d9632a)

rollandf (approved):

ykulazhenkov (approved):

PR #1004, opened 2024-07-24
chore: fail on error on make bundle

rollandf:
Make sure that `release.go` exit with error code on failure.

### DIFF
--- a/hack/scripts/ocp-bundle-postprocess.sh
+++ b/hack/scripts/ocp-bundle-postprocess.sh
@@ -23,7 +23,12 @@ if [[ "${TRACE-0}" == "1" ]]; then
 fi
 
 # Generate relatedImages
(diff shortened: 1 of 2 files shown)


ykulazhenkov (approved):

hack/scripts/ocp-bundle-postprocess.sh
```diff
@@ -23,7 +23,12 @@ if [[ "${TRACE-0}" == "1" ]]; then
 fi
 
 # Generate relatedImages
-cd hack && $GO run release.go --with-sha256 --templateDir ./templates/related-images/ --outputDir . && cd ..
+cd hack
```
adrianchiris, line 26: nit: you can use pushd and popd . same effect as current logic

adrianchiris (approved):

PR #1001, opened 2024-07-22
chore: Add unit test for state-sriov-dp

souleb:

PR #1006, opened 2024-07-25
test: implement testScope for state unit tests

tobiasgiese:
We have alot of duplicated code in our unit tets. To reduce this we can implement a test scope that can be reused in all of our state unit tests.


e0ne (approved):
LGTM

rollandf:
@tobiasgiese Can you please rebase?

rollandf (approved):

tobiasgiese:
We maybe should wait with merge until https://github.com/Mellanox/network-operator/pull/1001 is merged. Otherwise, @souleb must rebase, resolve conflicts and adopt the changes

pkg/state/state_hostdevice_network.go
```diff
@@ -55,7 +55,7 @@ func NewStateHostDeviceNetwork(k8sAPIClient client.Client, manifestDir string) (
 			description: stateHostDeviceNetworkDescription,
 			client:      k8sAPIClient,
 			renderer:    renderer,
-		}}, nil
+		}}, nil, nil
```
adrianchiris, line 58: no need to return state as MainfestRenderer instead of nil? (here and in other modified states in this PR)

in hindsight perhaps we should have embedded MainifestRenderer in State interface and just return one value.
but that not related to this PR.

pkg/state/state_rdma_shared_device_plugin_test.go
```diff
@@ -17,70 +17,45 @@ limitations under the License.
...
 var _ = Describe("RDMA Shared Device Plugin", func() {
-
-	var ctx context.Context
-	var rdmaDPState state.State
-	var rdmaDPRenderer state.ManifestRenderer
-	var testLogger logr.Logger
-	var catalog state.InfoCatalog
-	var client client.Client
-	var manifestDir string
+	var t testScope
```
adrianchiris, line 28: for me, this reminds me of the `t *testing.T` in (non ginkgo) unit-tests we see in the the Test<Something> funcs.
which is a bit confusing. any chance to rename to something else ?  (tScope, tscope, ts etc...)

that said, i can live with current name if you really like it.

pkg/state/state_hostdevice_network.go
tobiasgiese: The problem is that if we also want to return the renderer here we have to adapt the `getManifestObjects` and make it also public. We can do this for sure, but it is not needed and we can simply return `nil` here.
I don't have a real opinion about that, but I can also change it to return the renderer here as well. WDYT?

tobiasgiese:
overseen the multus-cni state tests, pushed them as well

pkg/state/state_hostdevice_network.go
adrianchiris: Ah i see, i though all states implement MainfestRenderer, i see this is not the case.

ack, lets keep it as is.

adrianchiris (approved):

tobiasgiese:
@rollandf @adrianchiris as #1001 is merged the PR is now ready to be merged. Added only the changes for the state sriov dp tests in https://github.com/Mellanox/network-operator/pull/1006/commits/26083b99f1aa72d2476af3ad9b6c1d506f8b365d

rollandf (approved):

adrianchiris (approved):

PR #1007, opened 2024-07-26
chore: Enable api and manifests regeneration before testing

souleb:
I would like to enable `fmt/vet` as well as updating module and regenerating `apis` and `manifests` in an automated manner as part of running tests targets.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -271,14 +271,14 @@ setup-envtest: $(SETUP_ENVTEST)  ## Install envtest binaries
 clean-envtest: setup-envtest ;## Clean up assets installed by setup-envtest
 	$Q $(SETUP_ENVTEST) cleanup
 
-check test tests: setup-envtest ; $(info  running $(NAME:%=% )tests...) @ ## Run tests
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -271,14 +271,14 @@ setup-envtest: $(SETUP_ENVTEST)  ## Install envtest binaries
 clean-envtest: setup-envtest ;## Clean up assets installed by setup-envtest
 	$Q $(SETUP_ENVTEST) cleanup
 
-check test tests: setup-envtest ; $(info  running $(NAME:%=% )tests...) @ ## Run tests
+check test tests: tidy generate fmt vet manifests setup-envtest ; $(info  running $(NAME:%=% )tests...) @ ## Run tests
```
rollandf, line 274: Do we really need to run these on `make test`? Won't it make running tests locally longer?
Maybe we should have build check that tidy/fmt/vet/generate are passing without any file changes, in the same manner as we have for manifests?
Here is the check for manifests: https://github.com/Mellanox/network-operator/blob/master/.github/workflows/build.yml#L52

souleb: time wise this is trivial compared to envtest setup and actual tests. My idea is to make sure that locally running test gives me the best possible confidence.

PR #1011, opened 2024-07-29
feat: Add description to helm values

souleb:
The `values.template` is updated to render a `values.yaml` containing the needed descriptions, and has to be updated for new fields if needed.

This is needed in order to properly render the `helm.rst` in Network-Operator-docs. **Note** All new fields will be in `General Parameters` section by default. If this is not desirable, the `helm.rst.gotmpl` has to be update in Network-Operator-docs.

See rendered doc: https://github.com/Mellanox/network-operator/blob/6ae08a018f019d12fc13d4630209e25cd22cc5ef/docs/charts/helm.rst

### DIFF
--- a/deployment/network-operator/values.yaml
+++ b/deployment/network-operator/values.yaml
@@ -17,21 +17,27 @@
 # Declare variables to be passed into your templates.
 
 nfd:
+  # -- Deploy Node Feature Discovery operator.
   enabled: true
+  # -- Deploy Node Feature Rules to label the nodes with the discovered features.
   deployNodeFeatureRules: true
 
+# -- Enable CRDs upgrade with helm pre-install and pre-upgrade hooks.
 upgradeCRDs: true
 
 sriovNetworkOperator:
+  # -- Deploy SR-IOV Network Operator.
   enabled: false
 
-# Node Feature discovery chart related values
+# Set both enableNodeFeatureApi and NodeFeatureAPI feature gate to false to disable.
 node-feature-discovery:
-  # set both enableNodeFeatureApi and NodeFeatureAPI feature gate to false to disable
+  # -- The Node Feature API is used to expose node-specific hardware and 
+  # software features to the Kubernetes scheduler.
   enableNodeFeatureApi: true
   featureGates:
     NodeFeatureAPI: true
 
+  # -- NFD master deployment configuration.
   master:
     serviceAccount:
       name: node-feature-discovery
@@ -39,15 +45,23 @@ node-feature-discovery:
     config:
       extraLabelNs: ["nvidia.com"]
   gc:
+    # -- Specifies whether the NFD Garbage Collector should be created
     enable: true
+    # -- Specifies the number of replicas for the NFD Garbage Collector
     replicaCount: 1
     serviceAccount:
-      # disable creation to avoid duplicate serviceaccount creation by master spec above
+      # -- The name of the service account for garbage collector to use.
+      # If not set and create is true, a name is generated using the fullname
+      # template and -gc suffix.
       name: node-feature-discovery
+      # -- disable creation to avoid duplicate serviceaccount creation by master
+      # spec above.
       create: false
+  # -- NFD worker daemonset configuration.
   worker:
     serviceAccount:
-      # disable creation to avoid duplicate serviceaccount creation by master spec above
+      # disable creation to avoid duplicate serviceaccount creation by master spec
+      # above
       name: node-feature-discovery
       create: false
     tolerations:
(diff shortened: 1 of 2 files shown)


hack/templates/values/values.template
```diff
@@ -122,35 +139,46 @@ sriov-network-operator:
       #       ...
       #      -----END EC PRIVATE KEY-----
 
-  # Image URIs for sriov-network-operator components
+  # Image URIs for sriov-network-operator components.
   images:
     operator: {{ .SriovNetworkOperator.Repository }}/{{ .SriovNetworkOperator.Image }}:{{ .SriovNetworkOperator.Version }}
     sriovConfigDaemon: {{ .SriovConfigDaemon.Repository }}/{{ .SriovConfigDaemon.Image }}:{{ .SriovConfigDaemon.Version }}
+    # -- For ARM-based deployments, it is recommended to use the
+    # ``ghcr.io/k8snetworkplumbingwg/sriov-cni:v2.8.0-arm64`` image.
```
rollandf, line 147: We should not have specific versions in the template.
@e0ne is this doc still relevant for next version or we have multi-arch images already?

rollandf (commented):
Looks good. Small question about specific version in some comments.

hack/templates/values/values.template
souleb: Good point. Waiting for @e0ne answer. If we still need that comment, I'll remove the tag.

hack/templates/values/values.template
```diff
@@ -17,37 +17,51 @@
...
 
 sriovNetworkOperator:
+  # -- Deploy SR-IOV Network Operator.
   enabled: false
 
-# Node Feature discovery chart related values
+# Set both enableNodeFeatureApi and NodeFeatureAPI feature gate to false to disable.
 node-feature-discovery:
-  # set both enableNodeFeatureApi and NodeFeatureAPI feature gate to false to disable
+  # -- The Node Feature API is used to expose node-specific hardware and
```
adrianchiris, line 34: this is incorrect IMO. that just an internal mechanism how NFD works.

when node feature API is enabled communication between nfd master and worker is done through NodeFeature CRs
otherwise through gRPC.

NFD will work the same when this is enabled or disabled. (in the future grpc commnication mode will be removed)

hack/templates/values/values.template
adrianchiris: its all muti-arch today.

sriov. ib-sriov cni , device plugin ,

hack/templates/values/values.template
```diff
@@ -228,53 +280,76 @@ ofedDriver:
   #       cpu: "300m"
   #       memory: "300Mi"
   terminationGracePeriodSeconds: 300
```
adrianchiris, line 282: do we need to add doc comment for it or its obvious ?

hack/templates/values/values.template
```diff
@@ -228,53 +280,76 @@ ofedDriver:
...
+      # An empty selector means all pods.
       podSelector: ""
-      # It's recommended to set a timeout to avoid infinite drain in case non-fatal error keeps happening on retries
+      # -- It's recommended to set a timeout to avoid infinite drain in case
+      # non-fatal error keeps happening on retries.
       timeoutSeconds: 300
+      # -- Delete pods local storage.
       deleteEmptyDir: true
     waitForCompletion:
       # specifies a label selector for the pods to wait for completion
```
adrianchiris, line 332: need to add `--` here ?

hack/templates/values/values.template
```diff
@@ -309,14 +393,20 @@ sriovDevicePlugin:
   #     limits:
   #       cpu: "150m"
   #       memory: "100Mi"
+  # -- Optional `resource requests and limits <https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/>`_
```
adrianchiris, line 396: the resources below actually defines the device plugin configuration. its not requests and limits for the sriovdp-container.

hack/templates/values/values.template
souleb: Thanks for explaining this. I will update it.

hack/templates/values/values.template
souleb: That's not needed because the field is commented.

deployment/network-operator/values.yaml
```diff
@@ -309,14 +386,19 @@ sriovDevicePlugin:
   #     limits:
   #       cpu: "150m"
   #       memory: "100Mi"
+  # -- Optional resources configuration for the ``kube-sriovdp`` container.
```
rollandf, line 389: Maybe something similar as the RDMA above?

# -- The following defines the RDMA resources in the cluster.
  # It must be provided by the user when deploying the chart.
  # Each entry in the resources element will create a resource with the provided
  # <name> and list of devices.

adrianchiris (dismissed):

deployment/network-operator/values.yaml
souleb: I updated the description. In practices I didn't succeed in using the `sriovDevicePlugin` fields, so we will have to revisit this I think.

rollandf: Thanks, let's get this in and we will compare with published version

rollandf (approved):

PR #1013, opened 2024-07-29
chore: backport a pre-delete helm hook for sriov-network-operator

almaslennikov:
Mutating and Validating webhook configs are created programmatically and are not managed by helm.
To clean them up on helm delete, adding a small pre-delete hook
Backport of: https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/751

### DIFF
--- a/deployment/network-operator/charts/sriov-network-operator/templates/predeletehook.yaml
+++ b/deployment/network-operator/charts/sriov-network-operator/templates/predeletehook.yaml
@@ -1,0 +1,24 @@
+{{ if .Values.operator.admissionControllers.enabled }}
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: "{{ include "sriov-network-operator.fullname" . }}-post-delete-hook"
+  namespace: {{ .Release.Namespace }}
+  annotations:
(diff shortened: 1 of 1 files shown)


deployment/network-operator/charts/sriov-network-operator/templates/predeletehook.yaml
```diff
@@ -0,0 +1,24 @@
...
+    spec:
+      serviceAccountName: {{ include "sriov-network-operator.fullname" . }}
+      containers:
+        - name: cleanup
+          image: bitnami/kubectl:latest
+          command:
+            - /bin/sh
+            - -c
+            - |
+              kubectl delete MutatingWebhookConfiguration -n {{ .Release.Namespace }} sriov-operator-webhook-config network-resources-injector-config
```
rollandf, line 21: What happens if these CRs are not available? Will the Job fail?
Do we need `;` at the end?
Maybe add ` || true;`

rollandf:
I think we can close this one.
Covered in https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/791

PR #1033, opened 2024-08-15
fix: Add openshift rolebinding to DOCA Telemetry Service

killianmuldoon:
Add support for OpenShift to the DOCA Telemetry Services deployment by adding a service account associated roles / rolebinding which gives access to the privileged securityContextConstraint.

### DIFF
--- a/pkg/state/common_test.go
+++ b/pkg/state/common_test.go
@@ -385,3 +385,17 @@ func runFuncForObjectInSlice(objects []*unstructured.Unstructured, objectKind st
 	}
 	return found
 }
+
+// assertUnstructuredListHasKinds will check that there is an object in the list for each Kind in `kinds`.
+// kinds may have duplicates and the length of kinds must exactly match the length of objects.
+// If multiple objects have the same kind it should be in the `kinds` list multiple times.
+func assertUnstructuredListHasExactKinds(objects []*unstructured.Unstructured, kinds []string) {
+	Expect(len(objects)).To(Equal(len(kinds)))
+	maps := []map[string]interface{}{}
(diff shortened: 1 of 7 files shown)


killianmuldoon:
@e0ne does this need to be backported?

rollandf:
> @e0ne does this need to be backported?

No release branch yet, so no need

rollandf (dismissed):

pkg/state/common_test.go
```diff
@@ -385,3 +385,17 @@ func runFuncForObjectInSlice(objects []*unstructured.Unstructured, objectKind st
 	}
 	return found
 }
+
+// assertUnstructuredListHasKinds will check that there is an object in the list for each Kind in `kinds`.
+// kinds may have duplicates and the length of kinds must exactly match the length of objects.
+// If multiple objects have the same kind it should be in the `kinds` list multiple times.
```
adrianchiris, line 391: will this behave as the the description?

say:

`assertUnstructuredListHasExactKinds( [roleObj, daemonsetObj] , ["role", "role"])`

the loop at L398 will always hit the map generated for the first obj and not fail.

killianmuldoon: You're right for sure - it's not trivial to get rid of the map here though. I'll take some other approach

killianmuldoon: Should be better now - and I think it's a lot simpler to reason about.

adrianchiris: much better, thx ! :)

adrianchiris (approved):

PR #1034, opened 2024-08-15
Feat: Add OCP bundle to CICD workflow

maze88:
Changes overview:
- Renames the `network-operator-docker-and-helm-ci.yaml` workflow to`ci.yaml`; workflow contains three core jobs:
  1. _docker-build-push_
  2. _helm-package-publish_
  3. _ocp-bundle_
- Adds job for OCP bundling and delivery:
  - Runs only when git _tags_ are pushed.
  - Makes the OCP bundle.
  - Creates a PR to _network-operator_, to update the local copy of the bundle.
  - Creates a PR to RedHat's [_certified-operators_](https://github.com/redhat-openshift-ecosystem/certified-operators) (only for GA releases).
  
(EDIT: corrected name again - from `cicd.yaml` to `cd.yaml`)


rollandf:
Please sign all commits.
In general better use `git mv` in order to rename, it is hard to see the changes.
Or rename in a different commit.

.github/workflows/cicd.yaml
```diff
@@ -0,0 +1,192 @@
...
+      run: |
+        pushd certified-operators
+
+        git config user.name  nvidia-ci-cd
+        git config user.email svc-cloud-orch-gh@nvidia.com
+        gh repo fork --remote --default-branch-only
+        gh repo sync $DOWNSTREAM_REPO_OWNER/certified-operators --source $UPSTREAM_REPO_OWNER/certified-operators --branch $UPSTREAM_DEFAULT_BRANCH
+
+        git checkout -b $FEATURE_BRANCH
+        mkdir -p $NEW_BUNDLE_DIR
```
rollandf, line 180: Where is `$NEW_BUNDLE_DIR` defined?

.github/workflows/cicd.yaml
```diff
@@ -0,0 +1,192 @@
...
+        git add bundle
+        git commit -m "task: update bundle to $VERSION_WITH_PREFIX"
+
+        git push -u origin $FEATURE_BRANCH
+        gh pr create \
+          --head $FEATURE_BRANCH \
+          --base $DEFAULT_BRANCH \
+          --title "task: update bundle to $VERSION_WITH_PREFIX" \
+          --body "Created by the *${{ github.job }}* job in [${{ github.repository }} OCP bundle CI](${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }})."
+        # gh pr merge --admin --merge $FEATURE_BRANCH  # requires repository settings to "allow auto-merge"  # TODO: only auto-merge once proper tests are implemented
```
rollandf, line 155: Remove this line.
Even when we will have tests and decide to auto-merge, it will not be merged here, but in a cron job checking if tests passed before merging.

.github/workflows/cicd.yaml
```diff
@@ -0,0 +1,192 @@
...
+        echo VERSION_WITHOUT_PREFIX=${git_tag:1} >> $GITHUB_ENV  # without the 'v' prefix
+    - name: Lookup image digest
+      run: |
+        network_operator_digest=$(skopeo inspect docker://$REGISTRY/$IMAGE_NAME:$VERSION_WITH_PREFIX | jq -r .Digest)
+        echo $network_operator_digest | wc -w | grep 1  # verifies value not empty
+        echo NETWORK_OPERATOR_DIGEST=$network_operator_digest >> $GITHUB_ENV
+    - name: Make bundle
+      env:
+        DEFAULT_CHANNEL: ${{ env.VERSION_WITH_PREFIX }}
+        CHANNELS: stable,${{ env.VERSION_WITH_PREFIX }}
```
rollandf, line 130: The channels should `stable` and the version with vMAJOR.MINOR only.
e.g. v24.7 and not v24.7.0

maze88: That makes sense, however it is not how it appears to have been managed historically ([for example in 24.4.1](https://github.com/maze88/certified-operators/blob/main/operators/nvidia-network-operator/24.4.1/metadata/annotations.yaml#L7-L8)).
Perhaps @e0ne can settle this?

.github/workflows/cicd.yaml
maze88: In this step's `env:`.

.github/workflows/cicd.yaml
adrianchiris: channels should be managed as @rollandf mentioned above.
that way patch version of the same vMAJOR.MINOR end up in the same channel.

maze88: Thanks for confirming, I've corrected this.

maze88:
> Please sign all commits. In general better use `git mv` in order to rename, it is hard to see the changes. Or rename in a different commit.
> 
> Also you have here not relevant commits.

For review convenience, here is an easier to view diff - that will ignore the file rename. It was generated with the changes as of 26/8: https://www.diffchecker.com/W0TqMiqh/

e0ne (approved):
LGTM

PR #1048, opened 2024-08-29
chore: Bump DOCA Driver and RDMA Shared Device Plugin

e0ne:


rollandf (approved):

rollandf (approved):

rollandf:
`image_scan CI` is not related to this change.
I am OK with merging

Issue #1052, opened 2024-09-01
Error pulling v24.7.0 container image using helm chart (Authorization required?)


drikster80:
**Environment**:
- 4-node GH200 Cluster (Vanilla Kubernetes 1.31)
  - Bluefield-3 DPU in each node & SN3700 Switch (Cumulus Linux)
- GPU Operator Deployed

Attempting to deploy Network Operator v24.7.0 with command:
```
helm show values nvidia/network-operator --version v24.7.0 > values.yaml

# Updated to disable nfd since it was deployed with GPU Operator (`nfd.enabled=false`)

helm install network-operator nvidia/network-operator -n nvidia-network-operator --create-namespace --version v24.7.0 -f ./values.yaml --wait
```

**What happened**:

`ErrImagePull`

Error showing in containerd:
```
Sep 01 18:01:24 gh200-1 containerd[1768487]: time="2024-09-01T18:01:24.687567471Z" level=info msg="PullImage \"nvcr.io/nvstaging/mellanox/network-operator:v24.7.0\""
Sep 01 18:01:25 gh200-1 containerd[1768487]: time="2024-09-01T18:01:25.333471500Z" level=error msg="PullImage \"nvcr.io/nvstaging/mellanox/network-operator:v24.7.0\" failed" error="failed to pull and unpack image \"nvcr.io/nvstaging/mellanox/network-operator:v24.7.0\": failed to resolve reference \"nvcr.io/nvstaging/mellanox/network-operator:v24.7.0\": unexpected status from HEAD request to https://nvcr.io/v2/nvstaging/mellanox/network-operator/manifests/v24.7.0: 401 Unauthorized"
Sep 01 18:01:25 gh200-1 containerd[1768487]: time="2024-09-01T18:01:25.333522412Z" level=info msg="stop pulling image nvcr.io/nvstaging/mellanox/network-operator:v24.7.0: active requests=0, bytes read=1068"
```

**What you expected to happen**:

The images should be pulled without credentials.

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

Should it be pulling from 'nvstaging'? (`nvcr.io/nvstaging/mellanox/network-operator:v24.7.0`)

Here is the result when attempting to pull with docker (on both nvstaging and nvidia):
```
docker pull nvcr.io/nvstaging/mellanox/network-operator:v24.7.0
Error response from daemon: unauthorized: <html>
<head><title>401 Authorization Required</title></head>
<body>
[...]

PR #1055, opened 2024-09-02
bug: fix helm values generation

rollandf:
Related to #1052

### DIFF
--- a/hack/templates/values/values.template
+++ b/hack/templates/values/values.template
@@ -201,9 +201,9 @@ operator:
                 operator: In
                 values: [ "" ]
   # -- Network Operator image repository.
-  repository: nvcr.io/nvstaging/mellanox
+  repository: {{ .NetworkOperator.Repository }}
   # -- Network Operator image name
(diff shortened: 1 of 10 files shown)


ykulazhenkov (approved):

adrianchiris (approved):

hack/templates/values/values.template
```diff
@@ -201,7 +201,7 @@ operator:
                 operator: In
                 values: [ "" ]
   # -- Network Operator image repository.
-  repository: nvcr.io/nvstaging/mellanox
+  repository: {{ .NetworkOperator.Repository }}
   # -- Network Operator image name
   image: network-operator
```
tobiasgiese, line 206:
```suggestion
  image: {{ .NetworkOperator.Image }}
```

We should change this also back to the values

rollandf: Done

tobiasgiese (approved):

souleb (approved):
lgtm!

PR #1058, opened 2024-09-03
chore: bundle for 24.7.1

rollandf:
- Add nework operator image to related images
- Use multi-arch SHA

Generated by:
DEFAULT_CHANNEL=v24.7.0 CHANNELS=v24.7.0,stable VERSION=24.7.1 TAG=nvcr.io/nvidia/cloud-native/network-operator@sha256:b8a07696d05fb8e5292991ba731758579313d6983ed3d75dd95e7356024d987c make bundle


e0ne:
@rollandf let's check it using `certified-operators` repo CI before merge

bundle.Dockerfile
```diff
@@ -5,8 +5,8 @@ LABEL operators.operatorframework.io.bundle.mediatype.v1=registry+v1
 LABEL operators.operatorframework.io.bundle.manifests.v1=manifests/
 LABEL operators.operatorframework.io.bundle.metadata.v1=metadata/
 LABEL operators.operatorframework.io.bundle.package.v1=nvidia-network-operator
-LABEL operators.operatorframework.io.bundle.channels.v1=v24.7.0-rc.1,stable
-LABEL operators.operatorframework.io.bundle.channel.default.v1=v24.7.0-rc.1
+LABEL operators.operatorframework.io.bundle.channels.v1=v24.7.0,stable
```
adrianchiris, line 8: for 24.10 will we have "v24.10,stable" as channel.v1 value ?

bundle.Dockerfile
```diff
@@ -5,8 +5,8 @@ LABEL operators.operatorframework.io.bundle.mediatype.v1=registry+v1
...
+LABEL operators.operatorframework.io.bundle.channel.default.v1=v24.7.0
```
adrianchiris, line 9: same comment as above

adrianchiris (approved):

bundle.Dockerfile
rollandf: Yes.
Should be handle in workflow:
https://github.com/Mellanox/network-operator/pull/1034/files#diff-944291df2c9c06359d37cc8833d182d705c9e8c3108e7cfe132d61a06e9133ddR131

rollandf: Do you think we should have here "v24.7.1", like we did in "v24.4.1"?
https://github.com/redhat-openshift-ecosystem/certified-operators/blob/main/operators/nvidia-network-operator/24.4.1/metadata/annotations.yaml#L8

e0ne (changes requested):
CSV should be updated to to include labels including arm64 support https://github.com/NVIDIA/gpu-operator/blob/420c023825c0e7a192fd5941ebb32ed1ea76fa62/bundle/v24.6.1/manifests/gpu-operator-certified.clusterserviceversion.yaml#L5-L20

e0ne (approved):
let's push it to certified-operators repo to check with RedHat CI before merging it

PR #1063, opened 2024-09-08
feat: add check to prevent releases with "latest" version dependencies

maze88:


Makefile
```diff
@@ -408,6 +408,7 @@ bundle-push: ## Push the bundle image.
 
 .PHONY: release-build
 release-build:
+	yq '.[].version'  hack/release.yaml | grep 'latest' && exit 1
```
rollandf, line 411: grep return with `1` if no matches. No need for: ` && exit 1`

maze88: But if we *find* 'latest' we *want* it to fail...

rollandf: right, you still need to add `|| true` at the end

`yq '.[].version' hack/release.yaml | grep -q 'latest' && exit 1 || true`

maze88: Good catch! Fixed.

rollandf (approved):

PR #1065, opened 2024-09-08
bug: add format hint to helm docs

rollandf:
In order to render complex objects as yaml,
add a type hint for helm-docs


e0ne (approved):
LGTM

PR #1073, opened 2024-09-18
bugfix: ensure that we don't panic on DP validations

vasrem:
Before, if a user wouldn't specify config for any of the rdma shared DP or sriov DP, the process would panic. This is now fixed.

Also, split a function into 2 to pass the linter.


almaslennikov (approved):
LGTM once CI passes

rollandf (approved):

PR #1076, opened 2024-09-18
fix: OCP bundle channels definition

maze88:


.github/workflows/ci.yaml
```diff
@@ -140,7 +140,7 @@ jobs:
         git checkout -b $FEATURE_BRANCH
         git status
         git add bundle
-        git commit -m "task: update bundle to $VERSION_WITH_PREFIX"
+        git commit -Sm "task: update bundle to $VERSION_WITH_PREFIX"
```
rollandf, line 143: I think we should use `s` not `S`

.github/workflows/docs-ci-trigger.yaml
```diff
@@ -9,5 +9,6 @@ jobs:
   trigger_downstream_workflow:
     uses: Mellanox/network-operator-docs/.github/workflows/docs-ci.yaml@main
     with:
+      token: ${{ secrets.GH_TOKEN_NVIDIA_CI_CD }}
```
rollandf, line 12: This is already in a different PR

rollandf: You still have 2 commits in this PR

rollandf (approved):

PR #1082, opened 2024-09-22
bug: fix doc workflow

rollandf:
Invalid workflow file: .github/workflows/docs-ci-trigger.yaml#L12

The workflow is not valid. .github/workflows/docs-ci-trigger.yaml
 (Line: 12, Col: 14): Unrecognized named-value: 'secrets'.
Located at position 1 within expression: secrets.GH_TOKEN_NVIDIA_CI_CD


rollandf:
When I tried to re-run the earlier version, it runs without this error:
https://github.com/Mellanox/network-operator/actions/runs/10921479665/job/30363895261

It failed on a make target in doc repo. (see https://github.com/Mellanox/network-operator-docs/pull/108)

PR #1085, opened 2024-09-27
fix: Remove Helm deployment of NICClusterPolicy

killianmuldoon:
Remove the ability to create the NICClusterPolicy CR through helm.values. This feature has been deprecated for a number of releases and the documentation was updated to remove reference to this feature in the most recent release.

Users should instead use the helm package to install the Network Operator and configure the system by creating the NICClusterPolicy in a separate step.


deployment/network-operator/templates/tests/test-rping.yaml
```diff
@@ -1,49 +0,0 @@
-{{- if and .Values.deployCR .Values.rdmaSharedDevicePlugin .deploy .Values.secondaryNetwork.deploy .Values.secondaryNetwork.multus.deploy .Values.secondaryNetwork.cniPlugins.deploy }}
```
killianmuldoon, line 1: Were these tests being used anywhere? They won't work after this change, but I'm wondering if this change means some loss in testing coverage.

almaslennikov (approved):

e0ne (approved):

rollandf:
Also, docs repo will need update of https://github.com/Mellanox/network-operator-docs/blob/main/templates/helm.rst.gotmpl

PR #1098, opened 2024-10-14
Fixing empty sriov/rdma device plugin empty config tests

heyvister1:
Changed empty rdma/sriov device plugin config test case to be invalid, and fixed validation function to return a non nil error


adrianchiris:
what is this attempting to fix ?

a nil/empty config allows the config map to be provided externally.

heyvister1:
Closing the PR: nil conifg map is a valid case for when the config map is created externally (not through NCP)

PR #1109, opened 2024-10-21
chore: Use bitnami docker image for kubectl

souleb:
This will enable having dependabot automatic updates for docker images.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,13 +19,9 @@ ARG ARCH
 FROM golang:1.23@sha256:540d3442f4758da82e787d03930bf6468cf6f8613474135e8946259b9e531ea0 AS builder
 
 WORKDIR /workspace
-# Add kubectl tool
+
 # Using the $ARCH in the name of the binary here ensures we don't get any cross-arch caching after this binary is downloaded.
 ARG ARCH
-# kubectl latest version can be retrieved by curl -L -s https://dl.k8s.io/release/stable.txt
-ARG KUBECTL_VERSION=v1.31.1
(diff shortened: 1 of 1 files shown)


almaslennikov (approved):

Dockerfile
```diff
@@ -51,15 +47,19 @@ RUN --mount=type=cache,target=/go/pkg/mod \
...
 
 ARG ARCH
 
 WORKDIR /
 COPY --from=builder /workspace/manager .
-COPY --from=builder /workspace/kubectl-${ARCH} /usr/local/bin/kubectl
 COPY --from=builder /workspace/crds /crds
 
+# Copy the kubectl binary
+COPY --from=kubectl /opt/bitnami/kubectl/bin/kubectl /usr/local/bin/
```
e0ne, line 61: We use UBI base image. bitnami/kubectl is debian-based. I don't think it's a good option to mix OS'ws

e0ne (changes requested):

Dockerfile
souleb: Can you tell more about that?

Here we just copy the Kubectl binary. Kubectl is written in Golang, so it's self-contained. Did I miss anything?

e0ne: if we use only one binary it's ok

e0ne (approved):

adrianchiris (approved):
LGTM.

@e0ne  BTW didnt we want to switch to a client-go binary to avoid using kubectl ?

e0ne:
We don't need this once #1118 will be merged

PR #1115, opened 2024-10-29
fix: Add 'nodeSelector' field for DTS deployment

e0ne:
We need to run it after DOCA driver is started to be able successfully restart driver by a container.

### DIFF
--- a/manifests/state-doca-telemetry-service/0010-doca-telemetry-service.yaml
+++ b/manifests/state-doca-telemetry-service/0010-doca-telemetry-service.yaml
@@ -18,6 +18,10 @@ spec:
       # Required by the sysfs collector only.
       hostNetwork: true
       serviceAccount: doca-telemetry-service
+      nodeSelector:
+        # We need to run it after DOCA driver is started to be able successfully restart driver by a container.
+        feature.node.kubernetes.io/pci-15b3.present: "true"
+        network.nvidia.com/operator.mofed.wait: "false"
       {{- if .NodeAffinity }}
       affinity:
         nodeAffinity:


adrianchiris (dismissed):

manifests/state-doca-telemetry-service/0010-doca-telemetry-service.yaml
```diff
@@ -18,6 +18,9 @@ spec:
       # Required by the sysfs collector only.
       hostNetwork: true
       serviceAccount: doca-telemetry-service
+      nodeSelector:
+        feature.node.kubernetes.io/pci-15b3.present: "true"
+        network.nvidia.com/operator.mofed.wait: "false"
```
adrianchiris, line 23: can you add a comment why this label is needed for posterity :) ?

e0ne: > can you add a comment why this label is needed for posterity :) ?

is'n it enough to have it in a commit message only?

rollandf, line 23: Will the label exist if MOFED is not configured in NCP?

e0ne: operator always creates this label

rollandf (approved):

manifests/state-doca-telemetry-service/0010-doca-telemetry-service.yaml
e0ne: done

rollandf (approved):

vasrem (approved):

Issue #1072, opened 2024-09-18
Use Go code instead of kubectl binary


rollandf:
`kubectl` binary is used in Helm hooks.
We should consider to migrate to use Go code to avoid CVEs on `kubectl` that might not be updated in a timely manner.

PR #1118, opened 2024-10-29
feat: Remove kubectl binary from container and apply CRDs with client-go

tobiasgiese:
Due too multiple CVEs in the kubectl binary we remove the dependency to it.
To be able to update our CRDs we introduce a custom Go binary that uses client-go
to apply our CRDs and also from our dependencies, like NFD and SRIOV.

Additionally we remove the scale-down Helm hook as it is no longer needed.

Fixes https://github.com/Mellanox/network-operator/issues/1072


cmd/apply-crds/main.go
```diff
@@ -0,0 +1,163 @@
...
+// applyCRDsFromFile reads a YAML file, splits it into documents, and applies each CRD to the cluster.
+func applyCRDsFromFile(ctx context.Context, client *clientset.Clientset, filePath string) error {
+	file, err := os.Open(filePath)
+	if err != nil {
+		return fmt.Errorf("open file %q: %v", filePath, err)
+	}
+	defer file.Close()
+
+	// Create a decoder that reads multiple YAML documents.
+	decoder := yaml.NewYAMLOrJSONDecoder(file, 4096)
```
souleb, line 100: I believe using a `reader` is more efficient.

```go
	reader := yaml.NewYAMLReader(bufio.NewReader(file))

	var crdsToApply []*apiextensionsv1.CustomResourceDefinition
	for {
		doc, err := reader.Read()
		if err != nil {
			if err == io.EOF {
				break
			}
		}
		crd := &apiextensionsv1.CustomResourceDefinition{}
		decoder := yaml.NewYAMLOrJSONDecoder(bytes.NewBuffer(doc), len(doc))
		if err := decoder.Decode(crd); err != nil {
			return fmt.Errorf("decode YAML: %v", err)
		}
		crdsToApply = append(crdsToApply, crd)
	}
```

`doc, err := reader.Read()`  this will give you a full Yaml document instead of limiting it to the buffer size.

cmd/apply-crds/suite_test.go
```diff
@@ -0,0 +1,65 @@
...
+
+import (
+	"testing"
+
+	"k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
+	"sigs.k8s.io/controller-runtime/pkg/envtest"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+
+	. "github.com/onsi/ginkgo/v2"
```
souleb, line 27: If we plan to release this as a lib, would you consider not using `gingkgo` to reduce the dependencies to the strict minimum?

tobiasgiese: If we are going to move this to the lib repo I will not use ginkgo, yes. It just because this repo is using ginkgo and that's why I wanted to use it here as well

cmd/apply-crds/main.go
tobiasgiese: Good point, will implement it 👍🏻 thanks

tobiasgiese: Tested it locally with 20.000 CRD updates. Both took about 44s, so no performance optimization unfortunately.
I will keep it as is

cmd/apply-crds/main.go
```diff
@@ -0,0 +1,163 @@
...
+		crd.SetResourceVersion(curCRD.GetResourceVersion())
+		_, err = crdClient.Update(ctx, crd, metav1.UpdateOptions{})
+		if err != nil {
+			return fmt.Errorf("update CRD %s: %v", crd.Name, err)
+		}
+	}
+	return nil
+}
+
+func getKubeConfig() (*rest.Config, error) {
```
vasrem, line 152: I think we can replace that with https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/client/config#GetConfig, unless we don't want to include controller runtime here

deployment/network-operator/templates/upgrade-crd-hook.yaml
```diff
@@ -84,12 +84,12 @@ spec:
             - /bin/sh
             - -c
             - >
-              kubectl apply -f /crds/network-operator;
+              /apply-crds --crds-dir=/crds/network-operator;
```
vasrem, line 87: nit: maybe we should adjust the script to take a list of dirs instead and split this in `command` and `args`

Dockerfile
```diff
@@ -38,30 +37,48 @@ RUN --mount=type=cache,target=/go/pkg/mod \
 # Copy the go source
 COPY ./ ./
 
+# Build
+RUN --mount=type=cache,target=/go/pkg/mod \
+    --mount=type=cache,target=/root/.cache/go-build \
+    CGO_ENABLED=0 GOOS=linux GOARCH=${ARCH} go build -ldflags="${LDFLAGS}" -gcflags="${GCFLAGS}" -o manager main.go
```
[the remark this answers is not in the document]
tobiasgiese: But this is rather considered as a general issue with the build, not PR related, right? If so I would address this as a follow-up to not blowup the PR. Also it will be more visible.

Dockerfile
```diff
@@ -38,30 +37,48 @@ RUN --mount=type=cache,target=/go/pkg/mod \
...
+    cp -r network-operator-chart/charts/sriov-network-operator/crds /workspace/crds/sriov-network-operator/ && \
+    cp -r network-operator-chart/charts/node-feature-discovery/crds /workspace/crds/node-feature-discovery/
 
 # Build
-ARG LDFLAGS
-ARG GCFLAGS
 RUN --mount=type=cache,target=/go/pkg/mod \
     --mount=type=cache,target=/root/.cache/go-build \
-    CGO_ENABLED=0 GOOS=linux GOARCH=${ARCH} go build -ldflags="${LDFLAGS}" -gcflags="${GCFLAGS}" -o manager main.go
+    CGO_ENABLED=0 GOOS=linux GOARCH=${ARCH} go build -ldflags="${LDFLAGS}" -gcflags="${GCFLAGS}" -o apply-crds main.go
```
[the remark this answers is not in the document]
tobiasgiese: https://github.com/Mellanox/network-operator/pull/1118#discussion_r1824045698

cmd/apply-crds/main.go
tobiasgiese: I haven't used the controller-runtime dep on purpose. Usually we only need `rest.InClusterConfig()`, only for dev purposes we need the fallback.

Dockerfile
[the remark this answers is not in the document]
vasrem: You removed a couple of ARGS that I believe break the image.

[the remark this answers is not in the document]
tobiasgiese: I haven't removed them, I just moved them to the top. AFAIK the ARGs are for the complete Dockerfile, not only for the certain stages. If I was right I will revert it.

cmd/apply-crds/main.go
vasrem: The `go.mod` has `sigs.k8s.io/controller-runtime v0.19.1` as a dep today, is this expected?

Dockerfile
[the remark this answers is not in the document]
vasrem: I'm not saying you are wrong, I'm just asking for a verification to ensure we don't break the docker image. If you believe this should be done in a next PR, I'm fine with that.

cmd/apply-crds/main.go
tobiasgiese: Good call, the dep came after the implementation of the package. We have the controller-runtime dependency because of the envtest usage.
I think it's fair to use this instead then 👍🏻

cmd/apply-crds/main.go
```diff
@@ -0,0 +1,163 @@
...
+	// Create a decoder that reads multiple YAML documents.
+	decoder := yaml.NewYAMLOrJSONDecoder(file, 4096)
+	var crdsToApply []*apiextensionsv1.CustomResourceDefinition
+	for {
+		crd := &apiextensionsv1.CustomResourceDefinition{}
+		if err := decoder.Decode(crd); err != nil {
+			if err == io.EOF {
+				break
+			}
+			return fmt.Errorf("decode YAML: %v", err)
```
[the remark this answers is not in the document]
tobiasgiese: Having empty yamls is not the issue, but this will produce an issue

```
---
kind: CustomResourceDefinition
---

---
kind: CustomResourceDefinition
```

Fixed it in the code and updated also the test yaml accordingly to cover this

deployment/network-operator/templates/upgrade-crd-hook.yaml
tobiasgiese: I'll take a look at this, good idea 👍🏻

Dockerfile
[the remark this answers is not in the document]
tobiasgiese: I can move the necessary ARGs back to the usage to make this more clear. The verification of the images can be done as a follow-up then. WDYT?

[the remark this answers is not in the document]
tobiasgiese: It is also stated in the documentation that it is only for the current scope

> An `ARG` instruction goes out of scope at the end of the build stage where it was defined. To use an argument in multiple stages, each stage must include the `ARG` instruction.

https://docs.docker.com/reference/dockerfile/#scope

deployment/network-operator/templates/upgrade-crd-hook.yaml
tobiasgiese: Implemented and tested locally in a kind cluster

```
+ network-operator-upgrade-crd-r5hmb › upgrade-crd
network-operator-upgrade-crd-r5hmb upgrade-crd 2024/10/31 09:29:32 Apply CRDs from file: /crds/network-operator/k8s.cni.cncf.io_networkattachmentdefinitions_crd.yaml
network-operator-upgrade-crd-r5hmb upgrade-crd 2024/10/31 09:29:32 Update CRD network-attachment-definitions.k8s.cni.cncf.io
network-operator-upgrade-crd-r5hmb upgrade-crd 2024/10/31 09:29:32 Apply CRDs from file: /crds/network-operator/mellanox.com_hostdevicenetworks.yaml
network-operator-upgrade-crd-r5hmb upgrade-crd 2024/10/31 09:29:32 Update CRD hostdevicenetworks.mellanox.com
network-operator-upgrade-crd-r5hmb upgrade-crd 2024/10/31 09:29:32 Apply CRDs from file: /crds/network-operator/mellanox.com_ipoibnetworks.yaml
network-operator-upgrade-crd-r5hmb upgrade-crd 2024/10/31 09:29:32 Update CRD ipoibnetworks.mellanox.com
network-operator-upgrade-crd-r5hmb upgrade-crd 2024/10/31 09:29:32 Apply CRDs from file: /crds/network-operator/mellanox.com_macvlannetworks.yaml
network-operator-upgrade-crd-r5hmb upgrade-crd 2024/10/31 09:29:32 Update CRD macvlannetworks.mellanox.com
network-operator-upgrade-crd-r5hmb upgrade-crd 2024/10/31 09:29:32 Apply CRDs from file: /crds/network-operator/mellanox.com_nicclusterpolicies.yaml
network-operator-upgrade-crd-r5hmb upgrade-crd 2024/10/31 09:29:32 Update CRD nicclusterpolicies.mellanox.com
network-operator-upgrade-crd-r5hmb upgrade-crd 2024/10/31 09:29:32 Apply CRDs from file: /crds/node-feature-discovery/nfd-api-crds.yaml
network-operator-upgrade-crd-r5hmb upgrade-crd 2024/10/31 09:29:32 Update CRD nodefeatures.nfd.k8s-sigs.io
network-operator-upgrade-crd-r5hmb upgrade-crd 2024/10/31 09:29:32 Update CRD nodefeaturerules.nfd.k8s-sigs.io
- network-operator-upgrade-crd-r5hmb › upgrade-crd
```

```
    Command:
      /apply-crds
    Args:
      --crds-dir=/crds/network-operator
      --crds-dir=/crds/node-feature-discovery
```

vasrem (approved):
LGTM

tobiasgiese:
Had to [rollback the docker image changes](https://github.com/Mellanox/network-operator/compare/c1227ad131f04e347b90ad4bd095ae870dc08de5..76abbe38b4b0a519a19d3c5889085613dc1f85cb) because of https://github.com/dependabot/dependabot-core/issues/2057

Dockerfile
```diff
@@ -38,30 +31,54 @@ RUN --mount=type=cache,target=/go/pkg/mod \
...
+COPY deployment/network-operator/ ./network-operator-chart/
+
 # copy CRDs from helm charts
 RUN mkdir crds && \
-    cp -r deployment/network-operator/crds /workspace/crds/network-operator/ && \
-    cp -r deployment/network-operator/charts/sriov-network-operator/crds /workspace/crds/sriov-network-operator/ && \
-    cp -r deployment/network-operator/charts/node-feature-discovery/crds /workspace/crds/node-feature-discovery/
+    cp -r network-operator-chart/crds /workspace/crds/network-operator/ && \
+    cp -r network-operator-chart/charts/sriov-network-operator/crds /workspace/crds/sriov-network-operator/ && \
+    cp -r network-operator-chart/charts/node-feature-discovery/crds /workspace/crds/node-feature-discovery/
```
rollandf, line 63: we are missing nic-configuration-operator-chart/crds

deployment/network-operator/templates/upgrade-crd-hook.yaml
```diff
@@ -81,15 +81,14 @@ spec:
...
+            - /apply-crds
+          args:
+              - --crds-dir=/crds/network-operator
         {{- if .Values.sriovNetworkOperator.enabled }}
-              kubectl apply -f /crds/sriov-network-operator;
+              - --crds-dir=/crds/sriov-network-operator
         {{- end }}
         {{- if .Values.nfd.enabled }}
-              kubectl apply -f /crds/node-feature-discovery;
+              - --crds-dir=/crds/node-feature-discovery
```
rollandf, line 91: we are missing nic-configuration-operator-chart/crds
Though not related to your PR

Dockerfile
tobiasgiese: Fixed and pushed, PTAL

deployment/network-operator/templates/upgrade-crd-hook.yaml
tobiasgiese: If we enable everything it will also apply everything. `crds-dir` is a list and can be called multiple times

tobiasgiese: see https://github.com/Mellanox/network-operator/pull/1118#discussion_r1824144369

rollandf (approved):

e0ne (approved):

PR #1132, opened 2024-11-05
chore: downgrade go version for apply-crds package

tobiasgiese:
There is a project that wants to consume this package.
As this project sticks to the Go version that is also used by the lates stable Kubernetes version we should also downgrade this version to be compatible.

### DIFF
--- a/cmd/apply-crds/go.mod
+++ b/cmd/apply-crds/go.mod
@@ -1,6 +1,6 @@
 module github.com/mellanox/network-operator/cmd/apply-crds
 
-go 1.23.2
+go 1.22.8
 
 require (
 	github.com/onsi/ginkgo/v2 v2.19.0


e0ne:
> There is a project that wants to consume this package.

what project is it? I don't like an idea of downgrading golang and having different version for several components

tobiasgiese:
> what project is it? I don't like an idea of downgrading golang and having different version for several components

It is an internal project. In this project we are pinning the version to the latest stable Kubernetes version -- which is currently [go1.22 for Kubernetes v1.31](https://github.com/kubernetes/kubernetes/blob/release-1.31/go.mod#L9). As the latest go1.22 patch version is go1.22.8 I want to set the go version for this package.
I understand your concerns, but it breaks the usage of this package for projects that stick to Kubernetes' Golang version.

ykulazhenkov (approved):
I don't like the idea of using different go version within the same code tree, but I don't see any good alternative for this case. 

We are downgrading go version for isolated app (it doesn't import anything from the pkg), so this change is acceptable from my point of view. 

I think CVEs are not an issue here, because we are still using the lates go image to build the binary.

tobiasgiese:
We will move the code to https://github.com/NVIDIA/k8s-operator-libs/pull/50 anway, so we can close this PR.
If something is against moving the tool to the libs repo I will re-open this PR.

PR #1135, opened 2024-11-05
bug: keep NCP on Helm upgrade

rollandf:
Creation of NicClusterPolicy was removed from the Helm Chart.

In order to avoid deletion of the CR on Helm upgrade, a Helm hook adds a 'keep' annotation to the NCP if needed, before running the actual upgrade.

### DIFF
--- a/cmd/keep-ncp/main.go
+++ b/cmd/keep-ncp/main.go
@@ -1,0 +1,97 @@
+/*
+2024 NVIDIA CORPORATION & AFFILIATES
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package main adds Helm 'keep' annotation on NCP if needed
+package main
+
+import (
+	"context"
+	"log"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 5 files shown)


cmd/keep-ncp/main.go
```diff
@@ -0,0 +1,97 @@
...
+	if apierrors.IsNotFound(err) {
+		log.Println("NicClusterPolicy does not exists. No annotation needed")
+		return nil
+	}
+	if err != nil {
+		log.Println("Failed to get NicClusterPolicy")
+		return err
+	}
+	labels := ncp.GetLabels()
+	if labels["app.kubernetes.io/managed-by"] != "Helm" {
```
tobiasgiese, line 75: If it's created by Helm, shouldn't we add the annotation directly via Helm instead of this binary?
Or is the idea to enforce this annotation, even if the NCP was added by a user-owned Helm Chart?

cmd/keep-ncp/main.go
```diff
@@ -0,0 +1,97 @@
...
+	if ncp.Annotations == nil {
+		ncp.Annotations = make(map[string]string)
+	}
+	ncp.Annotations[helmResourcePolicyKey] = helmKeepValue
+
+	err = c.Update(ctx, ncp)
+	if err != nil {
+		log.Println("Error updating NicClusterPolicy with 'keep' annotation")
+		return err
+	}
```
tobiasgiese, line 95: Do we need a retry here? If the update fails the complete helm upgrade fails - even if there is a retry mechanism in place AFAIK

rollandf: The Job will restart if fails:
https://github.com/Mellanox/network-operator/pull/1135/files#diff-3c496f10888e7f6d5070ee0a9cbeba5c8f6fc9f41153da44653ca3f5c45e6f58R84

Is it good enough? WDYT?

tobiasgiese: Ah that's good enough, sure 👍🏻

e0ne (approved):

PR #1174, opened 2024-11-20
chore: Use crdutil package for apply-crds from k8s-operator-libs repo

tobiasgiese:
Related to https://github.com/NVIDIA/k8s-operator-libs/pull/50 & https://github.com/NVIDIA/k8s-operator-libs/pull/53


rollandf (changes requested):
Need to update Docker file also:
https://github.com/Mellanox/network-operator/blob/master/Dockerfile#L50

tobiasgiese:
> Need to update Docker file also

Good catch 👍🏻 
Should be fixed in [this diff](https://github.com/Mellanox/network-operator/compare/eef6f606c74050ec433cc431412058d1bb6f710f..a8c64e33860451f14d220594c967e29d2e0f07ad)

rollandf (approved):

tobiasgiese:
I had to remove the test-files used by the unit-tests. As the tests are in the libs repo now we don't need them anymore.

Can you please take a look again @rollandf @vasrem? Thanks!

PR #1186, opened 2024-11-24
chore: add check for NIC Configuration Helm chart

rollandf:


Makefile
```diff
@@ -253,6 +257,12 @@ check-release-build: release-build
 	$(info checking for git diff after running 'make release-build')
 	git diff --quiet ; if [ $$? -eq 1 ] ; then echo "Please, commit templates after running 'make release-build' command"; exit 1 ; fi
 
+.PHONY: check-update-helm-sub-charts
+check-update-helm-sub-charts: update-helm-sub-charts
+	$(info checking for git diff after running 'make release-build')
+	git diff --quiet ; if [ $$? -eq 1 ] ; then echo "Please, commit Helm Chart after running 'make update-helm-sub-charts' command"; exit 1 ; fi
```
heyvister1, line 263: Consider the following change. This will also won't print `Please, commit Helm Chart after running...` upon success which might be confusing.
```suggestion
	@git diff --quiet || (echo "Error: Please commit Helm Chart changes after running 'make update-helm-sub-charts' command" && exit 1)
```

Makefile
```diff
@@ -414,6 +424,17 @@ release-build:
 	cd hack && $(GO) run release.go --templateDir ./templates/values/ --outputDir ../deployment/network-operator/
 	cd hack && $(GO) run release.go --templateDir ./templates/config/manager --outputDir ../config/manager/
 
+.PHONY: helm-update-nic-configuration
+helm-update-nic-configuration: $(HELM) $(HELM_CHART_DEP_CHECKDIR)
+	rm -rf ./build/helmcharts/nic-configuration-operator-chart
+	@NIC_CONF_VERSION=$$(yq '.nicConfigurationOperator.version' hack/release.yaml | sed 's/^v//'); \
```
heyvister1, line 430: What about changes in `nicConfigurationConfigDaemon` ?

Makefile
rollandf: Done

Makefile
rollandf: I took here an assumption that the version of the operator match the chart version.
We don't have currently support for chart versions in release.yaml.

Makefile
rollandf: For now I just updated the chart. Need to think about a way to check if all sub charts are updated.

almaslennikov (approved):

heyvister1 (approved):

PR #1191, opened 2024-11-26
task: Adding maintenance operator charts

heyvister1:
Adding maintenance operator charts

### DIFF
--- a/deployment/network-operator/Chart.yaml
+++ b/deployment/network-operator/Chart.yaml
@@ -24,3 +24,7 @@ dependencies:
   name: nic-configuration-operator-chart
   repository: ''
   version: 0.0.1
+- condition: maintenanceOperator.enabled
+  name: maintenance-operator-chart
+  repository: ''
+  version: 0.0.1
--- a/deployment/network-operator/values.yaml
+++ b/deployment/network-operator/values.yaml
@@ -33,6 +33,10 @@ nicConfigurationOperator:
   # -- Deploy NIC Configuration Operator.
   enabled: false
 
+maintenanceOperator:
+  # -- Deploy Maintenance Operator.
+  enabled: false  
+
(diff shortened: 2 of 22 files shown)


deployment/network-operator/Chart.yaml
```diff
@@ -24,3 +24,7 @@ dependencies:
   name: nic-configuration-operator-chart
   repository: ''
   version: 0.0.1
+- condition: maintenanceOperator.enabled
+  name: maintenance-operator-chart
+  repository: ''
+  version: 0.1.1
```
rollandf, line 30: should be: "0.0.1"

deployment/network-operator/values.yaml
```diff
@@ -181,6 +185,29 @@ nic-configuration-operator-chart:
       name: nic-configuration-operator-daemon
       tag: v0.1.14
 
+# Maintenance Operator chart related values.
+maintenance-operator-chart:
+  operator:
+    image:
+      repository: gghcr.io/mellanox
```
almaslennikov, line 192: typo here, use make release-build to propagate the changes from hack/release.yaml

almaslennikov (changes requested):

deployment/network-operator/values.yaml
heyvister1: Done

rollandf (approved):

almaslennikov (approved):

almaslennikov:
@rollandf do we merge with the documentation CI failing? Looks like some token is not set up there

heyvister1:
> @rollandf do we merge with the documentation CI failing? Looks like some token is not set up there

Its pending a fix which I have introduced
https://github.com/Mellanox/network-operator-docs/pull/134
PTAL

PR #1210, opened 2024-12-08
bug: add bundle.Dockerfile to CI PR

rollandf:
When creating the CI PR with updates to OCP bundle, the `bundle.Dockerfile` should be also commited.


maze88:
Lgtm, could instead make line above `git add bundle*`, but I think explicit (as is now) is clearer.

maze88 (approved):

heyvister1 (approved):

PR #1213, opened 2024-12-11
chore: Use staging image repository for development

e0ne:


heyvister1:
@e0ne did you use make release or did the changes manually? check if we need to fix also links for OCP bundle and CR example links

maze88 (approved):

e0ne:
> / check release-build

I used `make release-build` to generate values file. `check release-build` check succeed too

heyvister1 (approved):

hack/release.yaml
```diff
@@ -1,6 +1,6 @@
 NetworkOperator:
   image: network-operator
-  repository: nvcr.io/nvidia/cloud-native
+  repository: nvcr.io/nvstaging/mellanox/
```
heyvister1, line 3: Hey @e0ne, looks like this change breaks helm installation, since image name is getting double` //` after repositery name
```
k -n nvidia-network-operator get pods network-operator-keep-ncp-tcg8h -oyaml | grep image:
    image: nvcr.io/nvstaging/mellanox//network-operator:v24.10.0
```

e0ne: fixed

PR #1220, opened 2024-12-22
chore: fix CVE for net

rollandf:
CVE-2024-45338
https://github.com/advisories/GHSA-w32m-9786-jp63


rollandf:
Seems we get it from:
https://github.com/NVIDIA/k8s-operator-libs

adrianchiris (approved):

heyvister1 (approved):

adrianchiris:
@rollandf we should probably bump opereator lib at some point

PR #1240, opened 2025-01-06
Feature/update sriov tags to 73f66a5

e0ne:


deployment/network-operator/charts/sriov-network-operator/Chart.yaml
```diff
@@ -1,8 +1,8 @@
 apiVersion: v2
 name: sriov-network-operator
-version: 0.1.0
-kubeVersion: '>= 1.16.0-0'
-appVersion: 1.2.0
+version: 0.0.0-dev
```
rollandf, line 3: Do we want this change?

e0ne: let's revert it for now and discuss how do we want to implement chart sync

rollandf (dismissed):

rollandf (approved):

PR #1248, opened 2025-01-08
bug: support container image SHA256 version

rollandf:
In OCP disconnected environment, the mirrored images are not using tags but SHA256.

This change support boths format in the NCP except for DOCA-Driver.

### DIFF
--- a/pkg/render/render.go
+++ b/pkg/render/render.go
@@ -103,6 +103,14 @@ func nindentPrefix(spaces int, prefix, v string) string {
 	return strings.Replace(nindent(spaces, prefix+v), " ", "", len(prefix))
 }
 
+// imagePath returns container image path, supporting sha256 format
+func imagePath(repository, image, version string) string {
+	if strings.HasPrefix(version, "sha256:") {
+		return repository + "/" + image + "@" + version
+	}
+	return repository + "/" + image + ":" + version
+}
+
 // renderFile renders a single file to a list of k8s unstructured objects
 func (r *textTemplateRenderer) renderFile(filePath string, data *TemplatingData) ([]*unstructured.Unstructured, error) {
 	// Read file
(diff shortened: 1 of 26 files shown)


pkg/render/render.go
```diff
@@ -103,6 +103,14 @@ func nindentPrefix(spaces int, prefix, v string) string {
 	return strings.Replace(nindent(spaces, prefix+v), " ", "", len(prefix))
 }
 
+// imagePath returns contaimer image path, supporting sha256 format
+func imagePath(repository, image, version string) string {
```
e0ne, line 107: Why do we need it as a template function instead of generating it in the code and pass the value into template? IMO, it makes templates more complicated whithout additional benefits

rollandf: My thoughts that this way will require less changes.
In code, we will need for each state:
- add a new field 'imagePath' to the `ManifestRenderData`
- generate the value and add it in the `ManifestRenderData`
- update the manifests to use the new 'imagePath'

I am OK to do the change if you prefer that way

e0ne: Sounds good to me to keep it as proposed now

e0ne (dismissed):

almaslennikov (dismissed):

pkg/render/render.go
```diff
@@ -103,6 +103,14 @@ func nindentPrefix(spaces int, prefix, v string) string {
 	return strings.Replace(nindent(spaces, prefix+v), " ", "", len(prefix))
 }
 
+// imagePath returns contaimer image path, supporting sha256 format
```
adrianchiris, line 106: nit: contaimer -> container :)

adrianchiris (dismissed):

pkg/render/render.go
rollandf: done

adrianchiris (approved):

PR #1253, opened 2025-01-14
chore: fix CVEs 24.10

rollandf:


rollandf:
@e0ne The failure is on `test-helm-sriov-dp-nic-selector-device-id`
I think it is related to the change done in the SRIOV DP naming, where we changed also the test.

PR #1258, opened 2025-01-19
fix: refactor docs-ci downstream trigger to operate via workflow_dispatch only

maze88:
this change is coupled with the change in network-operator: https://github.com/Mellanox/network-operator-docs/pull/166

The workflow also prints the URL to the downstream workflow, and its result is propagated (upstream will fail if downstream fails).

### DIFF
--- a/.github/workflows/docs-ci-trigger.yaml
+++ b/.github/workflows/docs-ci-trigger.yaml
@@ -6,8 +6,17 @@ on:
     - "v*"
 
 jobs:
-  trigger_downstream_workflow:
-    uses: Mellanox/network-operator-docs/.github/workflows/docs-ci.yaml@main
-    with:
-      git_tag: ${{ github.ref_name }}
-    secrets: inherit
+  call_docs_ci_downstream_workflow:
+    runs-on: ubuntu-latest
+    env:
+      GH_TOKEN: ${{ secrets.GH_TOKEN_NVIDIA_CI_CD }}
+      GIT_TAG: ${{ github.ref_name }}
+    steps:
+    - run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/docs-ci-trigger.yaml
```diff
@@ -6,8 +6,17 @@ on:
...
-  trigger_downstream_workflow:
-    uses: Mellanox/network-operator-docs/.github/workflows/docs-ci.yaml@main
-    with:
-      git_tag: ${{ github.ref_name }}
-    secrets: inherit
+  call_docs_ci_downstream_workflow:
+    runs-on: ubuntu-latest
+    env:
+      GH_TOKEN: ${{ secrets.GH_TOKEN_NVIDIA_CI_CD }}
+      GIT_TAG: ${{ github.ref_name }}
```
rollandf, line 13: Note that in doc CI we also need:
```
 env:
        NGC_CLI_API_KEY: ${{ secrets.NVCR_TOKEN }}
```
https://github.com/Mellanox/network-operator-docs/blob/main/.github/workflows/docs-ci.yaml#L35

almaslennikov (approved):

.github/workflows/docs-ci-trigger.yaml
maze88: i don't think that'll be needed in this calling workflow.

it will only be used in the actual downstream workflow, where it's already been added: https://github.com/Mellanox/network-operator-docs/blob/main/.github/workflows/docs-ci.yaml#L35

rollandf (approved):

PR #1314, opened 2025-02-03
fix: if chart version already exists then *won't* fail on chart-push

rollandf:
Caveat:
If the specific error (`Chart '<chart>' already exists in the repository.`) occurs then the CI workflow (`ci.yaml`) *won't* fail. However it's important to note that the `make` command will behave normally around the failure (i.e. it will exit) - this is fine in the current configuration (Makefile & ci.yaml), however - it could cause issues in the following cases:
- Expantion of the make target with additional lines (will fail and exit on the current/single line).
- Addition of make goal(s) *after* the `chart-push` target (will fail and exit and not run any following targets).


rollandf:
backport of #1310

e0ne (approved):

heyvister1 (approved):

PR #1318, opened 2025-02-03
bug: fix admission_controller template while not using cert-manager

heyvister1:
In `ValidatingWebhookConfiguration` template `caBundle` was using server side tls cert instead of CA cert

### DIFF
--- a/deployment/network-operator/values.yaml
+++ b/deployment/network-operator/values.yaml
@@ -275,6 +275,11 @@ operator:
     # -- Use cert-manager for generating self-signed certificate.
     useCertManager: true
     # certificate:
+      # caCrt: |
+      #   -----BEGIN CERTIFICATE-----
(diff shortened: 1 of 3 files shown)


almaslennikov (dismissed):

deployment/network-operator/values.yaml
```diff
@@ -275,6 +275,11 @@ operator:
     # -- Use cert-manager for generating self-signed certificate.
     useCertManager: true
     # certificate:
+      # caCrt: |
```
rollandf, line 278: Update the template [here](https://github.com/Mellanox/network-operator/blob/master/hack/templates/values/values.template)
and run `make release-build`

heyvister1: Done

rollandf (approved):

rollandf:
@heyvister1 Do we want this one in 25.1 branch?

heyvister1:
> @heyvister1 Do we want this one in 25.1 branch?

Yes. Created the following PR #1327

PR #1332, opened 2025-02-12
bug: replace MOFED to DOCA on API docs

rollandf:
- The fix is only for the API docs that are reflected in official documentation
- API fields are not changed to not break API


heyvister1 (approved):
Assuming CRD manifest changes were generated and not changed inline

rollandf:
> Assuming CRD manifest changes were generated and not changed inline

Yes. There is also a check checking the manifests are updated: https://github.com/Mellanox/network-operator/blob/master/.github/workflows/checks.yaml#L52

e0ne (approved):

PR #1345, opened 2025-02-23
ci: update version from release branch

rollandf:
The PR updating the Network Operator Chart version and release.yaml for release flow should be based on the release branch if the version is not a beta.(RC+GA)

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -39,7 +39,8 @@ jobs:
       - run: |
           git config user.name  nvidia-ci-cd
           git config user.email svc-cloud-orch-gh@nvidia.com
-          git checkout -b cicd/update-network-operator-to-$RELEASE_VERSION
+          git fetch origin $BASE_BRANCH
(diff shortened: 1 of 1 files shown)


maze88 (changes requested):
Perhaps a cleaner implementation would be to [directly _checkout_ the appropriate branch](https://github.com/actions/checkout?tab=readme-ov-file#checkout-a-different-branch) of network-operator with:
```yaml
  with:
     ref: ${{ env.BASE_BRANCH }}
```
This requires moving the _Determine base branch_ step (and its two predecessors) to be before the checkout step.

rollandf:
> Perhaps a cleaner implementation would be to [directly _checkout_ the appropriate branch](https://github.com/actions/checkout?tab=readme-ov-file#checkout-a-different-branch) of network-operator with:
> 
> ```yaml
>   with:
>      ref: ${{ env.BASE_BRANCH }}
> ```
> 
> This requires moving the _Determine base branch_ step (and its two predecessors) to be before the checkout step.

See discussion in #1346

maze88 (approved):

PR #1346, opened 2025-02-24
ci: update version from release branch

rollandf:
The PR updating the Network Operator Chart version and release.yaml for release flow should be based on the release branch if the version is not a beta.(RC+GA)

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -39,7 +39,8 @@ jobs:
       - run: |
           git config user.name  nvidia-ci-cd
           git config user.email svc-cloud-orch-gh@nvidia.com
-          git checkout -b cicd/update-network-operator-to-$RELEASE_VERSION
+          git fetch origin $BASE_BRANCH
+          git checkout -b cicd/update-network-operator-to-$RELEASE_VERSION origin/$BASE_BRANCH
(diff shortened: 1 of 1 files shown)


maze88:
Isn't this implemented already in the [_Determine base branch_](https://github.com/Mellanox/network-operator/blob/master/.github/workflows/release.yaml#L19) step?

rollandf:
> Isn't this implemented already within the [_Determine base branch_](https://github.com/Mellanox/network-operator/blob/master/.github/workflows/release.yaml#L19) step?

The step only find out the branch to push to, but we need also to create the PR on top of it.

maze88:
> > Perhaps a cleaner implementation would be to [directly _checkout_ the appropriate branch](https://github.com/actions/checkout?tab=readme-ov-file#checkout-a-different-branch) of network-operator with:
> > ```yaml
> >   with:
> >      ref: ${{ env.BASE_BRANCH }}
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This requires moving the _Determine base branch_ step (and its two predecessors) to be before the checkout step.
> 
> This will fail if the release branch was not created yet.

I see, yes - so in this workflow we'll do it the original PR's way.

maze88 (approved):

PR #1356, opened 2025-02-26
chore: Use public Network Operator image

e0ne:


bundle/metadata/annotations.yaml
```diff
@@ -4,8 +4,8 @@ annotations:
   operators.operatorframework.io.bundle.manifests.v1: manifests/
   operators.operatorframework.io.bundle.metadata.v1: metadata/
   operators.operatorframework.io.bundle.package.v1: nvidia-network-operator
-  operators.operatorframework.io.bundle.channels.v1: stable,v25.1
-  operators.operatorframework.io.bundle.channel.default.v1: v25.1
+  operators.operatorframework.io.bundle.channels.v1: v25.1.0,stable
```
rollandf, line 7: channels should be without minor

rollandf: Did you run make bundle?
I don't see the doca-driver udpated

rollandf: When do we update Chart.yaml ?

e0ne: fixed

e0ne: Chart.yaml is already updated

rollandf (approved):

PR #1365, opened 2025-03-07
Cleanup lables

alekseysen:


almaslennikov (changes requested):
Hey @alekseysen, thanks for the PR!

Can you please add signed-off in commit:
Signed-off-by: Author Name <authoremail@example.com>

Also, the files in the deployment folder are generated from the templates in `hack/` with `make release-build` command. Can you update the relevant templates so that the CI passes?

PR #1373, opened 2025-03-11
feat: add NIC Configuration Operator as a NCP state

almaslennikov:


rollandf (approved):
LGTM 
nit about having loglevel in API

api/v1alpha1/nicclusterpolicy_types.go
```diff
@@ -288,6 +288,43 @@ type DOCATelemetryServiceSpec struct {
...
 }
 
+// NicFirmwareStorageSpec contains configuration for the NIC firmware storage
+type NicFirmwareStorageSpec struct {
+	// Create specifies whether to create a new PVC or use an existing one
+	// +kubebuilder:default:=true
+	Create bool `json:"create,omitempty"`
+	// PVCName is the name of the PVC to mount as NIC Firmware storage. Default value: "nic-fw-storage-pvc"
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=253
```
e0ne, line 298: Can we create PVC name longer than 63 symbols? This [1] mention 63 symbols limit

[1] https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-label-names

api/v1alpha1/nicclusterpolicy_types.go
```diff
@@ -288,6 +288,43 @@ type DOCATelemetryServiceSpec struct {
...
+	// +kubebuilder:validation:Pattern=`^[a-z0-9]([-a-z0-9]*[a-z0-9])?$`
+	// +kubebuilder:default:="nic-fw-storage-pvc"
+	PVCName string `json:"pvcName,omitempty"`
+	// StorageClassName is the name of a storage class to be used to store NIC FW binaries during NIC FW upgrade.
+	// If not provided, the cluster-default storage class will be used
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=253
```
e0ne, line 305: The same as above

config/rbac/role.yaml
```diff
@@ -117,6 +117,41 @@ rules:
   - get
   - list
   - watch
+- apiGroups:
```
e0ne, line 120: thanks for adding OCP support from the beginning

manifests/state-nic-configuration-operator/005-configuration.net.nvidia.com_nicconfigurationtemplates.yaml
```diff
@@ -0,0 +1,189 @@
+# 2025 NVIDIA CORPORATION & AFFILIATES
```
e0ne, line 1: Please, change ordering numbers in state names

deployment/network-operator/Chart.yaml
```diff
@@ -20,10 +20,6 @@ dependencies:
     name: sriov-network-operator
     repository: ''
     version: 0.1.0
-  - condition: nicConfigurationOperator.enabled
```
e0ne, line 23: It's good to delete. this in general but we meting this is mentioned in our documentation so it seems we need to support it at lease for one release as a deprecated way to deploy NIC Configuration Operator

[1] https://docs.nvidia.com/networking/display/kubernetes2501/customizations/helm.html

e0ne (changes requested):

api/v1alpha1/nicclusterpolicy_types.go
almaslennikov: Changed maxLength to 63

api/v1alpha1/nicclusterpolicy_types.go
almaslennikov: Changed maxLength to 63

manifests/state-nic-configuration-operator/005-configuration.net.nvidia.com_nicconfigurationtemplates.yaml
almaslennikov: Changed the ordering number so that the number do not repeat

deployment/network-operator/Chart.yaml
almaslennikov: Dropped the commit with deletion of the nic operator helm chart. Let's add the deprecation message in a separate commit with doc updates

api/v1alpha1/nicclusterpolicy_types.go
```diff
@@ -288,6 +288,43 @@ type DOCATelemetryServiceSpec struct {
...
+	// Image information for nic-configuration-operator
+	Operator *ImageSpec `json:"operator"`
+	// Image information for nic-configuration-daemon
+	ConfigurationDaemon *ImageSpec `json:"configurationDaemon"`
+	// NicFirmwareStorage contains configuration for the NIC firmware storage
+	NicFirmwareStorage *NicFirmwareStorageSpec `json:"nicFirmwareStorage,omitempty"`
+	// LogLevel sets the verbosity level of the logs. info|debug
+	// +kubebuilder:validation:Enum={"info", "debug"}
+	// +kubebuilder:default:="info"
+	LogLevel string `json:"logLevel,omitempty"`
```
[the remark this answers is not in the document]
almaslennikov: So the operator is a big project and produces a ton of debug logs which the users do not need by default. However, in case of troubleshooting we really need them. I would prefer to have the LogLevel setting but can come up with some different solution if we won't agree on it. 

@e0ne what's your opinion?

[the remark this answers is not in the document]
rollandf: I am OK with having it.

rollandf (approved):

e0ne (approved):

rollandf (approved):

PR #1376, opened 2025-03-13
bug: Cleanup labels

alekseysen:
chore: add cleanupLabels

Signed-off-by: Aleksey Senin <alekseys@nvidia.com>

### DIFF
--- a/deployment/network-operator/values.yaml
+++ b/deployment/network-operator/values.yaml
@@ -44,6 +44,8 @@ node-feature-discovery:
   enableNodeFeatureApi: true
   featureGates:
     NodeFeatureAPI: true
+  # -- Enable labels cleanup when uninstalling NFD
+  cleanupLabels: false
 
   # -- NFD master deployment configuration.
   # @notationType -- yaml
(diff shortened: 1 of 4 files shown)


almaslennikov (approved):
Thank you for addressing the failing CI jobs

deployment/network-operator/values.yaml
```diff
@@ -44,6 +44,7 @@ node-feature-discovery:
   enableNodeFeatureApi: true
   featureGates:
     NodeFeatureAPI: true
+  cleanupLabels: false
```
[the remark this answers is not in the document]
alekseysen: Done.
Thanks for the title.

rollandf (approved):

deployment/network-operator/values.yaml
```diff
@@ -44,6 +44,8 @@ node-feature-discovery:
   enableNodeFeatureApi: true
   featureGates:
     NodeFeatureAPI: true
+  # -- Enable labels cleanup when uninstalling NFD
+  cleanupLabels: false
```
rollandf, line 48: So this default is changing the current behavior?

alekseysen: Correct.

When uninstalling network-operator v24.10, the components defined in NicClusterPolicy continue running due to the presence of labels on the nodes. However, a newer version of NFD introduced a cleanup mechanism that removes these labels, resulting in the removal of network-operator components, including the MOFED pod, for example.

e0ne:
We need to create a follow up PR once https://github.com/kubernetes-sigs/node-feature-discovery/pull/2076 will be merged

PR #1385, opened 2025-03-20
chore: deprecate NIC Configuration Operator helm chart + update README

almaslennikov:


rollandf (commented):
LGTM.
Can you please also add the spec in the following:
https://github.com/Mellanox/network-operator/blob/master/hack/templates/crs/mellanox.com_v1alpha1_nicclusterpolicy_cr-full-ocp.template#L105

QA team uses this to check what to test in OCP

rollandf (dismissed):

rollandf (approved):

heyvister1 (approved):

PR #1387, opened 2025-03-24
fix: allow pr_checker job to override branch protection rules

almaslennikov:
Current branch protection rules require at least one approving review. If we want the cicd PRs to be merged automatically, we need to override these rules with `--admin` option


e0ne (approved):
LGTM

rollandf (approved):

heyvister1 (approved):

PR #1427, opened 2025-04-09
fix: cleanup stale PVCs and make their config immutable in NCP

almaslennikov:
PVC spec is immutable and since the network operator doesn't have the functionality to update the resource by deletion, let's make the PVC-related fields in the NCP immutable.
To change their values, the user needs to delete the nicFirmwareStorage section of the NCP and recreate it again with new values.


rollandf (approved):

e0ne (approved):
LGTM

PR #1448, opened 2025-04-23
fix: update ocp version in Makefile

rollandf:


e0ne (approved):
Thanks for fixing this! I missed it

PR #1452, opened 2025-04-24
chore: Bump Spectrum-X Operator to v0.0.3

e0ne:


rollandf (commented):
Do we need also to have it here? (and run `make manifests`)

https://github.com/Mellanox/network-operator/blob/master/controllers/nicclusterpolicy_controller.go#L104

I think that in order to create a role, you need to have it

e0ne:
> Do we need also to have it here? (and run `make manifests`)
`make manifests` is executed, nothing changed.

> https://github.com/Mellanox/network-operator/blob/master/controllers/nicclusterpolicy_controller.go#L104
> 
> I think that in order to create a role, you need to have it
could you please elaborate more on this?

e0ne:
> Since the NO is creating a role for the SpectrumX operator for SRIOV objects, I think it also should have the same objects defined in its role.

done, thanks for pointing me on that!

rollandf (approved):

PR #1450, opened 2025-04-24
chore: update doca-driver to 25.04-0.5.1.0-2

rollandf:

PR #1453, opened 2025-04-25
chore: bumping doca-driver version to be '25.04-0.4.2.0-1'

heyvister1:


example/crs/mellanox.com_v1alpha1_nicclusterpolicy_cr-full-ocp.yaml
```diff
@@ -24,7 +24,7 @@ spec:
   ofedDriver:
     image: doca-driver
     repository: nvcr.io/nvstaging/mellanox
-    version: 25.04-0.3.8.0-0
+    version: 25.04-0.4.2.0-0
```
rollandf, line 27: If you want to test the fix, the version should be '25.04-0.4.2.0-1'

heyvister1:
Same issue as in #1450 
```
E: Unable to correct problems, you have held broken packages.
Reading package lists...
Building dependency tree...
Reading state information...
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

The following packages have unmet dependencies:
 mlnx-ofed-kernel-modules : Depends: linux-modules-extra-5.15.0-106-generic but it is not installable
```

e0ne:
DOCA driver install fails: `[25-Apr-25_10:01:46] Command "apt-get install -y /mnt/drivers-inventory/5.15.0-130-generic/25.04-0.4.2.0/*.deb" failed with exit code: 100
`

PR #1502, opened 2025-05-12
fix: bump k8s-operator-lib version to fix nodemaintenance deletion

heyvister1:


rollandf (approved):

rollandf:
Please also have the same in master

PR #1527, opened 2025-05-20
chore: tag all managed repos on release

almaslennikov:
extend the existing release.yaml workflow to tag all managed repos that have the `sourceRepository` field in the release.yaml

Question to reviewers: do we want to extend the hack/release.yaml to contain some additional service info?

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -12,8 +12,20 @@ jobs:
       ISSUE_TITLE: ${{ github.event.issue.title }}
     steps:
       - uses: actions/checkout@v4
-      - run: |
-          echo "APP_VERSION=$(echo $ISSUE_TITLE | awk '{ print $2 }')" | tee -a $GITHUB_ENV
+      - id: set-version
+        run: |
+          # full release version, i.e. v25.4.0-rc.1
+          APP_VERSION=$(echo $ISSUE_TITLE | awk '{ print $2 }')
+          echo "app_version=${APP_VERSION}" >> $GITHUB_OUTPUT
+          echo "APP_VERSION=${APP_VERSION}" | tee -a $GITHUB_ENV
+
+          # major.minor version with v prefix and x suffix, i.e. v25.4.x
+          V_MAJOR_MINOR_X=$(echo $APP_VERSION | grep -Eo '^v[0-9]+\.[0-9]+').x
+          echo "V_MAJOR_MINOR_X=${V_MAJOR_MINOR_X}" | tee -a $GITHUB_ENV
+
+          # major.minor version without v prefix and with x suffix, i.e. 25.4.x
+          MAJOR_MINOR_X=$(echo ${V_MAJOR_MINOR_X#v})
+          echo "major_minor_x=${MAJOR_MINOR_X}" >> $GITHUB_OUTPUT
       - run: |
           echo "CHART_VERSION=$(echo ${APP_VERSION#v})" | tee -a $GITHUB_ENV
       - name: Determine base branch
@@ -21,8 +33,7 @@ jobs:
           if echo $APP_VERSION | grep -q beta; then
             base_branch=master
           else
-            v_major_minor=$(echo $APP_VERSION | grep -Eo '^v[0-9]+\.[0-9]+')
-            base_branch=$v_major_minor.x
+            base_branch=$V_MAJOR_MINOR_X
           fi
           echo BASE_BRANCH=$base_branch | tee -a $GITHUB_ENV
       - name: Verify release branch exists if "rc" version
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yaml
```diff
@@ -7,22 +7,31 @@ jobs:
   update-network-operator-version:
     if: startsWith(github.event.issue.title, 'Release v')
     runs-on: ubuntu-24.04
+    outputs:
```
maze88, line 10: nit: more intuitive to put `outputs` at bottom of job (after `steps`).

.github/workflows/release.yaml
```diff
@@ -7,22 +7,31 @@ jobs:
...
       ISSUE_TITLE: ${{ github.event.issue.title }}
     steps:
       - uses: actions/checkout@v4
-      - run: |
-          echo "APP_VERSION=$(echo $ISSUE_TITLE | awk '{ print $2 }')" | tee -a $GITHUB_ENV
+      - id: set-version
+        run: |
+          APP_VERSION=$(echo $ISSUE_TITLE | awk '{ print $2 }')
+          echo "app_version=${APP_VERSION}" >> $GITHUB_OUTPUT
+          echo "APP_VERSION=${APP_VERSION}" >> $GITHUB_ENV
```
maze88, line 22: recommend using `tee` when assigning to `GITHUB_ENV`, then it also gets printed to log output.
see example in line below with `CHART_VERSION`! (:

.github/workflows/release.yaml
```diff
@@ -7,22 +7,31 @@ jobs:
...
       - run: |
           echo "CHART_VERSION=$(echo ${APP_VERSION#v})" | tee -a $GITHUB_ENV
-      - name: Determine base branch
+      - name: Determine branches
+        id: determine-branches
         run: |
+          major_minor_x=$(echo $APP_VERSION | grep -Eo '^[0-9]+\.[0-9]+').x
```
maze88, line 28: if `APP_VERSION` still has "v" prefix then:
- piped `grep` regex will **fail** to isolate `<major.minor>`, since it searches for digits (`[0-9]+`) at the *beginning* (`^`) of the line...!
- since we use versions with and without "v" prefix, then it's important to emphasize it in variable name - change to `v_major_minor_x`.
- in complex variable assignment lines like this, it helps a lot if we add a comment with an example value at EOL - such as `# example: v25.04.x` (or without "v"?).

.github/workflows/release.yaml
```diff
@@ -59,31 +68,62 @@ jobs:
...
     runs-on: ubuntu-24.04
+    needs: [get-managed-components, update-network-operator-version]
+    strategy:
+      matrix:
+        component: ${{ fromJson(needs.get-managed-components.outputs.managed_components) }}
     env:
       GH_TOKEN: ${{ secrets.GH_TOKEN_NVIDIA_CI_CD }}
-      ISSUE_TITLE: ${{ github.event.issue.title }}
+      APP_VERSION: ${{ needs.update-network-operator-version.outputs.app_version }}
+      MAJOR_MINOR_X: ${{ needs.update-network-operator-version.outputs.major_minor_x }}
```
maze88, line 95: see earlier comment about "v" prefix regarding this env var too.

.github/workflows/release.yaml
```diff
@@ -59,31 +68,62 @@ jobs:
...
+          components=$(yq -o=json 'to_entries | map(select(.value.sourceRepository != null) | .value.sourceRepository)' hack/release.yaml | jq -c .)
+          echo $components
+          echo "managed_components=$(echo $components)" >> $GITHUB_OUTPUT
+
+  update-component-versions:
     runs-on: ubuntu-24.04
+    needs: [get-managed-components, update-network-operator-version]
+    strategy:
+      matrix:
+        component: ${{ fromJson(needs.get-managed-components.outputs.managed_components) }}
```
maze88, line 91: cool! ;]

maze88 (changes requested):

.github/workflows/release.yaml
```diff
@@ -58,32 +69,66 @@ jobs:
...
+
+            echo "Creating the version branch from the new release branch"
+            git checkout -b $APP_VERSION
+          fi
 
-          git checkout -b $APP_VERSION origin/$BASE_BRANCH
+          echo "Creating and pushing the tag"
           git tag network-operator-$APP_VERSION
-          git push origin --tags
+          git push origin --tags || echo "Failed to push tags"
```
maze88, line 134: not sure this is ideal... because `|| echo ...` will result in a successful exit status, and that seems like a false positive...

instead, just rely on the original `git push` command, which if fails should provide a decent error message **and exit status**!

maze88 (changes requested):

.github/workflows/release.yaml
almaslennikov: Good observation, thanks!

e0ne (approved):

PR #1544, opened 2025-05-29
fix: Implement Taint and Toleration Handling for NicClusterPolicy

dahalperin:
Introduces taint-aware node filtering and OFED deployment.

- Controller now watches Node taint changes.
- New `NodeTaintChangedPredicate` and composite node filter (labels & taints) for OFED node selection.
- `state_ofed` uses this filter and adds a default `nvidia.com/gpu` toleration to OFED DaemonSets.
- NCP tolerations are respected.
- Extensive new tests for controller, filtering, and OFED state.

### DIFF
--- a/controllers/predicate.go
+++ b/controllers/predicate.go
@@ -20,6 +20,8 @@ import (
 	"reflect"
 
 	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -98,4 +100,27 @@ func (p IgnoreSameContentPredicate) handleDeployment(oldObj, newObj *appsv1.Depl
 	revAnnotation := "deployment.kubernetes.io/revision"
 	delete(oldObj.Annotations, revAnnotation)
 	delete(newObj.Annotations, revAnnotation)
+}
+
+// NodeTaintChangedPredicate filters if node taints have changed
+type NodeTaintChangedPredicate struct {
+	predicate.Funcs
+}
+
+// Update returns true if the node taints have been changed
+func (p NodeTaintChangedPredicate) Update(e event.UpdateEvent) bool {
+	if e.ObjectOld == nil || e.ObjectNew == nil {
+		return false
+	}
+
+	oldNode, ok := e.ObjectOld.(*corev1.Node)
+	if !ok {
+		return false
+	}
+	newNode, ok := e.ObjectNew.(*corev1.Node)
+	if !ok {
+		return false
+	}
+
+	return !equality.Semantic.DeepEqual(oldNode.Spec.Taints, newNode.Spec.Taints)
 }
--- a/pkg/nodeinfo/filter.go
+++ b/pkg/nodeinfo/filter.go
@@ -141,3 +141,85 @@ func (b *NodeLabelNoValFilterBuilder) Reset() *NodeLabelNoValFilterBuilder {
 	b.filter = newNodeLabelNoValFilter()
 	return b
 }
+
+// A node taint filter. use NewNodeTaintFilterBuilder to create instances
+type nodeTaintFilter struct {
+	tolerations []corev1.Toleration
+}
+
+// Apply Filter on Nodes
+func (ntf *nodeTaintFilter) Apply(nodes []*corev1.Node) (filtered []*corev1.Node) {
+	for _, node := range nodes {
+		if ntf.toleratesTaints(node.Spec.Taints) {
+			filtered = append(filtered, node)
+		}
+	}
+	return filtered
+}
+
+// toleratesTaints checks if the network-operator pod can tolerate all taints on the node
+func (ntf *nodeTaintFilter) toleratesTaints(taints []corev1.Taint) bool {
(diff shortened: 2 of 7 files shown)


pkg/state/state_ofed.go
```diff
@@ -466,10 +494,19 @@ func (s *stateOFED) GetManifestObjects(
...
+	// Add default GPU toleration to the list of tolerations from CR
+	processedTolerations := addGPUToleration(cr.Spec.Tolerations)
+
+	// Create a composite filter for labels and taints
+	compositeFilter := nodeinfo.NewNodeFilterBuilder().
+		WithLabel(nodeinfo.NodeLabelMlnxNIC, "true").
+		WithTolerations(processedTolerations). // Used for filtering
+		Build()
+
+	nodePools := nodeInfo.GetNodePools(compositeFilter) // Apply the composite filter
```
rollandf: `GetNodePools` can get multiple filters. See [here](https://github.com/Mellanox/network-operator/blob/master/pkg/nodeinfo/node_info.go#L68)
So maybe we can skip the composite, and just pass two filters.

pkg/nodeinfo/filter.go
```diff
@@ -141,3 +141,136 @@ func (b *NodeLabelNoValFilterBuilder) Reset() *NodeLabelNoValFilterBuilder {
...
+}
+
+// Reset NodeTaintFilterBuilder
+func (b *NodeTaintFilterBuilder) Reset() *NodeTaintFilterBuilder {
+	b.filter = newNodeTaintFilter()
+	return b
+}
+
+// CompositeNodeFilter combines multiple filters into one
+type CompositeNodeFilter struct {
```
rollandf: I don't think we need this.

pkg/state/state_ofed.go
```diff
@@ -483,8 +520,9 @@ func (s *stateOFED) GetManifestObjects(
 
 	for _, np := range nodePools {
 		nodePool := np
-		// render objects
-		renderedObjs, err := renderObjects(ctx, &nodePool, useDtk, s, cr, reqLogger, clusterInfo, docaProvider)
+		// render objects, passing processedTolerations
+		renderedObjs, err := renderObjects(ctx, &nodePool, useDtk, s, cr,
+			processedTolerations, reqLogger, clusterInfo, docaProvider)
```
rollandf: Why do we use the `processedTolerations`?
GPU Toleration is set in the manifests template already.

pkg/nodeinfo/filter.go
```diff
@@ -141,3 +141,91 @@ func (b *NodeLabelNoValFilterBuilder) Reset() *NodeLabelNoValFilterBuilder {
...
+func (ntf *nodeTaintFilter) Apply(nodes []*corev1.Node) (filtered []*corev1.Node) {
+	for _, node := range nodes {
+		if ntf.toleratesTaints(node.Spec.Taints) {
+			filtered = append(filtered, node)
+		}
+	}
+	return filtered
+}
+
+// toleratesTaints checks if the tolerations can tolerate all taints on the node
```
heyvister1:
```suggestion
// toleratesTaints checks if network-operator pod can tolerate all taints on the node
```

controllers/predicate.go
```diff
@@ -99,3 +100,54 @@ func (p IgnoreSameContentPredicate) handleDeployment(oldObj, newObj *appsv1.Depl
...
+		oldTaints[taint.Key] = taint
+	}
+	for _, taint := range newNode.Spec.Taints {
+		newTaints[taint.Key] = taint
+	}
+
+	// Compare taint maps
+	if len(oldTaints) != len(newTaints) {
+		return true
+	}
```
almaslennikov: You can probably use `reflect.DeepEquals` or `equality.Semantic.DeepEquals` from `k8s.io/apimachinery/pkg/api/equality `instead of the multi-stage comparison

pkg/nodeinfo/filter.go
```diff
@@ -141,3 +141,91 @@ func (b *NodeLabelNoValFilterBuilder) Reset() *NodeLabelNoValFilterBuilder {
...
+	// it would normally be filtered. However, we need to special-case the cordon taint.
+	// The loop below will correctly handle this: if only a cordon taint exists, it will be skipped.
+	// If other taints exist, ntf.toleratesTaint() will return false if ntf.tolerations is empty.
+
+	for _, taint := range taints {
+		// If the taint is the specific "unschedulable" (cordon) taint,
+		// we effectively ignore it for the purpose of this filter.
+		// This means a node that is only cordoned (and has no other taints
+		// that aren't tolerated by ntf.tolerations) will not be filtered out by this logic.
+		if taint.Key == "node.kubernetes.io/unschedulable" && taint.Effect == corev1.TaintEffectNoSchedule {
```
almaslennikov: Use `TaintNodeUnschedulable` constant from `k8s.io/api/core/v1`

pkg/state/state_ofed.go
```diff
@@ -453,6 +455,31 @@ func (s *stateOFED) handleAdditionalMounts(
 	return nil
 }
 
+// addGPUToleration adds the nvidia.com/gpu:NoSchedule toleration if not already present
+func addGPUToleration(tolerations []v1.Toleration) []v1.Toleration {
```
almaslennikov, line 465: I see the OFED daemonset manifests explicitely has this toleration, why do we need to add it in the code as well?

dahalperin: We need to use the default toleration before the manifest is generated

rollandf (approved):

PR #1547, opened 2025-06-05
chore: update spectrum-x manifests

rollandf:


almaslennikov (approved):

api/v1alpha1/nicclusterpolicy_types.go
```diff
@@ -332,13 +332,10 @@ type NicConfigurationOperatorSpec struct {
...
 	// Image information for NVIDIA Spectrum-X Operator
 	ImageSpec `json:""`
-	// Spectrum-X Operator ConfigMap name
-	SpectrumXConfig *ConfigMapNameReference `json:"spectrumXConfig,omitempty"`
-	// SR-IOV Network Operator and related CRDs namespace
-	SriovObjNamespace string `json:"sriovObjNamespace,omitempty"`
 }
 
 // NicClusterPolicySpec defines the desired state of NicClusterPolicy
+// +kubebuilder:pruning:PreserveUnknownFields
```
heyvister1: Is this related to SpectrumX changes or its more of a general change in NCP?

pkg/state/state_spectrumx_test.go
```diff
@@ -22,30 +22,36 @@ import (
...
+		catalog = getTestCatalog()
+		catalog.Add(state.InfoTypeStaticConfig,
+			staticconfig.NewProvider(staticconfig.StaticConfig{CniBinDirectory: "custom-cni-bin-directory"}))
 	})
 
 	Context("should render", func() {
 		It("Kubernetes manifests", func() {
 			cr := getSpectrumXOperator()
-			objs, err := ts.renderer.GetManifestObjects(ts.context, cr, ts.catalog, testLogger)
+			objs, err := ts.renderer.GetManifestObjects(ts.context, cr, catalog, testLogger)
```
heyvister1, line 45: What is the difference between ts.catalog and catalog? if you go with catalog, I still see below usage of it.

api/v1alpha1/nicclusterpolicy_types.go
rollandf: Removed it

pkg/state/state_spectrumx_test.go
rollandf: catalog is a var, not the default one.
Updated to use it in all places

almaslennikov (approved):

e0ne (approved):

PR #1548, opened 2025-06-09
feat: Make Network Operator working in STIG-Enabled Kubernetes Cluster

e0ne:


deployment/network-operator/charts/sriov-network-operator/templates/operator.yaml
```diff
@@ -39,14 +39,22 @@ spec:
       {{- end }}
       {{- end }}
       containers:
-        - name: {{ include "sriov-network-operator.fullname" . }}
+        - name: sriov-operator-pod
```
almaslennikov: Why do we need this change?

deployment/network-operator/charts/sriov-network-operator/templates/operator.yaml
```diff
@@ -39,14 +39,22 @@ spec:
...
           image: {{ .Values.images.operator }}
           command:
             - sriov-network-operator
           resources:
             requests:
               cpu: 100m
               memory: 100Mi
+          securityContext:
```
rollandf: Any change in the SRIOV-NO Helm chart will be override by our downstream fork CI on next beta.
We should do the same in the Mellanox fork

e0ne: I'll move this change to sriov operator repo

deployment/network-operator/charts/sriov-network-operator/templates/operator.yaml
e0ne: We don't. I'll revert it

manifests/state-multus-cni/0050-multus-ds.yml
```diff
@@ -21,6 +21,8 @@ spec:
         name: multus
     spec:
       hostNetwork: true
+      securityContext:
+        privileged: true
```
heyvister1, line 25: nit. Does Multus requires all host capabilities? or we could limit them to specific ones?

e0ne: we can limit them in the future and leave only required capabilities

PR #1549, opened 2025-06-10
chore: update go version to 1.24

rollandf:
- Bump go version to 1.24
- Bump golangci-lint v1.64.8 to align with go version
- Update golangci.yml to align with golangci-lint v1.64.8 and go version 1.24
- Fix linter issues found after updating  golangci-lint

### DIFF
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -11,7 +11,7 @@ jobs:
       - name: Set up Go 1.x
         uses: actions/setup-go@v5
         with:
-          go-version: '1.23'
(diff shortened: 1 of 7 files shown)


heyvister1 (approved):
Thanks @rollandf 
Because this PR is not only bumping golang version, can you please add rest of your changes in PR description?

rollandf:
> Thanks @rollandf Because this PR is not only bumping golang version, can you please add rest of your changes in PR description?

Done

almaslennikov (approved):

PR #1584, opened 2025-06-26
feat: support deployment toleration and affinity

### LINKED ISSUE #1172: Tolerations are not propagated to nvipam controller
**What happened**: 
We required all network-operator pods to be deployed on nodes with taints. We added toleration in operator spec as well as in Toleration section for daemon sets. But These tolerations are not propagated to nvipam-controller. nvipam controller remains in pending state

**What you expected to happen**:
tolerations to propagate to nvipam controller also

**How to reproduce it (as minimally and precisely as possible)**:
Deploy network operator on cluster with all nodes tainted. Add toleration in helm chart. If nvipam controller deployment is true, its pods will remain in pending state

**Anything else we need to know?**:

**Logs**:
- NicClusterPolicy CR spec and state:
- Output of: `kubectl -n nvidia-network-operator get -A`:
- Network Operator version:
- Logs of Network Operator controller:
- Logs of the various Pods in `nvidia-network-operator` namespace:
- Helm Configuration (if applicable):
- Kubernetes' nodes information (labels, annotations and status): `kubectl get node -o yaml`:

**Environment**:
- Kubernetes version (use `kubectl version`): N/A
- Hardware configuration: N/A
  - Network adapter model and firmware version:
- OS (e.g: `cat /etc/os-release`): N/A
- Kernel (e.g. `uname -a`):N/A
- Others:N/A

rollandf:
Fixes #1172


manifests/state-ib-kubernetes/0070-deployment.yaml
```diff
@@ -55,6 +55,21 @@ spec:
                   - key: "node-role.kubernetes.io/control-plane"
                     operator: In
                     values: [ "" ]
+            {{- if .DeploymentNodeAffinity }}
```
heyvister1, line 58: nit: Consider using the following, to reduce nested conditions
```suggestion
{{- with .DeploymentNodeAffinity }}
  {{- with .PreferredDuringSchedulingIgnoredDuringExecution }}
    {{- range . }}
    - weight: {{ .Weight }}
      preference:
        {{ .Preference | toYaml | nindent 16 }}
    {{- end }}
  {{- end }}
{{- end }}
```

manifests/state-ib-kubernetes/0070-deployment.yaml
```diff
@@ -55,6 +55,21 @@ spec:
...
+            {{- if .DeploymentNodeAffinity.PreferredDuringSchedulingIgnoredDuringExecution }}
+            {{- range .DeploymentNodeAffinity.PreferredDuringSchedulingIgnoredDuringExecution }}
+            - weight: {{ .Weight }}
+              preference:
+                {{ .Preference | yaml | nindent 16 }}
+            {{- end }}
+            {{- end }}
+            {{- end }}
+          {{- if .DeploymentNodeAffinity }}
```
heyvister1, line 67: Do we need to check this twice? Its already being checked above `{{- if .DeploymentNodeAffinity }}`

manifests/state-ib-kubernetes/0070-deployment.yaml
rollandf: Seems the same number of conditions, no?

manifests/state-ib-kubernetes/0070-deployment.yaml
rollandf: Not in the same block

almaslennikov (approved):

manifests/state-ib-kubernetes/0070-deployment.yaml
heyvister1: Ok. I thought it was more readable

heyvister1 (approved):

PR #1585, opened 2025-06-27
chore: add logic for release staging

almaslennikov:
release.yaml now creates a new staging branch for each release.
Managed components will push version updates to this branch (PR TBD)
staging.yaml will trigger on each push to staging and validate that all expected components have updated their versions. If yes, it will squash changes and create a PR to the base branch
TODO: update release-pr-checker.yaml to delete the staging branch and create a tag
TODO: add workflow to merge PRs to staging


.github/workflows/staging.yaml
```diff
@@ -0,0 +1,127 @@
...
+          VALIDATION_FAILED=false
+          
+          # Get all components with sourceRepository field
+          versions=$(yq 'to_entries | map(select(.value.sourceRepository != null) | .value.version) | .[]' hack/release.yaml)
+          echo "Versions: $versions"
+          
+          # Check each version
+          for version in $versions; do
+            if [ "$version" != "$EXPECTED_VERSION" ]; then
+              echo "::error::Version mismatch found"
```
maze88, line 53: this won't be clear which component doesn't have the expected version... even though easy enough to lookup manually when troubleshooting.

.github/workflows/release.yaml
```diff
@@ -47,11 +47,17 @@ jobs:
               git push -u origin $BASE_BRANCH
             fi
           fi
-      - run: |
+      - id: create-staging-branch
```
maze88, line 50: could there be a case that the staging branch already exists? how would this handle that?

PR #1587, opened 2025-06-29
chore: simplify release CI

almaslennikov:
release.yaml now creates a single PR to release a new version. It creates release tags and branches in downstream components, then waits until all required images are available on the ngc staging. After that, it updates images versions in release.yaml, and pulls updated helm charts from sriov-network-operator and NFD.

Pros:
* A single run of CI is needed to test the release
* No PRs from downstream components
* Fork CI workflows now only need to build and push new images

Cons:
* None

### DIFF
--- a/.github/workflows/release-pr-checker.yaml
+++ b/.github/workflows/release-pr-checker.yaml
@@ -65,7 +65,7 @@ jobs:
               sleep $WAIT_INTERVAL
             fi
           done
-  update-network-operator-version:
+  merge-release-pr:
     needs:
       - wait-for-ci
     if: startsWith(github.event.pull_request.title, 'cicd:')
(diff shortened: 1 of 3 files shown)


.github/workflows/release-pr-checker.yaml
```diff
@@ -87,3 +87,33 @@ jobs:
...
+          TARGET_BRANCH=$(gh pr view $PR_NUMBER --json baseRefName -q .baseRefName)
+
+          if [[ "$SOURCE_BRANCH" =~ ^staging- ]]; then
+            # Extract version from staging branch name (staging-v25.4.0-rc.1 -> v25.4.0-rc.1)
+            VERSION=${SOURCE_BRANCH#staging-}
+            echo "Creating release: $VERSION"
+            echo "Target branch (release source): $TARGET_BRANCH"
+
+            # Determine if this is a pre-release (has '-' after version numbers)
+            if [[ "$VERSION" =~ ^v[0-9]+\.[0-9]+\.[0-9]+- ]]; then
```
maze88: isn't it enough (and more readable) to simply `if echo $VERSION | grep '-'; then` here?

.github/workflows/release.yaml
```diff
@@ -47,89 +59,227 @@ jobs:
...
+          repository: ${{ github.repository_owner }}/node-feature-discovery
+          path: node-feature-discovery
+          ref: ${{ needs.determine-versions.outputs.release_branch }}
+      - uses: actions/checkout@v4
+        with:
+          repository: ${{ github.repository_owner }}/network-operator
+          path: network-operator
+      - name: Create staging branch and update component versions
+        run: |
+          cd network-operator
```
maze88, line 210: i believe the best practice in flows like this is to use `pushd` (and terminate with `popd`).

maze88 (commented):
nice approach, i'd like to understand it better if/when we hold a meeting and you could present it or walk us through.
added some minor/"nit" comments.

.github/workflows/release.yaml
almaslennikov: There is only a single cd in this stage

PR #1615, opened 2025-07-06
fix: missing conditional in assignment operator for BUNDLE_OCP_VERSIONS

maze88:
bugfix:
the `?=` assignment operator is required for shell env var overrides of values, as used in other variables here.


rollandf:
@maze88 you have a failing CI check. Please fix.

PR #1616, opened 2025-07-06
chore: Adding '.Values.global.imagePullSecrets' 

heyvister1:


deployment/network-operator/values.yaml
```diff
@@ -320,8 +320,13 @@ operator:
 
 # -- An optional list of references to secrets to use for pulling any of the
 # Network Operator images.
+# Kept for backward compatibility.
 imagePullSecrets: []
 
+# -- Global values to be propagated into Network-Operator sub-charts
+global:
```
rollandf, line 327: The format seems different between charts. May be problematic.
Some uses:
```
imagePullSecrets:
- name: my-secret
```
Other uses:
```
imagePullSecrets:
-  my-secret
```

PR #1628, opened 2025-07-10
fix: webhook ofed validation doca prefix

rollandf:


api/v1alpha1/validator/nicclusterpolicy_webhook.go
```diff
@@ -720,8 +720,8 @@ func isValidOFEDVersion(version string) bool {
 	// Support multiple OFED version formats:
 	// 1. Old scheme: "23.10-0.2.2.0" or "23.10-0.2.2.0.1"
```
e0ne, line 721: nic: we can drop old scheme

e0ne (approved):

almaslennikov (approved):

PR #1641, opened 2025-07-15
chore: Update Maintenance OP Helm sub-chart

heyvister1:
`maintenanceoperatorconfigs` can be deployed by other k8s operators, utilizing Maintennace OP. Using the following change we can control whether to deploy this object or not.

* Update maintenance OP Helm sub-chart.
* Add support in values.yaml to control 'maintenanceoperatorconfigs' deployment


rollandf (approved):
Question: reusable workflow are not supposed to update charts also?
@almaslennikov

PR #1643, opened 2025-07-16
ofed-driver: Adjust mount path for Ubuntu system CA certificates

sudharsanomprakash:
This PR adds a postStart lifecycle hook in the MOFED container to run update-ca-certificates (Ubuntu/Debian) or update-ca-trust (RHEL/SLES), ensuring system trust store picks up any mounted CA certs.

Tested on: Ubuntu 24.04

This avoids the need to rebuild the image just to inject CA certs.

### DIFF
--- a/pkg/state/state_ofed.go
+++ b/pkg/state/state_ofed.go
@@ -114,7 +114,7 @@ const (
 //
 //nolint:lll
 var CertConfigPathMap = map[string]string{
-	"ubuntu": "/etc/ssl/certs",
+	"ubuntu": "/usr/local/share/ca-certificates",
 	"rhcos":  "/etc/pki/ca-trust/extracted/pem",
 	"rhel":   "/etc/pki/ca-trust/extracted/pem",
 	"sles":   "/etc/ssl",
--- a/pkg/state/state_ofed_test.go
+++ b/pkg/state/state_ofed_test.go
@@ -1265,7 +1265,7 @@ func verifyAdditionalMounts(mounts []v1.VolumeMount) {
 	cert := v1.VolumeMount{
 		Name:             "cert-cm",
 		ReadOnly:         true,
-		MountPath:        "/etc/ssl/certs/my-cert",
+		MountPath:        "/usr/local/share/ca-certificates/my-cert",
 		SubPath:          "my-cert",
 		MountPropagation: nil,
 		SubPathExpr:      "",


sudharsanomprakash:
@rollandf could you please review and approve this PR? Thank you

manifests/state-ofed-driver/0050_ofed-driver-ds.yaml
```diff
@@ -143,6 +143,21 @@ spec:
...
           {{- end }}
           {{- end }}
+          lifecycle:
+            postStart:
+              exec:
+                command:
+                  - /bin/sh
+                  - -c
+                  - |
+                    echo "[INFO] Updating system CA certificates..."; \
```
rollandf: Thanks for the PR!

Can you explain what is the issue you want to fix?
- In which flow it happens? Is it only if user configure additional 'certConfig' mount?
- What are the errors we get?
- Why not add this handling in [entrypoint](https://github.com/Mellanox/doca-driver-build/blob/main/entrypoint.sh) ?
- Will it work in disconnected env?
- We support also SLES, what it the relevant command?
- What happen if 'post-start' fails or time out?
- Why it the path change for Ubuntu needed? '`/usr/local/share/ca-certificates`

sudharsanomprakash: @rollandf 

- In which flow it happens? Is it only if user configure additional 'certConfig' mount?
In Ubuntu-based deployments, when a custom CA is mounted via the certConfig option in the `NicClusterPolicy`, it is currently mounted under `/etc/ssl/certs`. However, Ubuntu expects custom CA certificates to be placed in `/usr/local/share/ca-certificates`, and they must be activated by running update-ca-certificates. Without this step, the system does not trust the custom CA, and any TLS-based communication fail with certificate verification errors.

- What are the errors we get?
`Certificate verification failed: The certificate is NOT trusted. The certificate issuer is unknown.`  

- Why not add this handling in [entrypoint](https://github.com/Mellanox/doca-driver-build/blob/main/entrypoint.sh) ?
Initially considered this but decided against embedding environment-specific logic directly in the generic entrypoint.sh. Instead postStart ensures the CA update occurs early and independently of other init logic. That said, I can move it to entrypoint.sh if the project prefers all initialization logic to be centralized.

- Will it work in disconnected env?
Yes, tested and it works.

- We support also SLES, what it the relevant command?
This PR focuses on Ubuntu, which is the primary base in our deployments. I can look into adding SLES support in a separate PR since it requires a different path (`/etc/pki/trust/anchors/`) and flow.

- What happen if 'post-start' fails or time out?
Kubernetes will delay pod readiness until postStart completes. If the command hangs or fails, the pod will remain in a non-ready state.

- Why it the path change for Ubuntu needed? `/usr/local/share/ca-certificates`
This is the standard location for custom CA certificates on Ubuntu. It allows the container to retain system defaults in `/etc/ssl/certs` and append custom certs safely.

rollandf: Thanks for the reply!
I propose to split this PR.
Fix here the path to `/usr/local/share/ca-certificates`
And handle in `entrypoint` the update.

WDYT?

sudharsanomprakash: @rollandf  Thanks for the suggestion!
I’ve updated this PR to:
- Fix the certificate path to /usr/local/share/ca-certificates

- Move the update-ca-certificates logic into the entrypoint.sh for Ubuntu-based containers https://github.com/Mellanox/doca-driver-build/pull/104

- SLES support and update-ca-trust handling will be addressed in a follow-up PR.

Let me know if anything further needs adjustment!

rollandf:
@sudharsanomprakash thanks for the updated PR.
Could you please squash the commits to a single one, and check the failing unit-test?

rollandf:
@almaslennikov @e0ne @heyvister1 PTAL

PR #1648, opened 2025-07-21
fix: revert maintenance OP sub-chart 'imagePullSecrets' 

heyvister1:
Updating maintenance OP sub-chart changes
Mellanox/maintenance-operator#108


heyvister1:
Until we'll add auto-sync for maintenance-operator sub-charts in CI
cc @almaslennikov

e0ne (approved):

PR #1649, opened 2025-07-21
fix: add /dev mount to nic config daemon

almaslennikov:
This mount is needed for access to rshim devices


rollandf (commented):
What about sub chart? Until we remove it, should be also updated.

https://github.com/Mellanox/network-operator/blob/master/deployment/network-operator/charts/nic-configuration-operator-chart/templates/config-daemon.yaml

rollandf (approved):

PR #1650, opened 2025-07-21
chore: adding nodeMaintenance cleanup in upgrade-controller

heyvister1:
In case requestor-mode is set in driver's upgrade controller, there was a missing flow to cleanup existing `nodeMaintenance` objects, created by current requestor. 
New changes will cleanup `nodeMaintenance` objects once either `NicClusterPolicy` or `OFEDDriverSpec` are removed

### DIFF
--- a/controllers/upgrade_controller.go
+++ b/controllers/upgrade_controller.go
@@ -83,6 +83,10 @@ func (r *UpgradeReconciler) Reconcile(ctx context.Context, _ ctrl.Request) (ctrl
 
 	if err != nil {
 		if errors.IsNotFound(err) {
+			// Cleanup existing upgrade related resources
+			if err := r.cleanupUpgradeResources(ctx); err != nil {
+				return ctrl.Result{}, err
+			}
 			return ctrl.Result{}, nil
 		}
 		return ctrl.Result{}, err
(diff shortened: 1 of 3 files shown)


controllers/upgrade_controller.go
```diff
@@ -99,8 +103,8 @@ func (r *UpgradeReconciler) Reconcile(ctx context.Context, _ ctrl.Request) (ctrl
 		nicClusterPolicy.Spec.OFEDDriver.OfedUpgradePolicy == nil ||
 		!nicClusterPolicy.Spec.OFEDDriver.OfedUpgradePolicy.AutoUpgrade {
 		reqLogger.V(consts.LogLevelInfo).Info("OFED Upgrade Policy is disabled, skipping driver upgrade")
-		err = r.removeNodeUpgradeStateLabels(ctx)
-		if err != nil {
+		// Cleanup existing upgrade related resources
+		if err := r.cleanupUpgradeResources(ctx); err != nil {
```
almaslennikov, line 107: What will happen if we undeploy the network operator?

heyvister1: If you remove/scale down NO deployment, exiting nodeMaintenace objects will not get deleted, same NCP which exists

almaslennikov: Let's add an item to backlog to address that in the future

controllers/upgrade_controller.go
```diff
@@ -192,6 +217,66 @@ func (r *UpgradeReconciler) removeNodeUpgradeStateAnnotations(ctx context.Contex
...
+	ownedNMCount := 0
+	for i := range nodeMaintenanceList.Items {
+		nm := &nodeMaintenanceList.Items[i]
+
+		// Only delete objects created by our requestor ID
+		if nm.Spec.RequestorID == requestorOpts.MaintenanceOPRequestorID {
+			reqLogger.V(consts.LogLevelInfo).Info("Deleting NodeMaintenance object",
+				"name", nm.Name, "namespace", nm.Namespace, "nodeName", nm.Spec.NodeName)
+
+			ownedNMCount++
```
e0ne: we need to increase this counter only if no error ocured

rollandf (approved):
LGTM, after addressing @e0ne comment

controllers/upgrade_controller.go
heyvister1: You're right. Done

e0ne (approved):
Thanks for addressing my comments

Issue #952, opened 2024-05-22
Support probe timeout configuratoins


ivelichkovich:
**What would you like to be added**:

Support for timeouts on probe configurations, specifically for ofed-driver

**Why is this needed**:

lsmod can take longer than a second depending on what you have running on the host

rollandf:
The MOFED probe timeouts are already available to be customized via the NicClusterPolicy CR.
See here:
https://github.com/Mellanox/network-operator/blob/master/api/v1alpha1/nicclusterpolicy_types.go#L90-L94

Please tell me if that meets your requirement.

rollandf:
@ivelichkovich BTW, I saw you have interest in IPAM solution with slicing per node.
You can check https://github.com/Mellanox/nvidia-k8s-ipam that implements the same.

ivelichkovich:
> @ivelichkovich BTW, I saw you have interest in IPAM solution with slicing per node. You can check https://github.com/Mellanox/nvidia-k8s-ipam that implements the same.

Oh that's awesome, I'll explore that repo!

ivelichkovich:
> The MOFED probe timeouts are already available to be customized via the NicClusterPolicy CR. See here: https://github.com/Mellanox/network-operator/blob/master/api/v1alpha1/nicclusterpolicy_types.go#L90-L94
> 
> Please tell me if that meets your requirement.

so these do allow you to define the probes however PodProbeSpec only exposes these fields https://github.com/Mellanox/network-operator/blob/master/api/v1alpha1/nicclusterpolicy_types.go#L74 so having access to failureThreshold and timeoutSeconds would be nice to have. You could maybe just replace `PodProbeSpec` with the upstream probe object and then it would have 1:1 parity without needing to keep a copy of the struct in code.

rollandf:
Fixed in #1651

PR #1651, opened 2025-07-21
Support OFED probe configuration

sergey-arefev:
As described in the [issue #952](https://github.com/Mellanox/network-operator/issues/952): the liveness probe (especially in a cloud environment) can fail for reasons unrelated to OFED—for example, a burst in network SSD usage—which can interrupt a long-running learning job.

These changes allow completely disabling liveness (or other) probes to eliminate such painful cases or to configure them less strictly.

The default configuration stays the same, but IMHO, it's better to provide an easy way to disable the liveness probe for default installations.


pkg/state/state_ofed.go
```diff
@@ -841,24 +864,40 @@ func (e envVarsWithGet) Get(name string) *v1.EnvVar {
...
+	sanitizeProbeSpec(cr.Spec.OFEDDriver.ReadinessProbe, defaultProbeSpec)
+}
+
+// sanitizeProbeSpec checks and adjusts probe spec values to meet Kubernetes requirements
+func sanitizeProbeSpec(probeSpec *mellanoxv1alpha1.PodProbeSpec, defaultProbeSpec mellanoxv1alpha1.PodProbeSpec) {
+	if !probeSpec.Enabled {
+		return
+	}
+	if probeSpec.InitialDelaySeconds < 1 {
+		probeSpec.InitialDelaySeconds = defaultProbeSpec.InitialDelaySeconds
```
e0ne, line 956: this validation should be defined as `// +kubebuilder:validation:Minimum:=1` in CRD definition

manifests/state-ofed-driver/0050_ofed-driver-ds.yaml
```diff
@@ -143,33 +143,39 @@ spec:
...
           {{- end }}
           {{- end }}
+          {{- if .CrSpec.StartupProbe.Enabled }}
           startupProbe:
             exec:
               command:
                 [sh, -c, 'ls /run/mellanox/drivers/.driver-ready']
             initialDelaySeconds: {{ .CrSpec.StartupProbe.InitialDelaySeconds }}
-            failureThreshold: 60
+            failureThreshold: {{ .CrSpec.StartupProbe.FailureThreshold }}
```
e0ne, line 151: it's good to have this configurable

manifests/state-ofed-driver/0050_ofed-driver-ds.yaml
```diff
@@ -143,33 +143,39 @@ spec:
             {{- end }}
           {{- end }}
           {{- end }}
+          {{- if .CrSpec.StartupProbe.Enabled }}
```
e0ne: IMO, we don't need an option to disable startup probe

rollandf: Startup probe is important. It signal if drivers are built and loaded.
It affects other components (SRIOV-NO...)

pkg/state/state_ofed.go
sergey-arefev: I decided not to put it in the CRD to allow usage of any previous configs without these new fields
no config in NicPolicy -> 0 as default for int -> default values will be used, as before

manifests/state-ofed-driver/0050_ofed-driver-ds.yaml
sergey-arefev: fixed

rollandf:
I added some test in #1652 
Can you rebase on top, and see that tests are still passing.
Also please add some cases for the new flows

api/v1alpha1/nicclusterpolicy_types.go
```diff
@@ -79,10 +79,16 @@ type ImageSpecWithConfig struct {
 
 // PodProbeSpec describes a pod probe.
 type PodProbeSpec struct {
+	// Disabled indicates if the probe should be disabled (ignored for startup probe)
+	Disabled bool `json:"disabled,omitempty"`
```
e0ne: It'a still not clear to me why we need to disable probe instead of configuring timeouts and threshold

rollandf: Maybe in the case you don't want the probe at all?
@sergey-arefev is it the use case?

sergey-arefev: yes, any higher values for the probe (threshold and timings) just make the issue less probable, but if we use the network-operator to prepare the node and never unload ib modules, we don't need a liveness probe at all

sergey-arefev: @e0ne [initial commit](https://github.com/Mellanox/network-operator/commit/ea025f3d60a185ac44e79926040d65133920e3f2) that introduced liveness probes doesn't contain much explanation of what real cases could cause drivers to be gone, so I would like to have an option to disable the liveness probe completely

if you have a strong opinion that there shouldn't be such an option, I'm ok to remove it from the PR and use only FailureThreshold, but some configuration is necessary

sergey-arefev: @e0ne please clarify: should I remove `Disabled` from `PodProbeSpec`, or is it possible to merge the current version of the PR?

e0ne: @sergey-arefev I prefer to not introduce `Disable` option to keep our API  less complex

sergey-arefev: fixed

rollandf (approved):

rollandf:
@sergey-arefev thanks for updating the PR!

can you please rebase from master, squash your commits to a single one and run `make test` locally?

PR #1652, opened 2025-07-22
chore: add tests for ofed probe

rollandf:


almaslennikov (approved):
CI failure is probably unrelated

PR #1666, opened 2025-07-29
fix: update maintenance-operator 'imagePullSecrets' in values.yaml

heyvister1:


deployment/network-operator/values.yaml
```diff
@@ -202,13 +202,13 @@ maintenance-operator-chart:
   operatorConfig:
     # -- deploy MaintenanceOperatorConfig CR
     deploy: false
+  # imagePullSecrets for Maintenance Operator related images
+  #imagePullSecrets: []
```
rollandf: There is trailing spaces.
Please run 'make release-build'

rollandf (changes requested):

PR #1672, opened 2025-07-30
fix: remove probes for spectrumx

rollandf:
Since the pod is hostnetwork, ports can conflicts.


manifests/state-spectrum-x-operator/0050-flowcontroller-ds.yaml
```diff
@@ -37,8 +39,8 @@ spec:
         - command:
             - /flowcontroller
           args:
-            - --metrics-bind-address=:8082
-            - --health-probe-bind-address=:8084
+            - --metrics-bind-address=0
```
heyvister1, line 42: Are we ok with skipping liveness/readiness probe checks here? what about using 30000+ custom port in that case?

rollandf: I consulted with Michael F. He approved.
Any port could potentially be a conflict, and we don't have a way to specify them in the NCP.

heyvister1: Ok. We might be able to use check a range of ports available in host and pick one dynamically (e.g. using initContainer) and then use it for libeness/readiness probes.
This could be a suggestion for next release

heyvister1 (approved):
We should followup with a solution to be able to use readiness/liveness probes even if we're using hostNetwork

manifests/state-spectrum-x-operator/0050-flowcontroller-ds.yaml
rollandf: Not sure you can update the probe dynamically, it is part of the pod definition

PR #1673, opened 2025-07-30
chore: align master to rc images

rollandf:


heyvister1 (approved):

deployment/network-operator/charts/sriov-network-operator/templates/role.yaml
```diff
@@ -28,6 +28,15 @@ rules:
       - statefulsets
     verbs:
       - '*'
+  - apiGroups:
```
heyvister1, line 31: How is this related to aligning master images?
Isn't sriov charts being updated using our ci workflows?

rollandf: It is for releases.
I took same changes as in : https://github.com/Mellanox/network-operator/pull/1659

Now that we have release branch, master is not updated automatically

PR #1683, opened 2025-08-05
feat: add support for image building with a custom GOPROXY and builder

e0ne:


Dockerfile
```diff
@@ -35,6 +35,9 @@ COPY ./ ./
 ARG ARCH
 ARG LDFLAGS
 ARG GCFLAGS
+ARG GOPROXY=direct
+ENV GOPROXY=$GOPROXY
```
heyvister1, line 39: @e0ne  Shouldn't we use custom proxy configuration here?
GOPROXY=https://corporate-proxy.com

e0ne: no, we should use safe default. it will be overwritten by passing env variable with a value from a secret during the build procedure

.github/workflows/ci.yaml
```diff
@@ -16,7 +16,7 @@ env:
 
 jobs:
   docker-build-push:
-    runs-on: ubuntu-24.04
+    runs-on: linux-amd64-cpu4
```
maze88, line 19: there are many utilities that are pre-installed on the public runner... have you performed a successful test run?

e0ne: how can we test it without merge?

maze88 (approved):

.github/workflows/ci.yaml
maze88: simulate a trigger even on your fork, but i can be complicated. let's merge and see how it performs (and patch any missing utilities we require in the CI workflow).

i've also asked in advanced in their slack channel about what tools are pre-installed on the private runner.

PR #1688, opened 2025-08-07
feat: Use base distroless image

e0ne:


almaslennikov (approved):
Do we want this change in the 25.07?

PR #1703, opened 2025-08-11
fix:  prevent duplicate NetworkAttachmentDefinition CRD when SRIOV enabled

dahalperin:
- Move all CRDs from crds/ to templates/crds/ for Helm management
- Add condition to apply CRDs only when sriovNetworkOperator.enabled is false
- Update Dockerfile to create crds directory structure
- Remove args from upgrade-crd-hook as Helm now handles network operator CRDs


deployment/network-operator/templates/upgrade-crd-hook.yaml
```diff
@@ -83,7 +83,6 @@ spec:
           command:
             - /apply-crds
           args:
-              - --crds-dir=/crds/network-operator
```
rollandf, line 94: I think we need this line

rollandf: Check indentation here

e0ne:
nic_operator_helm CI failure looks valid

PR #1747, opened 2025-10-05
fix: add default DS tolerations for Node filtering

rollandf:
Add default DS tolerations for DOCA OFED DS Node filtering

### DIFF
--- a/pkg/state/state_ofed.go
+++ b/pkg/state/state_ofed.go
@@ -465,12 +465,54 @@ func (s *stateOFED) handleAdditionalMounts(
 	return nil
 }
 
-// addGPUToleration adds the nvidia.com/gpu:NoSchedule toleration if not already present
-func addGPUToleration(tolerations []v1.Toleration) []v1.Toleration {
-	gpuToleration := v1.Toleration{
-		Key:      "nvidia.com/gpu",
-		Effect:   v1.TaintEffectNoSchedule,
-		Operator: v1.TolerationOpExists, // Operator can be Exists if Value is not specified, or Equal if Value is specified.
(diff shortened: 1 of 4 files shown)


pkg/state/state_ofed.go
```diff
@@ -465,29 +465,76 @@ func (s *stateOFED) handleAdditionalMounts(
...
-// addGPUToleration adds the nvidia.com/gpu:NoSchedule toleration if not already present
-func addGPUToleration(tolerations []v1.Toleration) []v1.Toleration {
-	gpuToleration := v1.Toleration{
-		Key:      "nvidia.com/gpu",
-		Effect:   v1.TaintEffectNoSchedule,
-		Operator: v1.TolerationOpExists, // Operator can be Exists if Value is not specified, or Equal if Value is specified.
+// addDefaultDaemonSetTolerations adds tolerations that the DaemonSet controller automatically adds at runtime.
+// These tolerations must be included in the filtering logic to avoid incorrectly filtering out nodes
+// that the DaemonSet pods will actually be able to schedule on.
+// See: https://kubernetes.io/docs/concepts/workloads/controllers/daemonset/#taints-and-tolerations
```
heyvister1, line 471: Should we add a way to track default tolerations changes, according to supported k8s GA versions?
I mean fetch html file and parsing its default tolerations content, update `defaultTolerations` code through makefile

rollandf: I think it is an overkill, and we cannot rely on documentation to keep a certain structure

heyvister1: We could fetch it from GH code base though. 
Anyway, looks like this code hasn't been changed lately

e0ne (approved):

PR #1757, opened 2025-10-10
feat: adding internal drain requestor code

heyvister1:


deployment/network-operator/values.yaml
```diff
@@ -272,10 +272,18 @@ operator:
...
   maintenanceOperator:
+    # k8s namespace to be used for generated nodeMaintenance objects
+    nodeMaintenanceNamespace: default  
+    # enable upgrade controller requestor
     useRequestor: false
-    requestorID: "nvidia.network.operator"
+    # upgrade controller requestor ID to be used in nodeMaintenance objects
+    requestorID: "nvidia.network-operator-driver-upgrade-controller"
+    # nodeMaintenance name prefix to be used by upgrade-controller (default is "")
     nodeMaintenanceNamePrefix: "network-operator"
```
almaslennikov, line 282: Do we need to use the same name for node maintenance in nic config operator?

heyvister1: Only in case users would like to coordinate node operations. In case it uses a different name than E2E node operation will be performed again

PR #1793, opened 2025-10-28
chore: update OCP version

rollandf:
Even that we moved to File Based Catalog, when publishing a bundle the `com.redhat.openshift.versions` metada is checked. It does not affect the publishing.

Note that the publishing is done in a different PR, where the OCP versions are specified.

See [validation](https://github.com/redhat-openshift-ecosystem/certification-releases/blob/main/4.9/ga/troubleshooting.md#get-supported-versions).

See failing bundle [PR](https://github.com/redhat-openshift-ecosystem/certified-operators/pull/6418#issuecomment-3211853950) without the annotation.

image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -50,7 +50,7 @@ IMAGE_NAME?=network-operator
 CONTROLLER_IMAGE=$(REGISTRY)/$(IMAGE_NAME)
 IMAGE_BUILD_OPTS?=
 BUNDLE_IMG?=network-operator-bundle:$(VERSION)
-BUNDLE_OCP_VERSIONS?=v4.15-v4.18
+BUNDLE_OCP_VERSIONS?=v4.17-v4.20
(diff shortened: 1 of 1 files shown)


bundle/metadata/annotations.yaml
```diff
@@ -13,3 +13,4 @@ annotations:
   # Annotations for testing.
   operators.operatorframework.io.test.mediatype.v1: scorecard+v1
   operators.operatorframework.io.test.config.v1: tests/scorecard/
+  com.redhat.openshift.versions: v4.15-v4.18
```
e0ne, line 16: should it be `v4.17-v4.20`?

e0ne (changes requested):

bundle/metadata/annotations.yaml
rollandf: It will be updated on next bundle build (beta4)

PR #1809, opened 2025-11-03
feat: manage nvidia.com/operator.nic-configuration.wait label

almaslennikov:
When NIC Configuration Operator is disabled, manage the label similar to how the mofed.wait label is managed

### DIFF
--- a/controllers/nicclusterpolicy_controller.go
+++ b/controllers/nicclusterpolicy_controller.go
@@ -108,6 +108,8 @@ type NicClusterPolicyReconciler struct {
 
 // Reconcile is part of the main kubernetes reconciliation loop which aims to
 // move the current state of the cluster closer to the desired state.
+//
+//nolint:gocognit,funlen
 func (r *NicClusterPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	// Wait for migration flow to finish
 	select {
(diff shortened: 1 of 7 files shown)


controllers/nicclusterpolicy_controller_test.go
```diff
@@ -305,6 +320,16 @@ var _ = Describe("NicClusterPolicyReconciler Controller", func() {
...
 			}, timeout*3, interval).Should(BeTrue())
 
+			By("Verify Nic Configuration Wait Label is false")
+			Eventually(func() bool {
+				n := &corev1.Node{}
+				err = k8sClient.Get(context.TODO(), types.NamespacedName{Namespace: node.GetNamespace(), Name: node.GetName()}, n)
+				if err != nil {
+					return false
+				}
+				return n.ObjectMeta.Labels[nodeinfo.NodeLabelWaitNicConfig] == "false"
```
heyvister1, line 330: Shouldn't we use NIC configuration OP const in this case? Instead of duplicating it
https://github.com/Mellanox/nic-configuration-operator/pull/182/files#diff-6572dddc535b4d2c98746d71dde97840af1328797e193db28376939a320be8f8R141

controllers/nicclusterpolicy_controller.go
```diff
@@ -258,29 +282,54 @@ func (r *NicClusterPolicyReconciler) handleMOFEDWaitLabelsNoConfig(ctx context.C
...
+	}, client.RawPatch(types.StrategicMergePatchType, patch))
+
+	if err != nil {
+		return errors.Wrapf(err, "unable to patch %s label for node %s", label, node)
+	}
+	return nil
+}
+
 // set the value for the OFED wait label, remove the label if the value is ""
 func setOFEDWaitLabel(ctx context.Context, c client.Client, node, value string) error {
```
rollandf: Can we replace uses of this function with the new `setNodeLabel` ?
Very similar code

controllers/nicclusterpolicy_controller_test.go
almaslennikov: Currently we don't have a dependency on the nic op. I suggest not to add it for one constant

controllers/nicclusterpolicy_controller.go
almaslennikov: Thanks, deleted the `setOFEDWaitLabel` func

rollandf (approved):

controllers/nicclusterpolicy_controller_test.go
heyvister1: Ok

PR #1811, opened 2025-11-03
feat: Add sriov-network-operator-config-daemon-stig-fips image into release.yaml

e0ne:
…elease.yaml

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -243,6 +243,12 @@ jobs:
             if [ "$SOURCE_REPO" = "null" ]; then
               continue
             fi
+          
+            # We always need to add '-stig-fips' suffix for corresponding image version
+            # TODO: do not DOCA Driver image for now because $SOURCE_REPO is null one Network Operator will support a new version format 
+            if [[ "$component" == *StigFips ]]; then
+              COMPONENT_TAG="${COMPONENT_TAG}-stig-fips"
+            fi 
             
             echo "Updating component: $component"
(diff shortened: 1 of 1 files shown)


hack/release.yaml
```diff
@@ -24,6 +24,11 @@ SriovConfigDaemon:
   repository: nvcr.io/nvstaging/mellanox
   sourceRepository: sriov-network-operator
   version: network-operator-v25.10.0-beta.4
+SriovConfigDaemonStigFips:
+  image: sriov-network-operator-config-daemon-stig-fips
+  repository: nvcr.io/nvstaging/mellanox
+  sourceRepository: sriov-network-operator
+  version: network-operator-v25.10.0-beta.4-stig-fips
```
almaslennikov, line 31: The `-stig-fips` version tag will get overwritten by https://github.com/Mellanox/network-operator/blob/master/.github/workflows/release.yaml#L253
I suggest we add a `stig-fips` flag here in the release.yaml and handle it in the release workflow

almaslennikov: We can check for this flag like we do for helm chart: https://github.com/Mellanox/network-operator/blob/master/.github/workflows/release.yaml#L262
Here we need to also export the tag for stig-fips flavor: https://github.com/Mellanox/network-operator/blob/master/.github/workflows/release.yaml#L36

maze88: For consistency - in several of our internal CI jobs (e.g. NGC promotion pipeline), we include logic for handling such _StigFips_ or "_-stig-fips_" variants.

PR #1818, opened 2025-11-04
fix: update component to MofedStigFips and use local var in loop for setting component versions in release workflow

maze88:
changes:
- fix: update component MofedStig to MofedStig**Fips** (line 300).
- fix: use separate local variable when determining component tag in release workflow (lines 250-259).
- fix: set '-stig-fips' component tag when relevant when waiting for images.
- chore: cleanup of trailing whitespace (rest of diff).

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -179,7 +179,7 @@ jobs:
             echo "Checking availability of: ${image_url}"
 
             while [ $retry_count -lt $MAX_RETRIES ]; do
-              if docker manifest inspect "${DOCKER_REGISTRY}/${image}:${component_tag}" > /dev/null 2>&1; then
+              if docker manifest inspect "${image_url}" > /dev/null 2>&1; then
                 echo "Image available: ${image_url}"
                 break
               else


.github/workflows/release.yaml
```diff
@@ -160,14 +160,14 @@ jobs:
...
           read -a IMAGE_ARRAY <<< "$images"
-          
+
           MAX_RETRIES=120
           SLEEP_INTERVAL=60
           retry_count=0
-          
+
           for image in "${IMAGE_ARRAY[@]}"; do
             image_url="${DOCKER_REGISTRY}/${image}:${COMPONENT_TAG}"
```
almaslennikov: Here we also need to wait for stig-fips tag if the component is stig fips. We've encountered during the last run, that the job would wait for `<sriov-daemon-fips>:<non-fips-tag>`

.github/workflows/release.yaml
```diff
@@ -234,29 +234,29 @@ jobs:
...
             fi
-          
+
             # We always need to add '-stig-fips' suffix for corresponding image version
             # TODO: do not DOCA Driver image for now because $SOURCE_REPO is null one Network Operator will support a new version format 
             if [[ "$component" == *StigFips ]]; then
-              COMPONENT_TAG="${COMPONENT_TAG}-stig-fips"
-            fi 
-            
+              component_tag="${COMPONENT_TAG}-stig-fips"
```
almaslennikov, line 256: If a component is not stig-fips, where do we assign value to the local variable?

.github/workflows/release.yaml
```diff
@@ -160,20 +160,26 @@ jobs:
...
+            else
+              component_tag="${COMPONENT_TAG}"
+            fi
+
+            image_url="${DOCKER_REGISTRY}/${image}:${component_tag}"
             echo "Checking availability of: ${image_url}"
 
             while [ $retry_count -lt $MAX_RETRIES ]; do
-              if docker manifest inspect "${DOCKER_REGISTRY}/${image}:${COMPONENT_TAG}" > /dev/null 2>&1; then
+              if docker manifest inspect "${DOCKER_REGISTRY}/${image}:${component_tag}" > /dev/null 2>&1; then
```
almaslennikov: nit: we can use `$image_url` here

PR #1827, opened 2025-11-07
fix: support env vars for NIC operator

almaslennikov:
We need that for the DGX use case to support mlxfwreset before rebooting the node


almaslennikov:
@heyvister1 added an example comment to NCP example

PR #1871, opened 2025-11-27
fix: use Docker Buildx for multi-arch CI builds

rollandf:
Use docker/build-push-action with QEMU and Buildx to handle multi-arch builds automatically, eliminating manifest creation errors.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -18,27 +18,44 @@ jobs:
   docker-build-push:
     runs-on: linux-amd64-cpu4
     steps:
-    - uses: actions/checkout@v5
+    - uses: actions/checkout@v6
     - name: Determine docker tags
+      id: docker-tags
       run: |
         git_sha=$(git rev-parse --short=12 HEAD)                                  # short git commit hash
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yaml
```diff
@@ -18,27 +18,44 @@ jobs:
...
+            else
+              tags="$tags,${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:$tag"
+            fi
+          fi
+        done
+        echo "tags=$tags" | tee -a $GITHUB_OUTPUT
+    - uses: docker/setup-qemu-action@v3
+    - uses: docker/setup-buildx-action@v3
+      with:
+        buildkitd-config: /etc/buildkit/buildkitd.toml
```
heyvister1, line 44: According to the following [dockerdocs](https://docs.docker.com/build/buildkit/toml-configuration/) there is a difference in `buildkitd.toml` path for rootless or rootfull modes. 
Above path is for rootfull mode docker daemon. Is this our case here?

maze88 (approved):

PR #1966, opened 2026-01-09
fix: do not wait for stig images during release.yaml

almaslennikov:
STIG FIPS images are now built on gitlab asynchronously, do not block the pipeline for them here


rollandf:
All images are built asynchronously, not sure why we need this change

PR #1983, opened 2026-01-11
feat: add NIC conf Op fips images in release.yaml

rollandf:


hack/release.yaml
```diff
@@ -131,6 +131,18 @@ nicConfigurationConfigDaemon:
   sourceRepository: nic-configuration-operator
   version: network-operator-v26.1.0-beta.2
   nSpectScope: general
+nicConfigurationOperatorStigFips:
+  image: nic-configuration-operator-stig-fips
+  repository: nvcr.io/nvstaging/mellanox
+  sourceRepository: nic-configuration-operator
+  version: network-operator-v26.1.0-beta.2-stig-fips
+  nSpectScope: fedramp
```
e0ne, line 139: should it be gov-ready?

hack/release.yaml
```diff
@@ -131,6 +131,18 @@ nicConfigurationConfigDaemon:
...
+nicConfigurationConfigDaemonStigFips:
+  image: nic-configuration-operator-daemon-stig-fips
+  repository: nvcr.io/nvstaging/mellanox
+  sourceRepository: nic-configuration-operator
+  version: network-operator-v26.1.0-beta.2-stig-fips
+  nSpectScope: fedramp
```
e0ne, line 145: the same as above

e0ne (changes requested):

hack/release.yaml
rollandf: I copied from SRIOV config daemon:

```
SriovConfigDaemonStigFips:
  image: sriov-network-operator-config-daemon-stig-fips
  version: network-operator-v26.1.0-beta.2-stig-fips
  nSpectScope: fedramp
```

hack/release.yaml
maze88: from my perspective: lgtm - since items scoped as "fedramp" get mapped to the fedramp nspect release version, and to the gov-ready NGC publishing request.

PR #1985, opened 2026-01-12
task: add doca-driver-build tagging to release workflow

dahalperin:


maze88 (approved):
the decided approach is "safe" and isolated.

however, if release (tagging) cadence for doca-driver-build is coupled with that of network-operator (and all its components), then we should consider "just" adding it to the fully automated flow (i.e. by adding a `sourceRepository` field for it in `hack/release.yaml`).
if we do so, we'd have to consider the added field/value's behavior in any additional relevant workflows and [internal] jenkins pipelines.

rollandf (approved):

.github/workflows/release.yaml
```diff
@@ -105,6 +105,8 @@ jobs:
       fail-fast: false  # allow all jobs to run independently
       matrix:
         repo: ${{ fromJson(needs.get-managed-components.outputs.managed_repos) }}
+        include:
```
heyvister1, line 108: Nice! Can't we just align `doca-driver-build` same as other images in release.yaml and add "sourceRepository" in ofed attribute? what will it break?

cc @almaslennikov, @rollandf

maze88: as implied in [my earlier comment](https://github.com/Mellanox/network-operator/pull/1985#pullrequestreview-3650263272), this would impact multiple jenkins & github pipelines/workflows.
for example:
- in nspect automation we'd need to omit the step that explicitly adds this repo as a _source_ (which is nice - it would simplify that pipeline).
- however, in the _blackduck scan_ job it may require adding logic to explicitly skip scanning the doca-driver-build repository.

heyvister1 (approved):

PR #2035, opened 2026-01-19
feat: RDMA, SR-IOV, Multus auto-restart pods on config changes

### LINKED ISSUE #1831: Restart device plugins pods if configmap content changes
Currently, updating the configuration of device plugin (shared rdma, sriov) requires manual restart of the pods.

This could be handle by having a hash of the config map content in the DS spec, so it will have to restart the pod if it changes.

AdheipSingh:
Fixes #1831 

Adds automatic pod restart when device plugin ConfigMap content changes. A SHA256 hash of the  config is added to the DaemonSet pod template annotation (nvidia.network-operator.config-hash). When the config changes, the hash changes, triggering a Kubernetes rolling update.                 
                                                                                                     
  Changes:                                                                                           
  - pkg/utils/utils.go - Added GetStringHash() function                                              
  - pkg/consts/consts.go - Added ConfigHashAnnotation constant                                       
  - pkg/state/state_skel.go - Added SetConfigHashAnnotation() helper                                 
  - pkg/state/state_rdma_shared_device_plugin.go - Uses config hash                                  
  - pkg/state/state_sriov_dp.go - Uses config hash                                                   
  - pkg/state/state_multus_cni.go - Uses config hash                                                 
  - Unit tests added

### DIFF
--- a/pkg/consts/consts.go
+++ b/pkg/consts/consts.go
@@ -50,5 +50,6 @@ const (
 	OfedDriverSkipDrainLabelSelector = "nvidia.com/ofed-driver-upgrade-drain.skip!=true"
 	// ControllerRevisionAnnotation is the key for annotations used to store revision information on Kubernetes objects.
 	ControllerRevisionAnnotation = "nvidia.network-operator.revision"
+	// ConfigHashAnnotation is the key for annotations used to store config hash on pod templates.
 	ConfigHashAnnotation = "nvidia.network-operator.config-hash"
 )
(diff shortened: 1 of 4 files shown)


rollandf (commented):
Thanks for the PR!
Overall looks good.
Need to fix linter issue.
Also need to add to SriovDevicePlugin

AdheipSingh:
> Thanks for the PR! Overall looks good. Need to fix linter issue. Also need to add to SriovDevicePlugin

Lint is fixed. 
[Sriov](https://github.com/AdheipSingh/network-operator/blob/feat/config-hash-auto-restart/pkg/state/state_sriov_dp.go#L171-L176) is already added.

PR #2065, opened 2026-01-23
feat: add SOS report script

almaslennikov:


scripts/sosreport/kubectl-netop_sosreport
```diff
@@ -0,0 +1,953 @@
...
+    log_info "Collecting node information..."
+
+    local nodes_dir="$OUTPUT_DIR/nodes"
+
+    # All nodes (full YAML)
+    collect_resource "$nodes_dir/all-nodes.yaml" get nodes -o yaml
+
+    # Node summary
+    mkdir -p "$nodes_dir"
+    $KUBECTL_BIN get nodes -o wide > "$nodes_dir/nodes-summary.txt" 2>> "$ERROR_LOG" || true
```
heyvister1: I suggest adding an option (maybe as default) for label filter here to gather only relevant nodes details (e.g. `-l feature.node.kubernetes.io/pci-15b3.present="true"`). Since there could be cases were you're using this SOS report in a very large scale cluster

scripts/sosreport/kubectl-netop_sosreport
```diff
@@ -0,0 +1,953 @@
...
+    # Collect Services
+    collect_resource "$component_dir/services.yaml" get services -n "$OPERATOR_NAMESPACE" -l "$label_selector" -o yaml
+}
+
+# Collect all components
+collect_all_components() {
+    log_info "Collecting all component workloads..."
+
+    # Component definitions: name, label, type
+    declare -A components=(
```
heyvister1, line 691: I suggest to make this more generic. Maybe get these components out of release.yaml? 
1. Identify which release is used by customer (helm list ...)
2. Fetch release.yaml respectively
3. Build components map dynamically

scripts/sosreport/test-sosreport.sh
```diff
@@ -0,0 +1,278 @@
...
+    echo "PASS: Script version defined"
+else
+    echo "FAIL: Script version not defined"
+    exit 1
+fi
+
+# Test 9: Check CRD collection coverage
+echo ""
+echo "[Test 9] Checking CRD collection coverage..."
+required_crds=(
```
heyvister1, line 225: Same here. I suggest taking CRDs dynamically (network-operator/deployment, NCP etc')

scripts/sosreport/kubectl-netop_sosreport
```diff
@@ -0,0 +1,953 @@
...
+RED='\033[0;31m'
+GREEN='\033[0;32m'
+YELLOW='\033[1;33m'
+BLUE='\033[0;34m'
+NC='\033[0m' # No Color
+
+# Usage information
+usage() {
+    cat <<EOF
+Network Operator SOS-Report Collection Script v${SCRIPT_VERSION}
```
rollandf: Should we be consistent and use "NVIDIA Network Operator" in all placed?

scripts/sosreport/kubectl-netop_sosreport
heyvister1: When I think about, I suggest building network-operator workloads and CRD maps per each network-operator release. I mean, generate these maps and embed them under SOS report script. 
In this case customers can pull SOS report which is adapted to their network-operator release (also need to make sure its backward compatible with previous releases e.g. don't throw errors for missing CRDs/workload if they are not part of network-operator release).

We also need to make sure that SOS script can work in a disconnected environment, therefore pulling release.yaml from public GH could be blocked, so resource maps can be generated part of network-operator release.

scripts/sosreport/kubectl-netop_sosreport
```diff
@@ -0,0 +1,953 @@
...
+    if [ -n "$OPERATOR_NAMESPACE" ]; then
+        log_info "Using specified namespace: $OPERATOR_NAMESPACE"
+        return 0
+    fi
+
+    log_info "Auto-detecting Network Operator namespace..."
+
+    # Look for deployment with label app.kubernetes.io/name=network-operator
+    local ns
+    ns=$($KUBECTL_BIN get deployment -A -l app.kubernetes.io/name=network-operator -o jsonpath='{.items[0].metadata.namespace}' 2>/dev/null)
```
rollandf: Note that in Openshift I think we don't have this label. And with Helm the user can change the name with 'nameOverride'

scripts/sosreport/kubectl-netop_sosreport
```diff
@@ -0,0 +1,953 @@
+#!/bin/bash
+# Network Operator SOS-Report Collection Script
+# Copyright 2026 NVIDIA CORPORATION & AFFILIATES
```
rollandf: Please see why License check fails

almaslennikov: Added licenses to all needed files

scripts/sosreport/kubectl-netop_sosreport
almaslennikov: Added fallbacks for these cases to discover the net op namespace

scripts/sosreport/kubectl-netop_sosreport
almaslennikov: Thanks, fixed

scripts/sosreport/kubectl-netop_sosreport
almaslennikov: Added a script that generates the list of CRDs and components to collect. If a CRD is not present, the script will not fail, so backward compatibility should be fine

scripts/sosreport/kubectl-netop_sosreport
almaslennikov: Added a node selector option

scripts/sosreport/kubectl-netop_sosreport
```diff
@@ -0,0 +1,1092 @@
...
+        ns=$("$KUBECTL_BIN" get nicclusterpolicy -o jsonpath='{.items[0].metadata.annotations.meta\.helm\.sh/release-namespace}' 2>/dev/null)
+        if [ -n "$ns" ]; then
+            OPERATOR_NAMESPACE="$ns"
+            log_success "Detected operator namespace via NicClusterPolicy: $OPERATOR_NAMESPACE"
+            return 0
+        fi
+    fi
+
+    # Method 4: Check common namespaces for any network operator deployment
+    for ns in network-operator nvidia-network-operator openshift-network-operator kube-system; do
```
rollandf: `openshift-network-operator` is used by OCP Cluster Network Operator:

https://docs.redhat.com/en/documentation/openshift_container_platform/4.8/html/networking/cluster-network-operator#nw-cluster-network-operator_cluster-network-operator

scripts/sosreport/kubectl-netop_sosreport
```diff
@@ -0,0 +1,1092 @@
...
+# Collect metadata
+collect_metadata() {
+    log_info "Collecting cluster metadata..."
+
+    local metadata_dir="$OUTPUT_DIR/metadata"
+    mkdir -p "$metadata_dir"
+
+    # Collection info
+    cat > "$metadata_dir/collection-info.txt" <<EOF
+Network Operator SOS-Report
```
rollandf: NVIDIA Network Operator SOS-Report

rollandf (approved):

almaslennikov:
@heyvister1 PTAL

PR #2148, opened 2026-02-06
feat: add custom CLI arguments for Multus CNI 

e0ne:


pkg/state/state_multus_cni.go
```diff
@@ -129,6 +131,41 @@ func (s *stateMultusCNI) GetWatchSources() map[string]client.Object {
...
+// For args like "--key=value", it returns "--key".
+// For args like "--key", it returns "--key".
+func extractArgKey(arg string) string {
+	if idx := strings.Index(arg, "="); idx != -1 {
+		return arg[:idx]
+	}
+	return arg
+}
+
+// mergeMultusArgs merges default args with custom CliArgs, allowing CliArgs to override defaults.
```
adrianchiris, line 144: what if i want to remove an arg ?

maybe its better to:
1. have default set of args
2. if user provides args we use that (no merging)

e0ne: I agree, that's simplifies implementation and adds more flexibility. I'll update my pr

PR #2151, opened 2026-02-09
fix: Update rendered objects after manual changes

e0ne:
If user manually changes Network Operator manged objects (e.g. Multus CNI daemonset) operator should re-apply manifests.

We don't need to store object revision because it could be outdated after manual edit and Network Operator won't re-apply manifests.


adrianchiris:
@e0ne i thought we wanted this behaviour. i.e to allow custom edits in the field.
also object may be mutated by some webhook so we will end up always updating the object.

why do we need this change ?

PR #2257, opened 2026-03-12
feat: add module dependency check config and document UNLOAD_THIRD_PARTY_RDMA_MODULES

almaslennikov:
- Bridge UNLOAD_THIRD_PARTY_RDMA_MODULES boolean from driver
  container env vars to init container ConfigMap as
  unloadThirdPartyRdma, so the pre-flight check skips known RDMA
  modules the driver will handle
- Add moduleDependencyCheck section to init container ConfigMap
  with MOFED module list and /proc /sys volume mounts
- Document the env var and the three init container error categories
  (third-party RDMA, unknown kernel modules, userspace processes)
  in docs/mofed-container-env-vars.md

### DIFF
--- a/pkg/state/state_ofed.go
+++ b/pkg/state/state_ofed.go
@@ -244,6 +244,9 @@ type initContainerConfig struct {
 	InitContainerImageName string
 	SafeLoadEnable         bool
 	SafeLoadAnnotation     string
+	ModuleDepCheckEnable   bool
+	ModuleDepCheckModules  string
+	UnloadThirdPartyRDMA   bool
 }
 
 type ofedRuntimeSpec struct {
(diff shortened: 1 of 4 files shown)


pkg/state/state_ofed.go
```diff
@@ -734,12 +737,32 @@ func (s *stateOFED) getInitContainerConfig(
...
+	}
+
 	if image != "" {
 		initContCfg = initContainerConfig{
 			InitContainerEnable:    true,
 			InitContainerImageName: image,
 			SafeLoadEnable:         safeLoadEnable,
 			SafeLoadAnnotation:     upgrade.GetUpgradeDriverWaitForSafeLoadAnnotationKey(),
+			ModuleDepCheckEnable:   true,
+			ModuleDepCheckModulesJSON: `"mlx5_core", "mlx5_ib", "ib_umad",` +
```
e0ne: Since it's not a correct JSON, please rename this to something like `ModuleDepCheckModulesList`

PR #2323, opened 2026-03-31
feat: NCP Conditions

dahalperin:


controllers/nicclusterpolicy_controller_test.go
```diff
@@ -596,3 +596,235 @@ func getAppliedStateMessage(states []mellanoxv1alpha1.AppliedState, stateName st
 	}
 	return ""
 }
+
+// findNCPCondition returns the condition with the given type from the NicClusterPolicy status,
+// or nil if not found.
+func findNCPCondition(ncp *mellanoxv1alpha1.NicClusterPolicy, condType string) *metav1.Condition {
```
almaslennikov: use `meta.FindStatusCondition` instead

api/v1alpha1/nicclusterpolicy_types.go
```diff
@@ -44,6 +44,54 @@ const (
...
 )
 
+// Aggregate condition type constants — one set per NicClusterPolicy object.
+const (
+	// ConditionTypeReady is True when all configured components are ready.
+	ConditionTypeReady = "Ready"
+	// ConditionTypeDegraded is True when one or more components are in an error state.
+	ConditionTypeDegraded = "Degraded"
+	// ConditionTypeProgressing is True when one or more components are still deploying.
+	ConditionTypeProgressing = "Progressing"
```
almaslennikov: Should it be a single condition with three possible statuses?

rollandf: I agree, I think it will too much info.
maybe have `ready` only, with reason and message accordingly

dahalperin: These are intentionally separate conditions because they represent independent signals that can be true simultaneously — for example, Progressing=True and Degraded=True when some components are still deploying while others have errors. This is consistent with the SR-IOV conditions CRDs Design (see https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/1010/changes/17c3a7dfa2aadab2764dd6e977298bd7dc82fcdc), which uses the same Ready/Progressing/Degraded pattern.

PR #2433, opened 2026-04-14
fix: allow non-zero PCI domain in NicSelector PCI address validation

almaslennikov:
The kubebuilder validation pattern for pciAddresses hardcoded domain 0000, silently rejecting valid PCI addresses on systems with multiple PCI host bridges or PCIe switches (e.g., 0001:03:00.0). Relax the domain segment from literal "0000" to [0-9a-fA-F]{4}.


rollandf (approved):

e0ne (commented):
LGTM. Can we add some tests to cover this?

almaslennikov:
@e0ne I can add some unit tests in nic operator. Not sure if any tests are needed in this repo?

PR #2617, opened 2026-05-26
ci: branch only on RC for NVIDIA-owned repos in release pipeline

almaslennikov:
The release pipeline used to create a network-operator-MAJOR.MINOR.x branch in every managed repo on every release. That insulation is necessary for downstream forks of upstream projects, but pointless on repos NVIDIA fully owns — every beta leaves behind a permanent branch that nobody needs.

Introduce `downstreamFork` in hack/release.yaml (default false). When set to true, tagging behavior is unchanged: the release branch is created on every release and tags are cut from it. When omitted or false, beta tags are cut on the default branch, the release branch is created on the first RC, and GA on a missing branch hard-fails to surface a broken release process (RC was skipped or branch deleted).

Also pin the chart-source checkouts in create-release-pr to component_tag instead of release_branch — NVIDIA-owned chart providers have no release branch during the beta phase, but the tag always exists post create-tags-for-components.

doca-driver-build (added via matrix include, not in release.yaml) resolves to the NVIDIA-owned default via the same lookup.

### DIFF
--- a/hack/release.yaml
+++ b/hack/release.yaml
@@ -21,6 +21,7 @@ SriovNetworkOperator:
   image: sriov-network-operator
   repository: nvcr.io/nvstaging/mellanox
   sourceRepository: sriov-network-operator
+  downstreamFork: true
   version: network-operator-v26.4.0-beta.9
   chartLocation: deployment/sriov-network-operator-chart
   excludeChartFiles: ["Chart.yaml", "crds/k8s.cni.cncf.io_networkattachmentdefinitions_crd.yaml"]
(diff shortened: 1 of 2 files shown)


hack/release.yaml
```diff
@@ -13,6 +13,7 @@ SriovNetworkOperator:
   image: sriov-network-operator
   repository: nvcr.io/nvstaging/mellanox
   sourceRepository: sriov-network-operator
+  downstreamFork: true
```
e0ne, line 24: nit: it's enough to simplify it to `fork` or `isFork`

rollandf (approved):

rollandf (approved):

PR #2666, opened 2026-06-24
ci: sync Dependabot config to release branches for owned components

almaslennikov:
## What

Make the Release workflow keep Dependabot working on LTS release branches
(`network-operator-<MAJOR.MINOR>.x`) of the component repos.

## Why

Dependabot only acts on a repo's **default branch** unless a config entry
explicitly sets `target-branch`. So once we cut a release branch in a
component repo, that branch got **zero** dependency updates — security and
patch bumps never reached our LTS lines.

## How

When `create-tags-for-components` cuts a release branch in a component
repo, it now also opens a PR on that repo's **default branch** adding a
`target-branch` Dependabot entry for the new branch. The entry is a clone
of each existing default-branch update entry, retargeted and restricted to
**patch-only** bumps (major/minor ignored) to keep stable branches quiet.

Opt-in is **per component**, via a new `dependabotSyncManaged` flag in
`hack/release.yaml`. Only repos whose Dependabot config we own are touched
— forks whose config is synced from upstream, and repos with no config,
are excluded. Classification (derived by inspecting each source repo):

| Category | Repos |
|---|---|
| **Ours (flagged)** | network-operator-init-container, k8s-rdma-shared-dev-plugin, ib-kubernetes, ipoib-cni, nvidia-k8s-ipam, nic-feature-discovery, nic-configuration-operator, maintenance-operator, spectrum-x-operator, doca-driver-build, k8snetworkplumbingwg-plugins\* |
| **Upstream-managed (skip)** | sriov-network-operator, sriov-cni, ib-sriov-cni, sriov-network-device-plugin, ovs-cni, rdma-cni, node-feature-discovery, dra-driver-sriov |
| **No dependabot (skip)** | multus-cni |

\* `k8snetworkplumbingwg-plugins` is a fork, but upstream ships no
dependabot config and we added our own — so it's ours. A blanket
"skip all forks" rule would wrongly exclude it; the per-component marker
is authoritative.

### Workflow changes

- `determine-versions`: new `sync_dependabot` output — `true` for rc/ga,
  `false` for beta (betas come off `master`; the LTS branch is only real
  at rc).
- `get-managed-components`: new `dependabot_managed_repos` output, derived
  from the markers (Mofed maps to the `doca-driver-build` matrix include).
- `create-tags-for-components`: the single tag step is split into three
  ordered steps — **ensure branch → sync Dependabot → cut tag** — so the
  **tag is cut after** the Dependabot PR is raised. The sync step is gated
  on `sync_dependabot` + marker membership, **non-blocking**
  (`continue-on-error: true`) so housekeeping never holds up a release,
  and **idempotent**: skips when the branch is already in the config or a
  `dependabot-sync` PR is already open (covers rc.1/rc.2/ga re-runs).

## Scope note

This enables Dependabot **version updates** on LTS branches. Alert-driven
**security updates** remain default-branch-only — a GitHub limitation no
config can change.

## Validation

- `yq`/`jq` derivation of `dependabot_managed_repos` produces exactly the
  11 expected repos.
- The clone-retarget transform was run against `ib-kubernetes`'s real
  `dependabot.yaml`: 3 originals preserved, 3 retargeted clones added, the
  `gomod` clone keeps its controller-runtime/k8s ignores and gains the
  patch-only block; the idempotency check fires on re-run.
- `actionlint` passes clean on the workflow.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -36,6 +36,15 @@ jobs:
           # tag for components
           TAG="network-operator-${APP_VERSION}"
           echo "component_tag=${TAG}" >> $GITHUB_OUTPUT
+
+          # whether to sync Dependabot config to the release branch: rc/ga only,
+          # never beta (betas are cut from master, the LTS branch is only real at rc)
+          if echo "$APP_VERSION" | grep -q 'beta'; then
+            SYNC_DEPENDABOT=false
+          else
+            SYNC_DEPENDABOT=true
+          fi
+          echo "sync_dependabot=${SYNC_DEPENDABOT}" >> $GITHUB_OUTPUT
       - id: set-chart-version
         run: |
           CHART_VERSION=$(echo ${APP_VERSION#v})
@@ -79,6 +88,7 @@ jobs:
       docker_registry_managed_components: ${{ steps.determine-docker-registry.outputs.docker_registry_managed_components }}
       docker_registry_staging: nvcr.io/nvstaging/mellanox
       release_branch: ${{ steps.set-version.outputs.release_branch }}
+      sync_dependabot: ${{ steps.set-version.outputs.sync_dependabot }}
 
   get-managed-components:
     needs: determine-versions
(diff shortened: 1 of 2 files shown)


almaslennikov:
Thanks @greptile-apps — both addressed in the amended commit (force-pushed):

- **P1 (silent push failure after a closed sync PR):** the push now uses `git push --force-with-lease`. Since `actions/checkout` runs with `fetch-depth: 0`, the `refs/remotes/origin/dependabot-sync-*` tracking ref is present, so the lease re-cuts the CI-owned branch when a prior PR was closed without merging, while still refusing to clobber genuine concurrent pushes.
- **P2 (`MofedStigFips.dependabotSyncManaged` never read):** the `set-dependabot-repos` special-case now includes `doca-driver-build` when **either** `Mofed` **or** `MofedStigFips` is flagged, so the marker on `MofedStigFips` is no longer dead and clearing one Mofed entry alone cannot silently drop coverage. Verified: with `Mofed.dependabotSyncManaged` removed, `doca-driver-build` is still derived via the `MofedStigFips` flag.

`actionlint` passes clean on the workflow.

almaslennikov:
/retest-nic_operator_kind

The `nic_operator_kind` failure is environmental, not related to this change. The build failed because the GPU operator `ClusterPolicy` never became ready (`states not ready: [state-driver state-container-toolkit state-operator-validation state-device-plugin gpu-feature-discovery]`, pods stuck in `PodInitializing`), plus IB device setup errors in the harness (`/sys/class/net/net1/device/infiniband: No such file`). This PR only touches `.github/workflows/release.yaml` (the issue-triggered Release workflow, not run in PR CI) and `hack/release.yaml` (a data file) — `nic_operator_helm CI` and `check release-build` (which parses `hack/release.yaml`) both passed.

e0ne (approved):

rollandf (approved):

PR #2772, opened 2026-07-11
ci: pin reusable workflows for GA releases

almaslennikov:
Adds RC/GA release handling for reusable workflow branch pinning, scoped to the selected component release branches.

Changes:
- creates or reuses `cloud-orchestration-reusable-workflows` branch `network-operator-<major>.<minor>.x` for RC and GA releases
- opens a reusable-workflows Dependabot sync PR for that release branch with patch-only Docker updates
- pins reusable workflow references from `@main` to the release branch only for:
  - `nic-configuration-operator`
  - `sriov-network-operator`
  - `spectrum-x-operator`
- preserves existing non-`@main` workflow pins, such as SHA, tag, or branch refs
- captures the component tag commit after the optional pinning commit so RC/GA tags for selected components point at the pinned workflow state
- keeps beta releases flowing through a no-op reusable workflow prep step so beta tagging is not skipped

Validation:
- `go run github.com/rhysd/actionlint/cmd/actionlint@latest -shellcheck= .github/workflows/release.yaml`
- `ruby -e 'require "yaml"; YAML.load_file(".github/workflows/release.yaml")'`
- `git diff --check`
- local release-policy check confirming beta no-ops while RC/GA runs branch prep, Dependabot sync, and selected component pinning

Full actionlint with ShellCheck still reports existing quoting/style warnings in older release workflow blocks; structural actionlint and the changed release logic validate cleanly.

PR #2782, opened 2026-07-14
ci: pin reusable workflows on RC releases

almaslennikov:
Follow-up to #2772. The merged workflow pinned reusable workflows only for GA releases. This changes the same selected-component pinning flow to run for RC and GA releases, while beta releases remain excluded.

Changes:
- uses `needs.determine-versions.outputs.sync_dependabot == 'true'` as the shared RC/GA gate for reusable-workflows branch prep
- creates/reuses the reusable-workflows release branch for RC and GA
- runs reusable-workflows Dependabot sync for RC and GA
- pins reusable workflow references for the selected repos on RC and GA:
  - `nic-configuration-operator`
  - `sriov-network-operator`
  - `spectrum-x-operator`
- generates component release-branch Dependabot entries without copying default-branch ignore rules, so release branches can receive all version updates, including `k8s.io/*`, `sigs.k8s.io/controller-runtime`, and indirect Go dependency minor bumps
- keeps beta releases on the no-op path

Validation:
- `go run github.com/rhysd/actionlint/cmd/actionlint@latest -shellcheck= .github/workflows/release.yaml`
- `ruby -e 'require "yaml"; YAML.load_file(".github/workflows/release.yaml"); puts "release workflow YAML OK"'`
- `git diff --check`

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -99,12 +99,12 @@ jobs:
       RELEASE_BRANCH: ${{ needs.determine-versions.outputs.release_branch }}
       REUSABLE_WORKFLOWS_REPO: cloud-orchestration-reusable-workflows
     steps:
-      - name: Skip reusable workflows release prep for prereleases
-        if: ${{ contains(needs.determine-versions.outputs.app_version, '-') }}
-        run: echo "Reusable workflows release branch pinning runs only for GA releases"
+      - name: Skip reusable workflows release prep for beta releases
+        if: ${{ needs.determine-versions.outputs.sync_dependabot != 'true' }}
+        run: echo "Reusable workflows release branch pinning runs only for RC and GA releases"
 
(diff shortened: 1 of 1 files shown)


almaslennikov:
Amended the PR commit and force-pushed `566d7b0f`.

Dependabot release-branch sync now drops the default-branch `ignore` rules from generated component `target-branch` entries. The default branch config remains unchanged, but release branches can receive all version updates, including `k8s.io/*`, `sigs.k8s.io/controller-runtime`, and indirect Go dependency minor bumps such as `golang.org/x/net`.

Validation passed:
- `go run github.com/rhysd/actionlint/cmd/actionlint@latest -shellcheck= .github/workflows/release.yaml`
- `ruby -e 'require "yaml"; YAML.load_file(".github/workflows/release.yaml"); puts "release workflow YAML OK"'`
- `git diff --check`

PR #2947, opened 2026-07-29
chore: Update NAD CRD to match multus-cni changes

rollandf:
Add "nad" short name to CRD

See:
https://github.com/k8snetworkplumbingwg/multus-cni/pull/1476


e0ne (approved):
LGTM

PR #2950, opened 2026-07-30
feat: add Go integration coverage image and Helm toggle

BaryosOby:
Adds an optional Go integration-coverage path for the network-operator controller, intended for Jinx regression and similar E2E coverage workflows. Production behavior is unchanged unless coverage is explicitly enabled.
- CI publishes a second instrumented image (`<sha>-coverage`, plus `coverage` on `master`) alongside the normal image
- Runtime SIGUSR1 handler flushes in-memory counters to `GOCOVERDIR` without restarting the pod
- Helm overlay (`values-go-coverage.yaml`) and `operator.coverage.enabled` toggle wire up env, volume mount, and image selection

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -46,6 +46,12 @@ jobs:
         done
         echo "tags=$tags" | tee -a $GITHUB_OUTPUT
 
+        coverage_tags="${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:${git_sha}-coverage"
+        if [ -n "$git_tag" ]; then
+          coverage_tags="$coverage_tags,${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:${git_tag}-coverage"
+        fi
+        echo "coverage_tags=$coverage_tags" | tee -a $GITHUB_OUTPUT
+
         # all events reaching here are trusted:
         # - push to master/v*.x/tags: standard ci
         # - push to pull-request/*: copy-pr-bot is the trust gate; untrusted fork prs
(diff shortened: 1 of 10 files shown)


deployment/network-operator/templates/operator.yaml
```diff
@@ -53,16 +53,27 @@ spec:
         - name: {{ .Chart.Name }}
           # Replace this with the built image name
           image: "{{ .Values.operator.repository }}/{{ .Values.operator.image }}:{{ .Values.operator.tag | default .Chart.AppVersion }}"
+          {{- if .Values.operator.coverage.enabled }}
+          imagePullPolicy: IfNotPresent
```
rollandf: Why is this needed?

deployment/network-operator/values-go-coverage.yaml
```diff
@@ -0,0 +1,29 @@
+# Copyright 2026 NVIDIA
```
rollandf: I don't see a reason for this file

deployment/network-operator/README.md
```diff
@@ -164,6 +164,26 @@ If you prefer not to use `cert-manager`, set `operator.admissionController.useCe
...
 > Release_Namespace>.svc, e.g. network-operator-webhook-service.network-operator.svc  
 
+### Go integration coverage
+
+For integration test coverage collection (e.g. Jinx regression), enable the coverage overlay
+and set the immutable instrumented image tag published by CI (`<git-sha>-coverage`):
+
+```bash
+helm install network-operator ./deployment/network-operator \
+  -f deployment/network-operator/values-go-coverage.yaml \
```
rollandf: why not use the regular values.yaml and use `set` to enable the coverage?

rollandf (changes requested):
Thanks!
Please see comment about new helm values file

rollandf (approved):
LGTM

.github/workflows/ci.yaml
e0ne, line 80: can we re-use this step?

e0ne (changes requested):

.github/workflows/ci.yaml
BaryosOby: The two steps are almost the same on purpose — only the tags and GO_COVERAGE=1 differ. GitHub Actions doesn’t let us call the same step twice with different inputs unless we pull it into a composite action, and a matrix job would mess with the publish output the Helm/OCP jobs depend on. I’d rather leave the two steps as-is for now, but I can extract a composite action if you think it’s worth it.

e0ne (approved):

PR #2956, opened 2026-07-31
fix: allow skipping keep NCP hooks

almaslennikov:
## What this PR does

- adds a `keepNCP` Helm value, defaulting to `true` to preserve existing upgrade behavior
- gates the complete keep-NCP hook group, including its ServiceAccount, ClusterRole, ClusterRoleBinding, and Job
- documents `keepNCP: false` for first-time GitOps installations with no existing Helm-managed `NicClusterPolicy`
- adds Helm render tests for every `keepNCP` / `upgradeCRDs` combination and runs them in CI

## Why

The keep-NCP hook exists to protect a Helm-managed `NicClusterPolicy` when upgrading from older chart versions that created the custom resource. It has nothing to preserve on a fresh installation.

GitOps controllers such as Argo CD map both Helm `pre-install` and `pre-upgrade` hooks to `PreSync` and cannot distinguish an install from an upgrade. On affected Argo CD versions, the weight-0 RBAC hooks can remain running, so the weight-1 Job is never created and the initial sync cannot progress.

The new value gives fresh GitOps installations a supported way to omit these unnecessary hooks. Keeping the default at `true` avoids changing existing Helm upgrade behavior.

## Validation

- `make test-helm`
- `make lint-helm`
- `make check-release-build`
- `shellcheck hack/scripts/test-helm-hook-values.sh`

### DIFF
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -36,6 +36,8 @@ jobs:
         uses: actions/checkout@v7
       - name: lint helm chart
         run: make lint-helm
+      - name: test helm chart
+        run: make test-helm
   diff-manifests:
     name: check manifests
     runs-on: ubuntu-24.04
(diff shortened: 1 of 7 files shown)


rollandf (approved):
LGTM

PR #3019, opened 2026-08-14
fix: cap NicNodePolicy SR-IOV device plugin names

almaslennikov:
## Summary

- preserve the existing NicClusterPolicy SR-IOV device plugin DaemonSet name
- use `net-op-sriov-device-plugin-<policy-name>` for NicNodePolicy-owned DaemonSets
- cap the complete rendered DaemonSet name and selector label at 63 characters while retaining a valid terminal character
- keep SOS report discovery working across NicClusterPolicy and NicNodePolicy workloads
- document the policy-specific naming behavior

## Root cause

The SR-IOV device plugin template appended `-<NicNodePolicy name>` to the 36-character `network-operator-sriov-device-plugin` base. A valid 30-character policy name therefore produced a 67-character `name` selector label, which Kubernetes rejects because label values are limited to 63 characters.

## Verification

- `CGO_ENABLED=0 go build ./...`
- `CGO_ENABLED=0 go test ./pkg/state/... -count=1`
- focused max-length, exact-boundary, overlength, and trailing-separator render cases
- `CGO_ENABLED=0 make lint`
- `scripts/sosreport/generate-maps.sh --verify` in Linux/amd64
- `scripts/sosreport/test-sosreport.sh` in Linux/amd64

A repository-wide envtest run was also attempted with Kubernetes 1.36.2 assets. It was stopped after the unrelated `controllers` package made no progress for more than five minutes; the changed `pkg/state` suite and all checks listed above pass.

### DIFF
--- a/controllers/suite_test.go
+++ b/controllers/suite_test.go
@@ -44,6 +44,7 @@ import (
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
+	metricsserver "sigs.k8s.io/controller-runtime/pkg/metrics/server"
 
 	mellanoxcomv1alpha1 "github.com/Mellanox/network-operator/api/v1alpha1"
 	"github.com/Mellanox/network-operator/pkg/clustertype"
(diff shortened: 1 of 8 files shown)


almaslennikov:
Fixed the Unit-tests nil-pointer failures in a separate commit.

Root cause: the controller suite initialized its shared context inside the manager goroutine, allowing `BeforeAll` and the maintenance mock to observe a nil context. Context initialization is now synchronous, the goroutine no longer captures the shared `err`, and the envtest manager disables its fixed metrics listener to avoid local/parallel `:8080` collisions.

Verified locally with the exact CI target and supporting checks:
- `make test-coverage`
- all 28 controller envtest specs passed
- `CGO_ENABLED=0 go build ./...`
- `CGO_ENABLED=0 go test ./pkg/state/... -count=1`
- `CGO_ENABLED=0 make lint` (0 issues)

rollandf (approved):

PR #3020, opened 2026-08-14
fix: cap NicNodePolicy SR-IOV DP names on v26.4.x

almaslennikov:
Backport of #3019 to the `v26.4.x` release branch.

This uses `net-op-sriov-device-plugin-<policy>` for NicNodePolicy-owned SR-IOV device-plugin DaemonSets, caps the rendered name at 63 characters, and keeps selector/pod labels consistent. NicClusterPolicy and ConfigMap identities remain unchanged. The SOS-report collector uses the stable `app=sriovdp` label so it captures both NCP and NNP workloads.

Release-branch compatibility adaptations:
- omit the newer `NicNodePolicyStatus.Conditions` test fixture field
- preserve the generated SOS-report collector executable bit

Verification:
- `CGO_ENABLED=0 go test ./pkg/state/... -count=1`
- `CGO_ENABLED=0 go build ./...`
- `CGO_ENABLED=0 make lint`
- Linux/amd64 SOS-report shell syntax and generated-map verification
- exact `app=sriovdp` collector mapping check

The full release-branch SOS-report test retains a pre-existing failure for its stale `app=rdma-shared-dp` generated mapping; that unrelated behavior is not changed by this backport.

### DIFF
--- a/docs/heterogeneous-cluster-support.md
+++ b/docs/heterogeneous-cluster-support.md
@@ -78,6 +78,12 @@ Each NNP produces uniquely-named DaemonSets by appending the policy name as a su
 | NicClusterPolicy | `mofed-ubuntu22.04-<hash>-ds` |
 | NicNodePolicy `pool-a` | `mofed-ubuntu22.04-<hash>-pool-a-ds` |
 
+The SR-IOV device plugin keeps its existing
+`network-operator-sriov-device-plugin` DaemonSet name for NicClusterPolicy.
+NicNodePolicy instances use `net-op-sriov-device-plugin-<policy-name>` and the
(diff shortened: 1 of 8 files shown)


almaslennikov:
Follow-up 704f9621 fixes the failing release-build check without reverting the released v26.4.1 init-container digest. The v26.4.x branch had a generated digest pin that was not represented in hack/release.yaml, so make release-build rewrote it back to the tag on every PR. The release metadata now carries the existing digest, and make check-release-build is idempotent locally. No runtime image reference changed.

rollandf:
@almaslennikov should it be 26.7.x ?

rollandf:
Closing this one, not needed on 26.4.x

PR #3174, opened 2026-09-09
fix: do not cache Pods in the manager client

### LINKED ISSUE #3173: NicClusterPolicy controller caches every Pod in the cluster: operator memory scales with total pod count, including terminal pods
### What happened

The network-operator controller pod was OOMKilled in a loop against a 256Mi limit, then
again after we raised the limit to 512Mi. Its live heap was ~328 MiB while doing nothing
but idle reconciles of a NicClusterPolicy with three CNI components.

The cause is not a leak in the operator. The controller caches **every Pod in the cluster**
in full, so its memory is a function of total cluster pod count -- including terminal
(`Failed`/`Succeeded`) pods that it has no interest in.

At the time, one GPU node had accumulated 9,684 `Failed` pods from a node-level admission
failure loop. Cluster pod count was 9,895. Deleting the terminal pods, with no change to
the operator itself, dropped live heap 13x and it stayed flat:

| `go_memstats` | 9,895 pods in cluster | after deleting terminal pods (324 pods) |
| --- | --- | --- |
| `heap_alloc` (live) | 328.4 MiB | 24-29 MiB, flat over 3 samples |
| `next_gc` | 649.2 MiB | 44.4 MiB |
| `heap_released` | 52.7 MiB | 350.6 MiB |
| `process_resident_memory_bytes` | 454.1 MiB | 94.5 MiB, flat |

### Mechanism

1. `main.go:187` calls `ctrl.NewManager` with no `Cache` option, so the manager cache is
   unrestricted.
2. `NicClusterPolicyReconciler.Client` is `mgr.GetClient()` -- the cached client.
3. `handleWaitLabelsNoConfig` does
   `r.Client.List(ctx, pods, client.MatchingLabels{podLabel: ""})`
   (`controllers/nicclusterpolicy_controller.go:257`). A `List` through the cached client
   lazily starts an informer for that type, and `MatchingLabels` is applied **client-side**
   against the cache, so the informer caches all Pods cluster-wide with full spec and
   status.
4. The controller never watches Pods -- there is no `Owns(&corev1.Pod{})` or
   `Watches(&corev1.Pod{}, ...)` anywhere in `controllers/`. It watches only
   `NicClusterPolicy` and `Node`. So the cluster-wide Pod informer exists purely as a side
   effect of these `List` calls, to answer "which nodes currently have an OFED or
   nic-configuration daemon pod" -- a question about at most a handful of pods.

`next_gc` is what makes this diagnosable rather than looking like a leak. With `GOGC=100`
it is ~2x live-heap-at-last-GC, so `next_gc` of 649 MiB proves ~325 MiB was genuinely
reachable rather than garbage awaiting sweep (`heap_alloc` alone cannot distinguish those).
It also explains the OOM: nothing sets `GOMEMLIMIT` or `GOGC`, so with `next_gc` above the
cgroup limit the runtime waits for a GC it would only perform well past the ceiling, and
the kernel kills it first.

### Why this is easy to hit

Terminal pods are not garbage collected until kube-controller-manager's
`--terminated-pod-gc-threshold` (default 12500), so any repeating admission failure on a
single node quietly builds thousands of Pod objects while a controller retries. Two
triggers we can point at, both on stock configuration:

1. **Device plugin allocation failure.** A GPU in a bad state made the plugin's
   `GetPreferredAllocation` fail (`failed to get nvlink state: GPU requires reset`) while
   the plugin still advertised the full GPU count in both `capacity` and `allocatable`. The
   scheduler kept placing pods, kubelet admission kept failing, the ReplicaSet kept
   recreating: `UnexpectedAdmissionError` corpses at roughly 2,160/hour.
2. **Strict Topology Manager.** With `topologyManagerPolicy: restricted` (or
   `single-numa-node`) and `topologyManagerScope: pod`, a pod whose resources cannot be
   NUMA-aligned is rejected with `TopologyAffinityError` and left in `Failed`. A controller
   retrying against that node produces the same unbounded pile. Our GPU nodes run
   `topologyManagerPolicy: restricted`, `cpuManagerPolicy: static`,
   `memoryManagerPolicy: Static`, which is a normal configuration for pinned GPU workloads.

In both cases the network-operator is collateral: nothing is wrong with it, but it is the
process that dies, which sends whoever is on call in entirely the wrong direction. It also
means the operator's memory ceiling has to be provisioned against worst-case cluster pod
count rather than against its own working set, which is ~25 MiB.

### Suggested fix

Do not cache Pods. The codebase already establishes this idiom for exactly this reason --
`main.go:293` builds the drain client with

```go
// we need a client that doesn't use the local cache for the objects
drainKClient, err := client.New(restConfig, client.Options{
    Scheme: scheme,
    Cache: &client.CacheOptions{
        DisableFor: []client.Object{
            &sriovnetworkv1.SriovNetworkNodeState{},
            &corev1.Node{},
            &mcfgv1.MachineConfigPool{},
        },
    },
})
```

Adding `&corev1.Pod{}` to a `DisableFor` on the client the reconcilers use would remove the
cluster-wide Pod informer entirely. The Pod `List` calls run once per reconcile and their
label selectors are applied server-side when uncached, so they return only the handful of
pods actually wanted. The helpers in `controllers/mofed_wait_labels.go` already take
`c client.Client` as a parameter, so this does not require touching their logic.

Call sites that List Pods through the cached client:

- `controllers/nicclusterpolicy_controller.go:257`
- `controllers/nicnodepolicy_controller.go:149`
- `controllers/mofed_wait_labels.go:70`, `:76`, `:89`, `:131`
- `controllers/upgrade_controller.go:440`
- `pkg/migrate/migrate.go:175`, `:208`

Restricting the manager cache with `cache.Options.ByObject` for `&corev1.Pod{}` is the more
idiomatic alternative, but it is awkward here: the code needs pods matching
`OfedDriverLabel` **or** `NicConfigurationDaemonLabel`, and `ByObject.Label` accepts only a
single selector, which cannot express that disjunction.

Independently, setting `GOMEMLIMIT` from the container limit would make the runtime defend
the cgroup ceiling instead of being OOMKilled while below its GC target. That is a useful
guardrail but not a fix for the cache size.

### Environment

- network-operator v26.7.0 (Helm chart 26.7.0)
- Kubernetes v1.36.1, ~9-node bare-metal GPU cluster
- NicClusterPolicy with no `ofedDriver` (host OFED via the node image), carrying multus,
  container-networking-plugins and nv-ipam
- `operator.resources.limits.memory` 256Mi (chart default 128Mi), raised to 512Mi

NurullahMorshed:
Fixes: #3173

## Problem

The manager is built with no `Cache` option, and the reconcilers use
`mgr.GetClient()`. So the label-filtered Pod `List` in
`handleWaitLabelsNoConfig` lazily starts a cluster-wide Pod informer and applies
the selector client-side, caching every Pod in the cluster with full spec and
status.

No controller in `controllers/` ever watches Pods; there is no
`Owns(&corev1.Pod{})` or `Watches(&corev1.Pod{}, ...)`. The informer therefore
exists only to answer which nodes still run an OFED or nic-configuration daemon
pod, a question about a handful of pods.

The consequence is that the controller's memory scales with total cluster Pod
count rather than with its own working set, and terminal pods count. They are
not collected until kube-controller-manager's `--terminated-pod-gc-threshold`
(default 12500), so any repeating node-level admission failure quietly builds
thousands of Pod objects while a controller retries.

We hit this on a 9-node GPU cluster. A GPU that needed a reset made the device
plugin's `GetPreferredAllocation` fail while it still advertised the full GPU
count, so the scheduler kept placing pods, kubelet admission kept failing, and
a ReplicaSet kept recreating: 9,684 `Failed` pods on one node at roughly
2,160/hour. The network-operator was OOMKilled repeatedly as a result, first
against 256Mi and again after raising the limit to 512Mi, even though its own
working set is around 12 MiB.

Two triggers we can name, both on stock configuration:

1. Device plugin allocation failure, as above.
2. Strict Topology Manager. With `topologyManagerPolicy: restricted` or
   `single-numa-node` and `topologyManagerScope: pod`, a pod whose resources
   cannot be NUMA-aligned is rejected with `TopologyAffinityError` and left in
   `Failed`. A controller retrying against that node produces the same pile.
   That is a normal configuration for pinned GPU workloads.

In both cases the operator is collateral rather than the fault, which sends
whoever is on call in the wrong direction.

## Change

Disable the cache for `corev1.Pod` on the manager's client. This is the idiom
already used in this file for the drain client:

```go
// we need a client that doesn't use the local cache for the objects
drainKClient, err := client.New(restConfig, client.Options{
    Scheme: scheme,
    Cache: &client.CacheOptions{
        DisableFor: []client.Object{
            &sriovnetworkv1.SriovNetworkNodeState{},
            &corev1.Node{},
            &mcfgv1.MachineConfigPool{},
        },
    },
})
```

Setting it on `ctrl.Options.Client` covers every controller using
`mgr.GetClient()` in one place rather than touching the individual `List` call
sites (`nicclusterpolicy_controller.go`, `mofed_wait_labels.go`,
`nicnodepolicy_controller.go`, `upgrade_controller.go`, `pkg/migrate`).
Uncached, the label selector is applied server-side, so only the matching pods
are transferred and the informer is never started.

I also considered restricting the cache with `cache.Options.ByObject` for
`&corev1.Pod{}`, which is arguably more idiomatic. It does not fit here: the
code needs pods matching `OfedDriverLabel` **or** `NicConfigurationDaemonLabel`,
and `ByObject.Label` accepts a single selector, which cannot express that
disjunction.

## Measurement

Same cluster, same workload, only the image changed. The load was 500 Pod
objects with a nonexistent `spec.schedulerName`, so they stayed `Pending` as
pure API objects with no scheduler, kubelet or node involvement, each padded to
a realistic serialized size.

| arm | cluster pods | live heap | RSS | goroutines |
| --- | --- | --- | --- | --- |
| before, no pile | 323 | 21.9 MiB | 58.0 MiB | 259 |
| before, +500 | 825 | 40.6 MiB | 73.9 MiB | 259 |
| after, +500 | 824 | 13.3 MiB | 51.9 MiB | 248 |
| after, no pile | 325 | 12.8 MiB | 51.9 MiB | 248 |

38.1 KiB per Pod before, roughly 1 KiB after. After the change the controller at
824 pods sits below its own previous footprint at 323, because it no longer
holds any Pods. The 11 fewer goroutines are the Pod informer and its watch
machinery no longer running.

Live heap is `next_gc / 2` rather than `heap_alloc`: with `GOGC=100` the runtime
targets twice the live heap at the last GC, so that ratio isolates reachable
memory, whereas `heap_alloc` also counts unswept garbage and cannot distinguish
a leak from GC headroom.

## Verification

- `go build ./...`, `go vet ./...` and `gofmt -l` clean.
- Ran the built image on the affected cluster: `NicClusterPolicy` stayed `ready`
  with `MultusCNIReady`, `CNIPluginsReady`, `NVIPAMReady` and the aggregate
  `Ready=AllComponentsReady` all true.
- The `network.nvidia.com/operator.mofed.wait` node labels were identical to
  before: `false` on the five Mellanox-NIC nodes, absent on the other four. So
  the uncached List does not change behaviour.
- The NicClusterPolicy under test carries no `ofedDriver` (host OFED via the
  node image), which is the path that reaches `handleWaitLabelsNoConfig`.

### DIFF
--- a/main.go
+++ b/main.go
@@ -192,6 +192,18 @@ func main() {
 		HealthProbeBindAddress: probeAddr,
 		LeaderElection:         enableLeaderElection,
 		LeaderElectionID:       "12620820.mellanox.com",
+		// No controller watches Pods; they are only listed by label to find which
+		// nodes still run an OFED or nic-configuration daemon pod. A List through
+		// the cached client would lazily start a cluster-wide Pod informer and
+		// apply the label selector client-side, caching every Pod in the cluster
+		// so the operator's memory scales with total Pod count rather than with
+		// the handful of pods it cares about. Uncached, the selector is applied
+		// server-side and only the matching pods are transferred.
+		Client: client.Options{
+			Cache: &client.CacheOptions{
+				DisableFor: []client.Object{&corev1.Pod{}},
+			},
+		},
 	})
 	if err != nil {
 		setupLog.Error(err, "unable to start manager")


NurullahMorshed:
I've been distracted by the stuff this fix unblocked for me, but circling back to it now. I've rebased it @rollandf but found an issue with unit-tests. I'm following up with in another pr either tonight, or tomorrow

EDIT: PR #3265

almaslennikov (approved):

PR #3187, opened 2026-09-11
fix: collect complete sosreport context

almaslennikov:
## Summary

- report the collector version aligned with the chart and show the deployed Network Operator app version separately
- archive pod YAML plus current/previous logs for every container in every pod in the Network Operator namespace
- collect non-secret Helm release, chart, app-version, and revision metadata, with optional Helm CLI metadata
- render the operator, collector, release, and chart versions in the HTML report

## Testing

- `docker run --rm --platform linux/amd64 -v "$PWD:/workspace" -w /workspace python:3.13-bookworm bash scripts/sosreport/test-sosreport.sh`
- `docker run --rm --platform linux/amd64 -v "$PWD:/workspace:ro" -w /workspace python:3.13-bookworm bash scripts/sosreport/generate-maps.sh --verify`
- `shellcheck scripts/sosreport/kubectl-netop_sosreport scripts/sosreport/generate-maps.sh scripts/sosreport/test-sosreport.sh`
- `helm lint deployment/network-operator`
- `git diff --check`

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -665,9 +665,11 @@ jobs:
           yq -i         '.Mofed.repository = "${{ env.DOCKER_REGISTRY_MANAGED_COMPONENTS }}"' hack/release.yaml
           yq -i '.MofedStigFips.repository = "${{ env.DOCKER_REGISTRY_MANAGED_COMPONENTS }}"' hack/release.yaml
 
-          # Update chart versions
+          # Update chart and SOS-report collector versions
           yq -i '.version = "${{ env.CHART_VERSION }}"'  deployment/network-operator/Chart.yaml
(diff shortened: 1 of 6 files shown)


almaslennikov:
Addressed all three review findings in `60f688a4`:

1. **Automated release version sync:** `.github/workflows/release.yaml` now updates the SOS-report `SCRIPT_VERSION` from `APP_VERSION` before generating release artifacts.
2. **Non-Helm deployed version:** the collector now detects the operator Deployment version label or image tag/digest, with Helm appVersion as a fallback.
3. **Collision-safe workload grouping:** dynamically discovered component directories now include both workload kind and name (for example, `deployment-shared-worker` and `job-shared-worker`).

Added regression coverage for each case. Revalidated the Linux/amd64 SOS-report suite, generated maps, ShellCheck, Helm lint, actionlint on the edited workflow, and `git diff --check`.

rollandf (approved):

PR #3192, opened 2026-09-14
chore: add Rocky comment as not supported

rollandf:
Rocky Linux is not officially supported.
DOCA OFED driver containers are not available for Rocky Linux.


e0ne (approved):
Thanks, @rollandf!

PR #3193, opened 2026-09-14
fix: accept DOCA-Host build number in OFED version validation

rollandf:
From DOCA 3.6 the driver container tag carries the DOCA-Host build number instead of the MLNX_OFED version:

  before:  doca3.5.0-26.07-0.7.7.0-0-ubuntu24.04-amd64
  from 3.6: doca3.6.0-26.10-016000-ubuntu24.04-amd64

The DOCA branch of the version regex hard-required the MLNX_OFED segment right after the yy.mm part, so the new tags were rejected by the NicClusterPolicy and NicNodePolicy admission webhooks. Turn that segment into an alternation accepting either the MLNX_OFED version or the build number, keeping all previously valid versions working. The build number is matched as at least 6 digits since it is zero padded, which avoids accepting a truncated value.

No other changes are needed: state_ofed.go composes image names and precompiled tags with fmt.Sprintf, docadriverimages compares registry tags by exact equality, and the CRDs have no pattern constraint on ofedDriver.version.


almaslennikov (approved):

e0ne (commented):
LGTM

e0ne (approved):

PR #3219, opened 2026-09-15
feat: align DOCA Telemetry Service deployment with the DPF chart

stav-inbar:
The DaemonSet kept no state on the node and gave local clients no way to consume telemetry over IPC, so it could not stand in for the DTS deployment that DPF ships.
Give it the host paths it was missing: configuration and data now persist under /var/opt/doca-telemetry rather than inside the container, and logs land in /var/log/doca/telemetry. That prefix follows the OFED driver DaemonSet, which already keeps writable node-local state under /var/opt, instead of the /opt/mellanox/doca/services default documented for DTS.
Serving IPC needs both halves of its transport. Short messages go over UNIX domain sockets, which are filesystem objects that a shared IPC namespace does not expose, hence the ipc_sockets host path; payloads go through shared memory, hence hostIPC and the /dev/shm mount. How wide that mount should be is a user-facing choice, so it comes from a new shareHostSharedMemory field on DOCATelemetryServiceSpec: clients running directly on the node need the host's whole /dev/shm, while containerized ones only need /dev/shm/telemetry. It defaults to false, which narrows what DTS is given today, so node-local clients now have to opt in.
Persisting /config also changes what the container command has to do. Clearing it on every start would discard the node's configuration, and telemetry-init.sh only generates configuration into an empty /config, so FORCE_CONFIG_OVERWRITE is what keeps the ConfigMap authoritative. Drop PROMETHEUS_XCSET_JOIN_FIELDS and PROMETHEUS_XCSET_MANDATORY_TYPES while here, since they override behavior that belongs to the DTS configuration.


rollandf:
@jsofri PTAL

rollandf (approved):
Thanks, LGTM

rollandf:
Approved also by Roman from DTS

PR #3265, opened 2026-09-22
test: fix race and cleanup issues in NicClusterPolicy controller specs related to nv-iapm deletion

NurullahMorshed:
Previous PR #3174 ran into issues with unit-tests with cleanups and a race condition when removing nv-iapm from the nicclusterpolicies. i've added this fix, then had claude pick it apart and test it to make sure it resolve the issue. It spotted a few areas I missed that could use the fix as well.

---

The controllers suite fails intermittently in CI with five failures that share one root cause. Run 34412030344 is an instance, and run 34360573140, on a branch that only bumps a dependency version, shows the same signature, so it is not specific to any one change.

The root failure is in "should create nv-ipam and delete it after un-setting CR value". The spec reads the NicClusterPolicy, clears spec.nvIpam and calls Update, while the reconciler writes the status of that same object. When the reconciler wins the race the Update is rejected:

```
Operation cannot be fulfilled on nicclusterpolicies.mellanox.com "nic-cluster-policy":
the object has been modified; please apply your changes to the latest version and try again
```

The spec then aborts before the Delete at the end of its body. NicClusterPolicy is cluster scoped and every spec creates it under the same name, so the leaked object makes each following spec fail on Create with AlreadyExists, including the BeforeEach of the Upgrade Controller suite. One conflict is reported as five failures, and none of the five names the object that actually leaked.

What this changes:

Read-modify-write against an object the reconciler also writes is replaced with a merge patch, which carries no resourceVersion precondition and so cannot conflict. This is the pattern the neighboring specs in the same file already use for the same kind of mutation. The change is applied in three places: the nv-ipam spec, the Node taint update, and the NicClusterPolicy toleration update.

Cleanup is registered with DeferCleanup at the point each object is created, rather than deleted at the end of the spec body, so a spec that fails midway still removes what it created. The helper ignores NotFound, which lets the spec that deletes the NicClusterPolicy as part of what it verifies keep doing so. This does not stop a spec from failing, it keeps one failure from being reported as five.

How it was verified:

The failure was reproduced deterministically on master by patching the NicClusterPolicy between the spec's Get and its Update, standing in for the reconciler. That produces the exact CI error above and 6 failed specs out of 28.

Injecting a plain failure at the same point in the spec, to model the abort without the conflict, gives 6 failures on master and 1 on this branch, which is the cascade behavior the DeferCleanup change is meant to fix.

With the same deterministic conflict probe applied to this branch, the suite passes.

make test-coverage and make lint both pass locally against envtest 1.36, as do repeated runs of the controllers suite.

One note for maintainers, not addressed here: running the suite with -ginkgo.randomize-all fails two specs in upgrade_controller_test.go and macvlannetwork_controller_test.go. That reproduces identically on master with the same seed, so it is a pre-existing order dependency and not related to this change. CI does not use randomize-all.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


almaslennikov (approved):
LGTM. Thanks, @NurullahMorshed !

